PR #4, opened 2016-06-19
Io reactor

oktal:


src/common/reactor.cc
```diff
@@ -0,0 +1,594 @@
...
+ */
+struct SyncImpl : public Reactor::Impl {
+
+    SyncImpl(Reactor* reactor)
+        : Reactor::Impl(reactor)
+    {
+        shutdownFd.bind(poller);
+    }
+
+    Reactor::Key addHandler(
```
oktal: TODO:

`addHandler` is currently not thread-safe, meaning that if the function is called from multiple threads, things will go boom boom. However, it whould be thread-safe and wait-free with respect to reads. To make it thread-safe, we could use a `mutex` to protect the writes and use some kind of atomic `shared_ptr` implementation to protect the reads

PR #7, opened 2016-08-03
Fix build error on fedora 23

emonty:
random_device is in the random header here. Without this include,
the build fails.


win32asm:
fixed via commit ce51a51bb038649117dc3c69c33042100ebef2e5
Author: Yaroslav Leonov <alximik43@gmail.com>
Date:   Tue Aug 30 23:08:22 2016 +0300

    fix imports in tests/async_test.cc

Please close the issue.

PR #8, opened 2016-08-03
Rename output library to libpistache

emonty:
The library is pistache.io. Having the binary produced named "libnet" is
weird, and something that conflicts with libnet, which is a packet
creation/injection library.


emonty:
Whoops - upon further examination I understand the name libnet.

PR #9, opened 2016-08-30
potential fixes to memory leak

mikeburkat:


src/server/listener.cc
```diff
@@ -210,7 +210,7 @@ Listener::run() {
 
         int ready_fds = poller.poll(events, 128, std::chrono::milliseconds(-1));
         if (ready_fds == -1) {
-            if (errno == EINTR && g_listen_fd == -1) return;
+            if (errno == EINTR) continue;
```
oktal: Could you explain this change ?

mikeburkat: @wsourdeau can explain.

wsourdeau: Yes, the logic is that EINTR can occur in any case during an epoll_wait, and I don't see why the handling thereof shiould differ from one case to another. In practice, this caused a useless exception for me when executing the augmentor test under gdb, which was solved by the above.

PR #11, opened 2016-08-30
Update project structure for better project vendoring

misty-fungus:
Hi, I'm trying to vendor pistache project via cmake ExternalProject.

My intention is to setup an imported library and then link it to my targets.
A general mechanism for passing include directories with library is to add  `INTERFACE_INCLUDE_DIRECTORIES` property for target.

As of current pistache layout, i can not use source folder `include` directory in `INTERFACE_INCLUDE_DIRECTORIES`,  because this way headers will not be isolated in `pistache/*.h` folder, i'll have to import them in my project like `#include <http.h>`.

The other option is to compile && install pistache, and then add include directory of installed project, but
this is a hard way if you have N vendored projects.

A common solution for this is to place library headers in `${src}/include/<project>/` folder.

Just take a look at your vendored googletest headers layout.

This pull request moves headers to ``include/pistache` folder.


kiplingw:
@misty-fungus, I think you may be making things more difficult for yourself than you need to. This is what `pkg-config` is for.

PR #13, opened 2016-10-04
make headers incasesensitive

stevijo:
Headers are not case sensitive

### DIFF
--- a/src/common/http_headers.cc
+++ b/src/common/http_headers.cc
@@ -7,6 +7,7 @@
 #include "http_headers.h"
 #include <unordered_map>
 #include <iterator>
+#include <algorithm>
 #include <stdexcept>
 #include <iostream>
 
@@ -18,6 +19,12 @@ namespace Header {
 
 namespace {
     std::unordered_map<std::string, Registry::RegistryFunc> registry;
+
+    bool incasesensitiveCompare(std::string part1, std::string part2) {
+        std::transform(part1.begin(),part1.end(),part1.begin(),toupper);
+        std::transform(part2.begin(),part2.end(),part2.begin(),toupper);
+        return part1 == part2;
+    }
 }
 
 RegisterHeader(Accept);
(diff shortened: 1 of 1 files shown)


nelsonhf:
I would prefer you named the comparison function caseInsensitiveCompare...
Maybe also create a function findCaseInsensitive that would encapsulate all 3 calls to make the code easier to understand. I tried to do it but it seems my C++ is not that good (I had a problem with pointers and references).
Except for that, I tested it and it solves issue #86.

stevijo:
After looking at it again ist would probably also be better to just save all headers lower case/ upper case and then just use a transform before using find. Because iterating over the unordererd map is not particulary fast.

nelsonhf:
On the other hand, you'll almost never have more than 10-15 headers so optimizing may not be very important.

One problem with modifying the headers before saving is that debugging may be harder. But you'd detect the case where a request is sent with the same header in different cases.

I figured out how to make the function I was talking about:

```
std::unordered_map<std::string, Registry::RegistryFunc >::iterator
Registry::findCaseInsensitive(const std::string& name)
{
    return std::find_if(registry.begin(),registry.end(),[&](const std::pair<std::string,Registry::RegistryFunc>& header) {
        return incasesensitiveCompare(name,header.first);
    });
}
```
Then your Registry functions would look like

```
std::unique_ptr<Header>
Registry::makeHeader(const std::string& name) {
    auto it = findCaseInsensitive(name);
    if (it == std::end(registry)) {
        throw std::runtime_error("Unknown header");
    }

    return it->second();
}

bool
Registry::isRegistered(const std::string& name) {
    return findCaseInsensitive(name) != std::end(registry);
}
```

Something similar may be done for `Collection::remove()` but I'm not sure it would be any improvement

xoac:
this can now be closed. This was merged:
https://github.com/oktal/pistache/commit/54a373157f7bcdb4ceb3ddd6914bc6f587fd6b58

Tachi107:
Fixed by https://github.com/pistacheio/pistache/commit/54a373157f7bcdb4ceb3ddd6914bc6f587fd6b58

PR #14, opened 2016-10-05
Add PATCH method

dsoklic:
These two commits add PATCH to the router and client.


oktal:
Looks good to me, thanks for the contribution

PR #15, opened 2016-10-20
added missing returns, listener check if bound on deconstruct

underdoeg:
Those are some minor bugfixes.


oktal:
Looks good, thanks for the fixes

PR #20, opened 2016-10-25
BUGFIX: make it compile with -Wall -Werror

dan-42:
Hey everyone,

lately I checked out your awesome lib and I wanted to play around a bit with it. As from the examples the API looked very interesting.

As I always compile with `-Wall -Werror` there showed quite a view errors up like `signed-compare`, `unused variables`or `reorder`  I fixed all off the issues and If you are interested you can merge them back.

it might have not fixed it correctly, as I did it quickly. 
So please have a closer look at it before merging it.

I'm also open for discussions, if you see a big problem with this in general

cheers

dan

PS: in addition I'll fix that utf8 issue in the copyright comment


dan-42:
Hey guys,
also clang found some issues to, like structs vs. classes.  so they are fixed too!
I hope you can use my changes, as they really just make it more robust and usable when used in software that uses ```-Wall and -Werror``` as this is important when optimisation is enabled as well.

Your liibrary is very cool, I just missed that it is not multi platform (yet) but I would need that from such a library. And I don't have time to add that now. 

So keep up the good work, I'll be interested where this lib is heading :+1: 

so long dan

oktal:
Hey dan,

Thanks a lot for your contributions, fixing warnings is always a good thing.

I'll take a look when I have time this week-end

oktal:
Thanks for your Pull-Request. I commented some changes that you made that I'm not sure to fully understand.

Other than that, rest looks good

dan-42:
@oktal I might be blind, But I can't see any comments. I'm happy to explain my changes in more detail, but please point out what you don't understand.

But to see what I fixed, with the changes, just try to compile with  ``` -Wall -Werror ```

oktal:
You're right, I didn't submit submit the final review as I did not see github changed its way of reviewing Pull Requests, damn it.

You should be able to see my comments now

include/listener.h
```diff
@@ -43,7 +43,7 @@ class Listener {
 
     Listener(const Address& address);
     void init(
-            size_t workers,
+            ssize_t workers,
```
dan-42: bacause the value will be compared somewhere and the compile warns that you compare signed value with an unsigned one. e.g. hardware_concurrency() returns init which is a signed value.

Also you could make all size_t, but you should decide which one to use and stick to it

include/stream.h
```diff
@@ -296,7 +296,7 @@ class StreamCursor {
 
     };
 
-    bool advance(size_t count);
+    bool advance(ssize_t count);
```
dan-42: bacause the value will be compared somewhere and the compile warns that you compare signed value with an unsigned one. e.g. hardware_concurrency() returns init which is a signed value.

Also you could make all size_t, but you should decide which one to use and stick to it

src/common/http_header.cc
```diff
@@ -283,6 +281,10 @@ Connection::write(std::ostream& os) const {
     case ConnectionControl::KeepAlive:
         os << "Keep-Alive";
         break;
+    case ConnectionControl::Ext:
```
dan-42: When it should not happen, the you shoudl throw a execption or so. but a modern compile will complain that that value from the enum is not handled in the switch-case

tests/view_test.cc
```diff
@@ -41,30 +41,30 @@ TEST(view_test, test_vector)  {
...
     ASSERT_EQ(v1[0], 4);
     ASSERT_EQ(v1[3], 7);
 
     auto v2 = make_view(arr1, 2);
-    ASSERT_EQ(v2.size(), 2);
+    ASSERT_EQ(v2.size(), size_t(2));
     ASSERT_EQ(v2[1], 5);
     ASSERT_THROW(v2.at(3), std::invalid_argument);
 
-    std::array<int, 4> arr2 { 6, 8, 1, 2 };
```
dan-42: yep but I cannot recall what it was

src/common/reactor.cc
```diff
@@ -262,12 +264,12 @@ struct SyncImpl : public Reactor::Impl {
...
 
         static Polling::Tag encodeTag(const Reactor::Key& key, uint64_t value) {
-            auto index = key.data();
+          //  auto index = key.data();
             // The reason why we are using the most significant bits to encode
             // the index of the handler is that in the fast path, we won't need
             // to shift the value to retrieve the fd if there is only one handler as
             // all the bits will already be set to 0.
-            auto encodedValue = (index << HandlerShift) | value;
+           // auto encodedValue = (index << HandlerShift) | value;
```
dan-42: srsly?  because it is unused, and a moden compiler does not like that and clean code should not have this

dan-42:
@oktal please see my comments. As I told before you can use my changes or not, that up to you. But I will not put any more effort in this library, as I personaly focus on cross-plattform development, and there is your library not suitable(Linux only).  

But anyway, I liked the user-api and keep up the good work!  

BTW: happy new year!

include/listener.h
oktal: Alright so I'd rather cast the return type of `hardware_concurrency()` to unsigned. workers should definitely not be negative

src/common/http_header.cc
oktal: Right. I'm actually not very happy with the current handling of extensions in various headers. I think I'll try to use `std::variant` to represent them and have a special `Extension` struct

src/common/reactor.cc
oktal: You are actually god damn right and guess what, you discovered a severe bug. I was actually lucky because I'm registering only one handler in the Reactor system but part of this code should be fixed, good catch !

oktal:
Alright,

Thanks for your work. I'll definitely  take your comments into account and fixes some of the issues you reported.

Thanks for taking time to help on that Pull Request, I'll keep it open until I address it completely.

Happy new year to you too :)

PR #28, opened 2016-11-14
Throw only when there is a polling error

lsimard:


oktal:
Hello,

This fix should definitely be merged-in. However, what about directly modifying the `Poller::poll` function so that every part of the code that uses the polling mechanism will benefit from it ?

lsimard:
Looks like it is already done!

PR #32, opened 2016-12-16
Fix Date header and add Last-Modified and Content-Disposition headers

jdoverholt:
I added the missing implementation for the Date::write() method, so that it functions, and created new implementation for two additional headers: Last-Modified and Content-Disposition.


oktal:
Hello,

Thanks for the contribution. Could you please add some unit-tests in the corresponding test file ? (headers_test.cc)

Other than that, it all looks good to me

kiplingw:
@jdoverholt

PR #39, opened 2017-01-10
Fix for FD leak

### LINKED ISSUE #38: FD leak in streamFile
There is lead of the file descriptor used to check existence using fstat. Similarly the fd used to read the file is not closed. This is critical issue bringing down the system.

run with valgrind --tool=memcheck --track-fds=yes ... to check.

santanusinha:
Fixes #38


src/common/http.cc
```diff
@@ -681,6 +681,11 @@ serveFile(ResponseWriter& response, const char* fileName, const Mime::MediaType&
         throw HttpError(Code::Internal_Server_Error, "");
     }
 
+    res = ::close(fd);
```
oktal: I think that the fd should also be closed if the `fstat` call above fails.

Actually,  while thinking about it, maybe could we benefit from a scope guard kind of construct to make sure the fd is correctly closed on all possible paths when an exception arises.

We would have something like `Scope_Failure(::close(fd));` which will close the fd in its destructor if unwinding occurs (see [std::uncaught_exception](http://en.cppreference.com/w/cpp/error/uncaught_exception))

src/common/stream.cc
```diff
@@ -13,16 +13,28 @@
...
 FileBuffer::FileBuffer(const std::string& fileName)
     : fileName_(fileName)
+    , fd_(-1)
 {
     init(fileName.c_str());
 }
 
+FileBuffer::~FileBuffer()
+{
+    if( -1 != fd_ ) {
```
oktal: I did not see when are you setting the fd in the `FileBuffer` ? I would expect this if to always fail

src/common/stream.cc
```diff
@@ -13,16 +13,28 @@
...
+        int res = ::close(fd_);
+        if (res == -1) {
+            throw std::runtime_error("Could not get file stats");
```
oktal: I would avoid throwing in a destructor, as throwing during stack unwinding will call end up calling `std::terminate`, which we do not want.

src/common/http.cc
santanusinha: Makes sense .. will change

src/common/stream.cc
santanusinha: hm.. no other way than to eat this up silently...

src/common/http.cc
santanusinha: A better option is to probably create something like FdGuard(int &fd_) and use that as a member/in function scope to close the fd in a guaranteed manner? That way it will close irrespective of whether there was an exception or not.

oktal: Yep, that would work as well

oktal:
Hello, any update on this PR ? I'd like to merge it as it fixes a severe issue.

santanusinha:
Got caught up with some office work. I will push the changes early this
week.

On Mar 11, 2017 10:32 PM, "Mathieu Stefani" <notifications@github.com>
wrote:

Hello, any update on this PR ? I'd like to merge it as it fixes a severe
issue.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/oktal/pistache/pull/39#issuecomment-285881002>, or mute
the thread

xoac:
Hi, do you have any update?

To FDGuard here is my implementation:
```struct FD {
    explicit FD(const int fd): _fd(fd) {}
    FD(FD &&fd):_fd(fd._fd) {fd._fd = -1;} //move

    explicit operator bool() const {return _fd > -1;}
    int& pure_descriptor() {return _fd;}

    ~FD() {if(!(*this)) close(_fd); }

private:
    FD& operator=(const FD & _) = delete;
    FD(const FD& _) = delete;

    int _fd;
};
static_assert(sizeof(FD) == sizeof(int), "***");```

And second option when `fd` is shared between threads. And the last one should close this:
```using sharedFD_Guard = std::shared_pointer<int, FDcloser>
//and FDcloser
struct FDcloser {  //file descriptor closer!
    int operator()(const int *fd) const {
        return close(*fd);
    }
};```

More information about second option can be found [here]( https://dev.krzaq.cc/post/you-dont-need-a-stateful-deleter-in-your-unique_ptr-usually/)

**Comment by xoac** [comment]
You should also rebase. The `include` changed to `include/pistache`. And probably more changes. 
At my level of knowledge I don't know how to add you pull request with patch.

This link create incorrect patch https://github.com/oktal/pistache/pull/39.patch

**Comment by chshaiiith** [comment]
@oktal can you please merge these changes if they are correct? I am stuck with the same issue.

File: include/pistache/os.h [hunk 4]
```diff
@@ -139,4 +139,13 @@ class NotifyFd {
     int event_fd;
 };
 
+class FdScopeGuard {
+public:
+    FdScopeGuard(int &fd);
+
+    ~FdScopeGuard();
```
milipili: Reference should not be used I think

src/common/os.cc
```diff
@@ -294,4 +294,15 @@ NotifyFd::tryRead() const {
...
 }
 
+FdScopeGuard::FdScopeGuard(int &fd)
+    : fd_(fd)
+{
+}
+
+FdScopeGuard::~FdScopeGuard()
+{
+    if(fd_ > 2) {
```
milipili: Simply `>= 0`. Nothing guarantees that fd won't be 1. Nothing forces standard output, error and input to be open, and they are not sometimes.

include/pistache/os.h
milipili: Actually I don't think that this class does anything that could be done in `FileBuffer`

dennisjenkins75: I agree.  ISTM that a much simpler solution is to ensure that the constructor for FileBuffer forces "fd_" to -1, and the destructor closes fd_ if it is >= 0.

santanusinha:
Go ahead and close it. The proposed fix is better for obvious reasons. This
was more of a hack I had to put in place to stop breaking production.

PR #45, opened 2017-02-01
Address::fromUnix gets the port wrong

GeorgeKT:
It does not convert from network byte order to host byte order for the port field of the struct sockaddr_in.
This is easily solved by using ntohs

IMHO, you should also rename this. fromUnix suggests that it is expecting a sockaddr_un (an address for a unix domain socket), not a sockaddr_in.


oktal:
Looks good :+1: 

About the name of the function, what do you suggest ? Would `fromPosix` be a better fit ?

GeorgeKT:
I would go for fromSocketAddress, fromPosix or fromUnix are just pointing to a platform, and not what the thing is you are converting from.

oktal:
Alright, merging this PR for now and will be thinking about the naming later.

Thanks for the contribution

Issue #29, opened 2016-11-17
Crashing in multhreaded application wtih GDB


michalpelka:
Framework crashes when working in multithread app with GDB connected.
Everything works fine unitl no breakpoints are set in code.
 `
terminate called after throwing an instance of 'Net::Error'
  what():  Polling: Interrupted system call
`

Code to reproduce with GDB:
https://gist.github.com/michalpelka/d7b5ee940597fe9dc5032abdc979c75c

Thank you for help.

michalpelka:
I suggest a little fix 
https://gist.github.com/michalpelka/4963eabd2c3fe4a27b199bef8b471604

It resolves my problem.

tskomudek:
Had the same issue with debugging and can confirm your fix solved it. 
Thx a lot.

nabilasalmi:
Also had the same issue on gdb while debugging a multithreaded application, and this fix solved my problem. Thanks very very much !!!

RossRogers:
@oktal  would @michalpelka 's edit be acceptable to the trunk?

RossRogers:
@oktal,
Here's the edit:


```patch
diff --git a/src/common/os.cc b/src/common/os.cc
index f826f70..083fa63 100644
--- a/src/common/os.cc
+++ b/src/common/os.cc
@@ -187,8 +187,12 @@ namespace Polling {
     Epoll::poll(std::vector<Event>& events, size_t maxEvents, std::chrono::milliseconds timeout) const {
         struct epoll_event evs[Const::MaxEvents];
 
-        int ready_fds = epoll_wait(epoll_fd, evs, maxEvents, timeout.count());
-        if (ready_fds > 0) {
+        int ready_fds = -1;
+        do {
+            ready_fds = epoll_wait(epoll_fd, evs, maxEvents, timeout.count());
+        }while (ready_fds < 0 && errno == EINTR);
+
+        if (ready_fds >0) {
             for (int i = 0; i < ready_fds; ++i) {
                 const struct epoll_event *ev = evs + i;
```

oktal:
Hello,

Would you like to submit a PR directly to the project ?

RossRogers:
Certainly @oktal .  Just asking you if that edit looks ok.

oktal:
Closing this issue, a fix has been merged in master

aprole:
I'm experiencing the same problem on the oktal master branch. I don't see that change in there.

kiplingw:
[...]

PR #48, opened 2017-02-06
Enables gdb debugging by fixing the polling code to allow interrupts.

RossRogers:
@michalpelka hit an issue ( #29 ) where attaching gdb and stepping
through code fails with a `terminate called after throwing an instance of
'Net::Error' what(): Polling: Interrupted system call`.  @michalpelka then
identified the fix which involves continuing to call `epoll` in the event
of an interrupt

After build `make test` passed and I can confirm that the fix enables the 
use of `gdb` and `gdbtui`.


oktal:
Alright, looks good to me.

Thanks for the contribution :+1:

PR #50, opened 2017-02-21
fixes #49

### LINKED ISSUE #49: HTTP server stops responding when sending large response
Sometimes when sending a large response the HTTP server stops responding. I believe the culprit is this line https://github.com/oktal/pistache/blob/master/src/common/transport.cc#L218 which should read "totalWritten == buffer.size()"

alanbirtles:
fix Buffer::Detach copying incorrect data range
fix Transport::asyncWriteImpl expecting the incorrect number of bytes to
be written when a write is split over multiple sends

### DIFF
--- a/include/stream.h
+++ b/include/stream.h
@@ -143,7 +143,7 @@ struct Buffer {
         char *newData = new char[len - fromIndex];
         std::copy(data + fromIndex, data + len, newData);
 
-        return Buffer(newData, len, true);
+        return Buffer(newData, len - fromIndex, true);
     }
 
     const char* const data;
--- a/include/stream.h
+++ b/include/stream.h
@@ -141,7 +141,7 @@ struct Buffer {
             throw std::invalid_argument("Invalid index (> len)");
 
         char *newData = new char[len - fromIndex];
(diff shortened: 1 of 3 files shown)


alanbirtles:
I'm still seeing an issue even with the above fixes, when asyncWriteImpl is called with status == Retry and the write blocks the number of bytes successfully written isn't stored anywhere so asyncWriteImpl continuously tries to write the same bytes until the client receives the expected number of bytes then the server blocks and wont respond to any further requests

alanbirtles:
I think 9d5a413 completes the fix but I'm not sure if this is the best way to implement it?

oktal:
Hello,

I left a comment on your PR to discuss the way you implemented your fix

antoniocoratelli:
Hi @oktal ,
I was looking for a modern C++11 http server implementation, and I want to thank you for making available on github this awesome library.
I need to serve really big files ( 500+ MB ) and I'm having issues with multiple file requests -- from the second http request the server stops responding.
Do you think this PR is stable enough to be used? Is there any chance that this PR is merged in master?

Best,
Antonio

Issue #51, opened 2017-02-26
How does one add header CORS in Pistache ?


GuacheSuede:
Title is question, cannot seem to find any documentation on it. Thanks

GuacheSuede:
@oktal

Loufylouf:
Looking at the set of headers supported, it seems that Access-Control-Allow-Origin is not one of them, so you would have to add it following the documentation http://pistache.io/guide/#defining-your-own-header . I need it for a project too, so once tested, I'll probably send a PR with this header part of the official set of headers supported.

Loufylouf:
@GuacheSuede As you can see, this is fairly simple, then you just have to add a header as you would normally with Pistache. I think it has its place in the set of headers by default as it can be very useful for jQuery users that would like to use a REST API provided by Pistache (for example).

GuacheSuede:
Great, Thank you for the pull request.

xinthose:
```
response.headers().add<Http::Header::AccessControlAllowOrigin>("*");
response.send(Pistache::Http::Code::Ok, "Hello World");
```

xinthose:
> > ```
> > response.headers().add<Http::Header::AccessControlAllowOrigin>("*");
> > response.send(Pistache::Http::Code::Ok, "{\"message\":\"success\"}");
> > ```
> 
> Hi xinthose
> i add this code to my doLigin() method, but nothings happen.
> where should I add this code?

At the beginning of the route function.  So like so:

```
void doLogin (const Pistache::Rest::Request& request, Pistache::Http::ResponseWriter response)
{
    response.headers().add<Pistache::Http::Header::AccessControlAllowOrigin>("*");

    // your code
}
```

msoodb:
> > > ```
> > > response.headers().add<Http::Header::AccessControlAllowOrigin>("*");
> > > response.send(Pistache::Http::Code::Ok, "{\"message\":\"success\"}");
> > > ```
> > 
> > 
> > Hi xinthose
> > i add this code to my doLigin() method, but nothings happen.
> > where should I add this code?
> 
> At the beginning of the route function. So like so:
> 
> ```
> void doLogin (const Pistache::Rest::Request& request, Pistache::Http::ResponseWriter response)
> {
>     response.headers().add<Pistache::Http::Header::AccessControlAllowOrigin>("*");
> 
>     // your code
> }
> ```

many thanks. this is very helpful.

PR #53, opened 2017-03-09
Add the header "Access-Control-Allow-Origin" to the set of headers de…

LaurentLouf:
…fined by default in Pistache. Fixes issue #51.

### DIFF
--- a/include/http_header.h
+++ b/include/http_header.h
@@ -270,6 +270,32 @@ private:
     uint64_t value_;
 };
 
+
+
+class AccessControl : public Header {
+public:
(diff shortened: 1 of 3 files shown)


include/http_header.h
```diff
@@ -270,6 +270,32 @@ class ContentLength : public Header {
...
+        : origin_(origin)
+    { }
+
+    explicit AccessControl(const std::string& origin) :
+        origin_(origin)
+    { }
+
+    void parse(const std::string& data);
+    void write(std::ostream& os) const;
+
```
oktal: You should also add a method to retrieve the `origin_` string

PR #57, opened 2017-03-23
Fixed 4k limitation

cneira:
Hi, 
This patch fixes the 4k limitation on requests, the buffer will resize to match the data from the resquest, I have tested using 70k json files. The downside is that the input is greater that our memory, we'll get a memory exception.

### DIFF
--- a/include/common.h
+++ b/include/common.h
@@ -52,6 +52,7 @@ namespace Const {
 
     static constexpr int MaxBacklog = 128;
     static constexpr int MaxEvents = 1024;
-    static constexpr int MaxBuffer = 4096;
+    static constexpr int SmallBuffer = 4096;
+    static constexpr int MaxBuffer = 80096;
     static constexpr int ChunkSize = 1024;
 }
--- a/include/common.h
+++ b/include/common.h
@@ -52,7 +52,7 @@ namespace Const {
 
     static constexpr int MaxBacklog = 128;
     static constexpr int MaxEvents = 1024;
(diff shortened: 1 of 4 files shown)


oktal:
Hello,

Thanks for the Pull Request. Before I look into it more deeply, I have several comments to make.

First, I noticed that you changed the coding style of files you changed.
While I understand that it might be better for you to work with (maybe your editor automatically reformatted the code the way you wanted), I think it pollutes the Pull Request and also make it inconsistent with the rest of the project.
Could you please revert that changes and only keep your fix ?

Now, about your changes, I think that `std::vector` is indeed the way to go. However, I don't think that it should allocate memory for *every request*. What I mean by that is that even small requests will now require memory allocation which is not great. I think that instead, we should have a **small requests optimization** whereby small requests (<4k ?) won't require any memory allocation at all and will use space allocated on the stack.

Finally, I don't think that the underlying buffer should grow indefinitely. Otherwise, like you said, it will run out of memory which is not ideal. Instead, I think that we should, for now, put a hard limit on the amount of memory that the server is allowed to allocate to parse a given request (it can be hardcoded for now). If we reach that particular threshold, we can then `throw` which will it turn generate an `500 Internal Server Error` response.

cneira:
Hi, 
I totally agree with the points you have exposed. 
I'll change the allocation to only do heap allocation when the request is greater that the current defined  threshold.
As you said we should stack allocate small requests which would be defined as those < 4k.

PR #58, opened 2017-03-23
Patches for Arch Linux packaging

Martchus:
This PR contains 2 patches I created for Arch Linux packaging.

The first patch renames the library to `pistache` to avoid a conflict with libnet (https://www.archlinux.org/packages/extra/x86_64/libnet). I understand if you don't like renaming the targets because all projects using it would need to adopt. Note that libnet is more popular and there already packages for different distributions (https://repology.org/metapackage/libnet/versions).

The second patch builds currently on top of the first one. It adds CMake config files so the library can be used more easily from other CMake projects. Using the lib via such a config is better because it preserves link libs, compile definitions and other target properties.

BTW: For the package see https://aur.archlinux.org/packages/pistache-git / https://github.com/Martchus/PKGBUILDs/tree/master/pistache/git

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required (VERSION 2.8.7)
+cmake_minimum_required (VERSION 3.0.2)
 project (pistache)
 include(CheckCXXCompilerFlag)
 
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -40,3 +40,23 @@ install(
   DIRECTORY "${PROJECT_SOURCE_DIR}/include/pistache"
   DESTINATION ${include_install_dir}
   FILES_MATCHING PATTERN "*.*h")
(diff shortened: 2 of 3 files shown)


jeffvandyke:
As at least one user, I too would much rather have it be called libpistache instead of libnet, that always kind of bugged me.

oktal:
Hello,

I recently renamed the library to libpistache. Could you please try to rebase your changes on top of my last changes, as the CMakeLists was also edited in the process ?

Thanks for the contribution !

Martchus:
Rebased - there is only the commit for exporting targets left. But maybe I should test it first by rebuilding the package.

Maybe it would also make sense to have separate targets to explicitly build a static and a shared lib.

Martchus:
Just rebuilt the package, rebased commit should work now.

oktal:
Great, I trust you on that one, merging it in.

Thanks again

PR #60, opened 2017-03-27
Fix missing bind() and wrong comment

Martchus:
I assume Endpoint::bind() is supposed to be a wrapper for the bind() method of the underlying listener.


oktal:
Hello,

Thanks for the Pull Request, merging it

PR #62, opened 2017-03-29
Small buffers optimization

cneira:
- Fixed indentation to follow project standards (do you clangformat?, if so, could you share your settings? )
- Added Smallbuffer and reallocation 
- Removed dynamic allocation

### DIFF
--- a/include/common.h
+++ b/include/common.h
@@ -52,6 +52,7 @@ namespace Const {
 
     static constexpr int MaxBacklog = 128;
     static constexpr int MaxEvents = 1024;
-    static constexpr int MaxBuffer = 4096;
+    static constexpr int SmallBuffer = 4096;
+    static constexpr int MaxBuffer = 80096;
     static constexpr int ChunkSize = 1024;
 }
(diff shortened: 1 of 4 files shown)


oktal:
Hello,

Thanks for the Pull Request.

That's a good start and you're nearly done. The only thing that I have to comment is the way you implemented the "small buffer optimization". You modified `ArrayStreamBuf<N>` to now use a `std::vector<char>` instead of a fixed-size array. However, even though you are reserving size of the vector, it is not small-size optimized. For requests whose size < N, we should use the old `std::array<char, N>` as a backend storage. Only when requests exceed that size, we should fallback to `std::vector<char>` and start allocating memory. That way, small requests won't require any memory allocation at all.

I then propose renaming `ArrayStreamBuf<N>` to `SmallStreamBuf<N>` and use both `std::array<char, N>` and `std::vector<char>` as memory storage for the underlying buffer. Only when requests size exceeds `N` should we use `std::vector<char>` (and memcpy `array`'s data to the `vector`).

Does it make sense ?

cneira:
My apologies for the delay, I totally agree with your approach as it makes sense to use a vector as fallback and keep a small non reallocating buffer as first default storage. I'll work on this change next week as I have been busy with other projects.

Bests

PR #65, opened 2017-04-06
Include serializer code in CMake install script

cyremur:
In commit 76db321 when rapidjson dependency was removed, `/include/serializer/rapidjson.h` was introduced. However, the CMakeLists.txt was not adjusted to consider subdirectories of `/include` for installation. Therefore, projects modelled based on `/examples/rest_description.cc` fail to compile because `#include "serializer/rapidjson.h"` does not work.

This fixes the error by including `/include/serializer` in the CMake installation script.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,6 +16,9 @@ include_directories (${CMAKE_CURRENT_SOURCE_DIR}/include)
 file(GLOB HEADER_FILES "include/*.h")
 install(FILES ${HEADER_FILES} DESTINATION include/pistache)
 
+file(GLOB SERIALIZER_HEADER_FILES "include/serializer/*.h")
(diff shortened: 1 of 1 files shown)


oktal:
Hello,

Thanks for the contribution, makes sense to me.

I'm just wondering if we should install the header file even though RapidJson is not found but that's not a big deal as of today, merging it anyway :+1:

cyremur:
Hi, 

I believe we rather should install the serialized code. Consider the use case that the user first installs pistache and tries out some simple examples that don't need JSON serialization and then decides to use the framework for a bigger project and installs rapidJSON afterwards. He then would have to reinstall pistache to make it work, which seems like an unnecessary source of compiler errors. 

On the other hand, I don't think there is a big downside to installing the serializer code as well even if it's unused at the time of installation.

PR #68, opened 2017-04-14
Avoid macro invocation in examples

rob-h-w:
Alpine defines macros 'major' & 'minor', so the functions modified are
parsed as macro invocations. The rename avoids this collision.


rob-h-w:
Ping - is there something wrong with this commit?

rob-h-w:
Yay! Thanks.

PR #83, opened 2017-06-01
Support requests larger than 4k

Nnamdi:
A request size of 4k is too small for some use cases.  This change makes it a runtime configurable parameter, with the option to make it unbounded, by setting the limit to 0.

There is a small buffer optimisation that will utilise the existing `ArrayStreamBuf` for requests < 4k, and the `DynamicStreamBuf` for anything larger.

### DIFF
--- a/include/common.h
+++ b/include/common.h
@@ -52,6 +52,6 @@ namespace Const {
 
     static constexpr int MaxBacklog = 128;
     static constexpr int MaxEvents = 1024;
-    static constexpr int MaxBuffer = 4096;
+    static constexpr size_t BufferSize = 4096;
     static constexpr int ChunkSize = 1024;
 }
--- a/include/endpoint.h
+++ b/include/endpoint.h
@@ -22,11 +22,13 @@ public:
         Options& threads(int val);
         Options& flags(Flags<Tcp::Options> flags);
         Options& backlog(int val);
+        Options& maxBufferSize(size_t val);
 
     private:
         int threads_;
         Flags<Tcp::Options> flags_;
         int backlog_;
+        size_t maxBufferSize_;
         Options();
     };
     Endpoint();
--- a/include/http.h
+++ b/include/http.h
@@ -670,7 +670,7 @@ namespace Private {
 
         State parse();
 
-        ArrayStreamBuf<Const::BufferSize> buffer;
+        SmallStreamBuf<Const::BufferSize> buffer;
         StreamCursor cursor;
 
     protected:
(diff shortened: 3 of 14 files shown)


oktal:
Hello,

We're discussing your PR in https://github.com/oktal/pistache/issues/87

nchiarappa:
Hi, I have the same problem: I would like to use requests larger than 4k. I read this pull and I would like to know if there are some updates regarding it, given that I read in #87 that @koallen tried the new code but found some difficulties.

jeffvandyke:
One thing I noticed relating to this:
We had this issue, so I set `MaxBuffer = 4096 * 1024 // 4M buffer`, and was going to set `ChunkSize`, but noticed that ChunkSize is not being used anywhere. As far as I know, it could be deleted.

bocse:
Is there a workaround available for uploading moderate-sized files (5-10MB) with reasonable speed (without it taking one second), without any writing to disk or chunking? (i.e. assuming there is enough RAM available)

While Pistache is very useful and versatile (especially in combination with OpenAPI v3), this issue renders it almost unusable in a production setting.

Looking forward to any suggestion.

Thank you.

dev884:
Hello. How can I enable the dynamic buffer on the client ? When I try to get more than 4k i get an empty body, no exception thrown. 

Thanks for your help

mohsenomidi:
@dennisjenkins75 

 Hi Dennis, today i faced the same limitation in server side, and i know this project was pend and there is no development on it, staying for pull request may take long time to fixing the static max size (need dynamic memory allocation in run-time). 
Many of rest requests are larger than 4K especially when the client want to send the picture or in the file operations, even the server needs to send data to client and it may exceed the 4K limitation. 

is there any plan to fix this item ?

Best Regards.

Asimov500:
This is a section of my code where you can set the maxRequestSize. I hope it helps
```
 Pistache::Address addr(Pistache::Ipv4::any(), Pistache::Port(9080));
    auto opts = Pistache::Http::Endpoint::options()
    .maxRequestSize(15000)
        .threads(10).flags(
        Pistache::Tcp::Options::ReuseAddr);
    Http::Endpoint server(addr);
    server.init(opts);
    server.setHandler(router.handler());
    server.serve();
    return 0;
```

dev884:
Thanks but i don't use the server. I'm just using the client to query an existing rest api.

dev884:
@dennisjenkins75

Hello, I have updated conf.h to have Const::MaxBuffer  = 819200.
I still cannot make query that returns payload greater than 4k.

I'm just using the client to query an existing rest api. I do not get any error or exceptions. Just an empty result.

PR #79, opened 2017-05-23
Fixed the inability to serve static files greater than 140 KB

jeffvandyke:
Added offset_ to BufferHolder so that retries would remember where they left off,
Fixed range of copying in struct Buffer so that large dynamic responses still work.

PR #85, opened 2017-06-01
Large response fix

Nnamdi:
The detach method was not properly copying data to the new buffer, and was passing the wrong length to
the Buffer object.  This resulted in corrupt data transfers when the data to be sent was too big to be sent
at once, and ::send call in Transport::asyncWriteImpl fails with EAGAIN/EWOULDBLOCK.


jeffvandyke:
I believe this change is also implemented in my pull request #79, just waiting for that to get merged in.

PR #91, opened 2017-06-13
Fix class\struct inconsistency warnings

Starl1ght:
Warnings reported by Visual Studio. Nothing serious, but still, less warnings - better.


oktal:
Hello,

Thanks for the contribution, less warnings is always a good thing

Issue #92, opened 2017-06-14
Request: Add Access-Control-Allow-Origin to include/http_header.h


cyremur:
When combining a pistache backend server with interchangable javascript frontends that may be hosted on a different URL, one needs to set `Access-Control-Allow-Origin` origin headers to allow access to resources. 

Since this is a rather common and useful header, I would like to request adding it to the default headers implemented in the standard pistache build.

Alternatively, it would be nice to be able to have a simple way to add "raw" text headers to response. I tried `response.headers().addRaw(Http::Headers::Raw("Access-Control-Allow-Origin", "*"))` 
but that didn't seem to be actually written to the output and I am not completely sure what `addRaw()` is supposed to do.

oktal:
Hello,

I initially added raw headers so that parsing would not fail when encountering a header not known to the system.

Raw headers could also be used for your particular use-case. However, raw headers are currently not serialized into the final http response sent back to the client.

I guess this is something we could also add so that you could add any header you want, even though it's not built-in into pistache. You would lose the type-safety aspect but that's a trade-of to make

cyremur:
Hi,
So for this project I just added the class `AccessControlAllowOrigin`. 20 lines of code for a simple KVP header boilerplate is actually not that bad so raw headers are definitely not a priority and probably not even needed at all.

What's your stance on including more headers into the pistache repository? Should we just PR every new header we implement in order to expand standard pistache?

On another note, is there a way to set headers for multiple URLs sharing the same basepath or description? I need to set some headers, i.e. the access control headers, globally for many URLs and it would be very convenient if you could just set headers at description level.

oktal:
> What's your stance on including more headers into the pistache repository? Should we just PR every
> new header we implement in order to expand standard pistache?

Yes, if the header you want to add is somewhat standard, feel free to submit a PR.

> On another note, is there a way to set headers for multiple URLs sharing the same basepath or description?
> I need to set some headers, i.e. the access control headers, globally for many URLs and it would be very convenient if you could just set headers at description level.

Currently, you need to add headers to every route. So you're saying that all your routes are sharing the same sets of headers and you'd like to add your set of headers at the description level so that you don't need to repeat them every-time you add a route ?

In your case, I think I'll considering creating a function that does that for you, i.e you tell it the route you want to add and it automatically adds the route to the description plus your headers.

API-wise, I'm still not sure if we should provide such a facility, feel free to discuss that

cyremur:
[...]

PR #96, opened 2017-07-02
add AccessControlAllowOrigin header implementation

cyremur:
This adds an AccessControlAllowOrigin header that works analogously to the UserAgent header.


cyremur:
Adds header implementation as discussed in issue #92.

oktal:
Hello,

I added a unit-test and merged-it in master.

Thanks for the contribution

PR #107, opened 2017-07-09
Ignore case in HTTP headers

### LINKED ISSUE #80: Json in query doesn't work
I was trying the client method, it works with normal url and query.

eg: localhost:8083/myservice?data=xyz

But if I give json value to data in the query, it does not generate a request.

Should I add any json parser before constructing the query string? Please let me know, what I am missing

### LINKED ISSUE #86: Consider making header names case insensitive
When posting a content-type header, pistache endpoint only correctly parses the header when "Content-Type" is capitalized. I am not sure about standards, but when I ran a package analysis on Swagger UI example requests http://petstore.swagger.io/, I noticed that they use "content-type" **not** capitalized.

I think `request.headers().get<Http::Header::ContentType>()` should accept both "Content-Type" and "content-type". Currently, if you use `curl localhost -X POST -H "Content-Type: application/json"` and look at the `mime()`, it is correctly recognized as `application/json`. Using the non-capitalized version however, it throws `Could not find header`.

I understand this issue might be up for discussion, but I believe being more lenient on capitalization is the more graceful thing to do here, especially if there are other finished products that you may want to use in conjunction with pistache.io and use "content-type" in their framework, since that case will make debugging and fixing the problem for "end-user" developers rather tedious.

### LINKED ISSUE #106: The request body is being ignored for some POST requests (server)
I just noted that the request body of some POST requests is being ignored by the Pistache server.

To reproduce this issue, take the [http_server](https://github.com/oktal/pistache/blob/master/examples/http_server.cc) as an example. If you dispatch a POST request with some data using curl:

```
curl -H "Content-Type: text/plain" -X POST -d "foo" http://localhost:9080/echo
```

The server responds correctly:

```
foo
```

But if you dispatch the same request using [artillery.io](https://artillery.io/docs/index.html):

```
DEBUG=http:response artillery quick http://localhost:9080/echo -p 'foo' -c 1 -q
```

The server responds with an **empty** body (the first `http:response` corresponds to the response header and the second to the response body):

```
http:response {
  "content-type": "text/plain",
  "connection": "Keep-Alive",
  "content-length": "0"
} +0ms
http:response "" +1ms
```

Am I missing something here?

danielmoraes:
Closes #106.
Closes #86.
Closes #80.
Replaces #13.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
+# Build files
+build
+
 # Compiled Object files
 *.slo
 *.lo
--- a/include/pistache/common.h
+++ b/include/pistache/common.h
@@ -54,7 +54,7 @@ namespace Const {
 
     static constexpr int MaxBacklog = 128;
     static constexpr int MaxEvents = 1024;
-    static constexpr int MaxBuffer = 4096;
+    static constexpr int MaxBuffer = 1048576;
     static constexpr int ChunkSize = 1024;
 } // namespace Const
 } // namespace Pistache
--- a/include/pistache/http_headers.h
+++ b/include/pistache/http_headers.h
@@ -6,6 +6,8 @@
 
 #pragma once
 
+#include <functional>
+#include <algorithm>
 #include <unordered_map>
 #include <vector>
 #include <memory>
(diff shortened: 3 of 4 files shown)


oktal:
Hello,

Your Pull Request has been merged in b66415aaa0dc443b7874acbbb488058b34c2dc01
I just rebased it to remove your last commit from the PR as it is not really related to the PR itself.

The MaxBuffer size problem will be addressed by a separate change.

Thanks for your contribution

PR #119, opened 2017-07-30
Prevent race between queueing req and connecting

dgreatwood:
On my application I was seeing a client hang about 1 run in 10, which I think was due to this race condition.

In client.cc, Connection::perform would check the connectionState_ and, if connected would call performImpl to perform the request; or if not yet connected, would push a request onto requestsQueue for later processing.

Meanwhile over in Connection::connect, once connection was made, connectionState_ would be changed from Connecting to Connected and then processRequestQueue would be called.

To avoid a couple of race conditions, for Connection::perform the act of checking connectionState_ and queuing request if not yet connected needs to be atomic with respect to, in Connection::connect, the act of changing connection state to Connected and calling processRequestQueue.

With these changes I have not been able to recreate the client hang.

### DIFF
--- a/include/pistache/client.h
+++ b/include/pistache/client.h
@@ -76,9 +76,18 @@ struct Connection : public std::enable_shared_from_this<Connection> {
     void connect(Address addr);
     void close();
     bool isIdle() const;
-    bool isConnected() const;
+    bool isConnected();
     bool hasTransport() const;
     void associateTransport(const std::shared_ptr<Transport>& transport);
+
+    bool isConnectedPushReqEntryIfNot(Async::Resolver & resolve,
+                                      Async::Rejection & reject,
+                                      const Http::Request& request,
+                                      std::chrono::milliseconds timeout,
+                                      const OnDone & onDone);
+
+    void setConnectedProcessRequestQueue();
+    void setConnecting();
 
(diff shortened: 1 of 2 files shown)


include/pistache/client.h
```diff
@@ -76,10 +76,19 @@ struct Connection : public std::enable_shared_from_this<Connection> {
     void connect(Address addr);
     void close();
     bool isIdle() const;
-    bool isConnected() const;
```
dennisjenkins75: You don't need to remove 'const' to lock a mutex.  Leave const but add the "mutable" attribute to the mutex definition in the class definition.

See:
  https://en.cppreference.com/w/cpp/language/cv
  https://herbsutter.com/2013/05/24/gotw-6a-const-correctness-part-1-3/

include/pistache/client.h
```diff
@@ -126,6 +135,7 @@ struct Connection : public std::enable_shared_from_this<Connection> {
     };
 
     std::atomic<uint32_t> state_;
+    std::mutex connectionStateMutex_;
```
dennisjenkins75: Make this mutex "mutable".  eg:
mutable std::mutex connectionStateMutex_;

PR #123, opened 2017-08-08
Changes to the root CMakeLists to improve out of the box experience.  

gocarlos:
* added options to the cmake file to build examples and unit tests only if needed. 
* added a fatal message in case the compiler is too old
* added prefix the path to the src and include paths in order to be able to build this as subproject.


gocarlos:
cool, all the changes that I'd like are now fixed... closing this PR

PR #124, opened 2017-08-08
Feature/some c11 improvements

gocarlos:
In this PR replaced the old functions to get the number of possible threads with c++11 ones


gocarlos:
PR #123 should be merged before this one.

PR #130, opened 2017-09-02
Fix memory leaks discovered by Valgrind

jgh-:
Fix memory leaks in two ways:

1. `freeaddrinfo` must be called after `getaddrinfo`.  

2.  The `sentinel` in `Queue<T>` was not being deleted in the destructor.  In addition to deleting the sentinel, I also added code to check if an Entry has been initialized before calling the `Entry` destructor.

### DIFF
--- a/include/pistache/mailbox.h
+++ b/include/pistache/mailbox.h
@@ -155,6 +155,10 @@ public:
     struct Entry {
         friend class Queue;
 
+        Entry () {
+            memset(&storage, 0, sizeof(storage));
+        }
+
         const T& data() const {
             return *reinterpret_cast<const T*>(&storage);
         }
@@ -164,8 +168,10 @@ public:
         }
 
         ~Entry() {
-            auto *d = reinterpret_cast<T *>(&storage);
-            d->~T();
+            if(nullptr != *reinterpret_cast<void**>(&storage)) {
+                auto *d = reinterpret_cast<T *>(&storage);
+                d->~T();    
+            }
         }
     private:
         typedef typename std::aligned_storage<sizeof(T), alignof(T)>::type Storage;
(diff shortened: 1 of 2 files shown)


sandakith:
I see similar memory issues. Would it be possible to evaluate this pull request and merge. I tried to locally merge this pull request to build and test. But I am still getting the valgrind errors at 

==10711== 1,920 bytes in 20 blocks are definitely lost in loss record 10 of 11
==10711==    at 0x4C2E1BF: operator new(unsigned long) (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
**==10711==    by 0x48769A: Pistache::Tcp::Transport::Transport(std::shared_ptr<Pistache::Tcp::Handler> const&) (in service)**
==10711==    by 0x4879A5: Pistache::Tcp::Transport::clone() const (in /_service)

include/pistache/mailbox.h
```diff
@@ -155,6 +155,10 @@ class Queue {
     struct Entry {
         friend class Queue;
 
+        Entry () {
+            memset(&storage, 0, sizeof(storage));
```
dennisjenkins75: Will this template ever be instantiated with a non-POD data type?
Consider instead:

   Entry(): storage() {}

jgh-: I'm not sure.  To be honest it's possible this PR isn't even relevant anymore, it's been almost a year.  I think you make a good point though.

include/pistache/mailbox.h
```diff
@@ -183,6 +189,7 @@ class Queue {
 
     virtual ~Queue() {
         while (auto *e = pop()) delete e;
+        delete tail; // delete sentinel.
```
dennisjenkins75: I see, 'pop()' won't pop the sentinel.  ACK.

PR #102, opened 2017-07-04
Support musl based compiler

xoac:
Hi,
I am doing a application that use a [musl](https://www.musl-libc.org/) as a standard library, There is a some bug in `strptime` that not correctly allow parse RFC1123 Data. So I changed it to `std::get_time` That make it works on musl and glibc.

I don't know why std::get_time can't parse `ASCI C asctime() format`. In this way I used old `strptime` (this is correct also for musl). Here is [some](https://stackoverflow.com/questions/44901711/convert-ansi-cs-asctime-format-using-stdget-time/44902196#44902196) that can cause this. It's probably bug in GCC.

Pass tests:
- gcc version 7.1.1 20170528 (GCC) (glibc)
- gcc version 6.3.0 (GCC) (musl)

PR #131, opened 2017-09-04
Support musl based compilers

xoac:
This is exactly #102 but old was on my master forked branch. And I messed that branch up.

Pass tests:
- gcc version 7.1.1 20170528 (GCC) (glibc)
- gcc version 6.3.0 (GCC) (musl)

Require:
-gcc >= 5.1


xoac:
Close. Reason: #132 was merger.

PR #138, opened 2017-09-14
Re-use fixed-size recv buffer in client for larger receives.

### LINKED ISSUE #116: Working data buffer too small / not correctly re-used in Transport::handleIncoming
In client.cc, Transport::handleIncoming, a buffer size Const::MaxBuffer is used (4KB). I'm looking at an example from a third-party service where the buffer would need to be 18K to accommodate the response.

I'm not sure that buffer really does need to be bigger, however. After all, following each "recv", handleResponsePacket can be called to do deal with data gathered up to that point.

I tried adding a buffer reset thus:

```C++
for (;;) {

        ssize_t bytes;

        if (totalBytes + 32 > Const::MaxBuffer)
        { // <= 32 bytes remaining in buffer, we reset buffer usage
            handleResponsePacket(connection, buffer, totalBytes);
            
            totalBytes = 0;
            memset(buffer, 0, sizeof buffer);
        }

        bytes = recv(connection->fd, buffer + totalBytes,
                     Const::MaxBuffer - totalBytes, 0);
...
```
This does fix the issue in my case, though I'm unclear how generally correct it is (if at all).

dgreatwood:
Avoids client-side hangs/crashes with larger responses
Fixes #116

### DIFF
--- a/src/client/client.cc
+++ b/src/client/client.cc
@@ -331,7 +331,17 @@ Transport::handleIncoming(const std::shared_ptr<Connection>& connection) {
 
         ssize_t bytes;
 
-        bytes = recv(connection->fd, buffer + totalBytes, Const::MaxBuffer - totalBytes, 0);
+        if (totalBytes + 32 > Const::MaxBuffer)
+        { // <= 32 bytes remaining in buffer, we reset buffer usage
+            handleResponsePacket(connection, buffer, totalBytes);
+            
+            totalBytes = 0;
+            memset(buffer, 0, sizeof buffer);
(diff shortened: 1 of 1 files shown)


dgreatwood:
That "if" condition could should not be met. Only if recv incorrectly
returns a value greater than *length* = Const::MaxBuffer - totalBytes; it's
just defensive.

Yes, ">" would make sense.

xoac:
@dgreatwood  You responded using e-mail. I have edited my question in the meantime.

PR #144, opened 2017-09-17
Ssl client for openssl

dgreatwood:
Basic SSL client, using openssl
Avoids changing the existing core polling loop of Pistache, instead doing its own selects and retries for blocking write.
User should call Pistache::Http::Connection::setHostChainPemFile(<filepath-and-name>), with a suitable cert "pem" file, before using https. One method of getting a cert file that will likely serve is here: https://curl.haxx.se/docs/caextract.html; you would also want a way to keep the cert file up up to date in that case.
This SSL client appears to work correctly against the both the Azure and Google cloud storage web APIs. It has not been tested otherwise.
To use for Azure and Google, other fixes are also needed: pull requests 142, 141, 140, 139, 138, 119 and 118.
To disable SSL support in the source, you could make with PIST_EXCLUDE_SSL set, or edit ssl_do_or_dont.h to comment out PIST_INCLUDE_SSL define. You could also edit src/CMakeLists.txt to remove crypto and ssl from target_link_libraries, if not needed. TO DO: Make the later dependent on make configuration.

### DIFF
--- a/include/pistache/client.h
+++ b/include/pistache/client.h
@@ -165,7 +165,7 @@ struct Connection : public std::enable_shared_from_this<Connection> {
 
     void close();
     bool isIdle() const;
-    bool isConnected();
+    bool isConnected() const;
     bool hasTransport() const;
     void associateTransport(const std::shared_ptr<Transport>& transport);
 
@@ -233,7 +233,7 @@ private:
     };
 
     std::atomic<uint32_t> state_;
-    std::mutex connectionStateMutex_;
+    mutable std::mutex connectionStateMutex_;
     ConnectionState connectionState_;
     std::shared_ptr<Transport> transport_;
     Queue<RequestData> requestsQueue;
--- a/include/pistache/client.h
+++ b/include/pistache/client.h
@@ -107,9 +107,6 @@ struct ConnectionEntry {
         #endif
 };
 
-typedef std::function<void()> OnDone;
-
-
 struct Connection : public std::enable_shared_from_this<Connection> {
 
     friend class ConnectionPool;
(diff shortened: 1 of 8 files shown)


include/pistache/client.h
```diff
@@ -126,6 +233,7 @@ struct Connection : public std::enable_shared_from_this<Connection> {
     };
 
     std::atomic<uint32_t> state_;
+    std::mutex connectionStateMutex_;
```
xoac: I would suggest make it mutable. This alow function `isConnected` still be const.

dgreatwood: Agreed, done

partidor:
Any interest in merging this PR @oktal?

dgreatwood:
> Ping. This PR is rather large, implements a requested feature, but has merge conflicts and lacks unit tests. I'd happily merge it if it has unit tests and were mergable. However, I intent to close this PR in a few weeks if not ready. We're clearing out stale PRs. I would love to merge this one though. :)

Dennis - thanks for picking up "pistache"; it would be good if there were a well done, non-GPL, C/C++ HTTP library IMHO.

FWIW, I did just confirm that this PR's code still runs successful against the Google and Azure storage APIs.

Unfortunately, I am too busy with other work to update the PRs and/or look at unit tests. My apologies. If that changes, I may come back and revive these PRs; I'd likely reach out to you at that time.

This applies to my other PRs on the project as well. If I may, I would suggest you look at the other ones at least, which are much simpler; they deal with some fairly basic bugs in the original code like running out of buffer space, so far as I recall.

Meanwhile - good luck!
-DG

PR #145, opened 2017-09-20
Add operator!= for MediaType

xoac:
Just add function for cmp MediaTypes to avoid  `!((MediaType)m1 == (MediaType)m2)`


oktal:
Hello.

Thanks for the contribution. It indeed does make sense to implement `operator!=` for MediaType, merging it in.

PR #149, opened 2017-10-09
BUGFIX: Fix warnings from GCC and Clang

jpihl:
Hello,

This PR is basically a duplicate of #20, but it seems the work there has stalled a bit. I tried looking at your suggestions to @dan-42 and address the warnings accordingly.

Let me know if you think something needs to be handled differently.


jpihl:
Hi @oktal,

I know you are busy, but it would be great to get this PR merged. The warnings seems to be problematic for many people.

markusgau:
Could this be rebase once more mabye? There are still a lot of issues when building for ARM64 with -pedantic and -Wextra enabled.
I think chances are that this might get merged then. The new maintainer seems much more responsive...

examples/hello_server.cc
```diff
@@ -1,33 +1,38 @@
...
-
-using namespace Pistache;
-
-class HelloHandler : public Http::Handler {
+class HelloHandler : public Pistache::Http::Handler {
 public:
 
     HTTP_PROTOTYPE(HelloHandler)
 
-    void onRequest(const Http::Request& request, Http::ResponseWriter response) {
```
dennisjenkins75: Please do not reformat the code when making changes unrelated to formatting.

Its fine that you "(void)request;", but there was no need to split line 17 into 4 lines. 

I am heavily inclined to NACK your PR as-is, as it introduces many unnecessary diffs.

jpihl: Sorry about that. The initial PR was a bit cleaner and I may have been a bit careless during this second "update". Give me a few minutes and I'll fix the worst parts.

jpihl:
@dennisjenkins75 I tried cleaning up the PR a bit. I've used the `UNUSED` macro in favor of omitting the names of unused parameters. There's a lot of tailing white space which have been removed but my life is too short for reverting that - I hope you can see past it.

examples/hello_server.cc
```diff
@@ -15,19 +15,24 @@ class HelloHandler : public Http::Handler {
...
+    Pistache::Http::Endpoint server(addr);
     server.init(opts);
-    server.setHandler(Http::make_handler<HelloHandler>());
-    server.serve();
+    server.setHandler(Pistache::Http::make_handler<HelloHandler>());
+
+    std::cout << "Starting server. Test with the following command: "
+              << "\"curl http://127.0.0.1:" << addr.port() << "\"\n";
 
+    std::cout << "Press Enter to Exit" << '\n';
```
dennisjenkins75: Why did you change the runtime behavior of this test program?

examples/hello_server.cc
```diff
@@ -15,19 +15,24 @@ class HelloHandler : public Http::Handler {
     HTTP_PROTOTYPE(HelloHandler)
 
     void onRequest(const Http::Request& request, Http::ResponseWriter response) {
-        response.send(Http::Code::Ok, "Hello World");
+        (void) request;
```
dennisjenkins75: Did you want to change this to "UNUSED(request)" as well?

include/pistache/client.h
```diff
@@ -136,8 +136,8 @@ struct Connection : public std::enable_shared_from_this<Connection> {
     Private::Parser<Http::Response> parser_;
 };
 
-struct ConnectionPool {
-
+class ConnectionPool {
```
dennisjenkins75: Why did you make this change?  What bug/problem does it address?  A class w/ all public methods/member vars is the same as a struct, so this change should be a no-op... but why make it?

include/pistache/common.h
```diff
@@ -37,7 +37,7 @@
 #define TRY_RET(...) \
     [&]() { \
         auto ret = __VA_ARGS__; \
-        if (ret < 0) { \
+        if ((ssize_t)ret < 0) { \
```
dennisjenkins75: Please use "static_cast<ssize_t>(ret)" instead of the older C-style casting.  Old-style casts are flagged as warnings when compiled with "-Weffc++".

include/pistache/http.h
```diff
@@ -549,7 +547,7 @@ class ResponseWriter : public Response {
        return &buf_;
     }
 
-    DynamicStreamBuf *rdbuf([[maybe_unused]] DynamicStreamBuf* other) {
```
dennisjenkins75: Why are you removing the "[[maybe_unused]]"?  It was added only 28 days ago with PR #259.

include/pistache/http.h
```diff
@@ -613,31 +611,35 @@ namespace Private {
         Message *message;
     };
 
-    struct RequestLineStep : public Step {
```
dennisjenkins75: Why are you changing the struct to a class?  What warning (and in what compiler version) is this change addressing?

src/common/http_defs.cc
```diff
@@ -62,9 +62,10 @@ CacheDirective::delta() const
             return std::chrono::seconds(data.maxStale);
         case MinFresh:
             return std::chrono::seconds(data.minFresh);
+        default:
+            throw std::domain_error("Invalid operation on cache directive");
+            return std::chrono::seconds();
```
dennisjenkins75: This return will never execute.  Move it outside of the switch statement?  It won't execute their either...  I'm surprised that the compiler is not throwing a warning about missing a return statement.  Must be smart enough to realize that all cases are covered and that all will exit the method before the normal exit at the bottom.

I'm ambivalent on this one.  It just looks weird I guess.

src/common/os.cc
```diff
@@ -19,7 +19,7 @@ using namespace std;
 
 namespace Pistache {
 
-int hardware_concurrency() {
+uint hardware_concurrency() {
```
dennisjenkins75: https://en.cppreference.com/w/cpp/thread/thread/hardware_concurrency

std::thread::hardware_concurrency() returns an int.  I agree that it _should be_ an unsigned int, but the c++11 standard has it signed.

In fact, consider removing the entire body of this method and just "return std::thread::hardware_concurrency();".

src/common/reactor.cc
```diff
@@ -55,8 +55,8 @@ struct Reactor::Impl {
 /* Synchronous implementation of the reactor that polls in the context
  * of the same thread
  */
-struct SyncImpl : public Reactor::Impl {
-
+class SyncImpl : public Reactor::Impl {
```
dennisjenkins75: Why change struct to class here?

src/common/stream.cc
```diff
@@ -71,7 +71,7 @@ DynamicStreamBuf::reserve(size_t size)
 
 bool
 StreamCursor::advance(size_t count) {
-    if (count > buf->in_avail())
+    if ((ssize_t)count > buf->in_avail())
```
dennisjenkins75: Please use a C++ static_cast instead of a C-Style cast.

src/server/listener.cc
```diff
@@ -220,7 +222,7 @@ Listener::run() {
                 else {
                     if (event.flags.hasFlag(Polling::NotifyOn::Read)) {
                         auto fd = event.tag.value();
-                        if (fd == listen_fd)
+                        if ((ssize_t)fd == listen_fd)
```
dennisjenkins75: Listener::listen_fd is an int (as it should be for file descriptors).  Just replace lines 224,255 with:

if (static_cast<int>(event.tag.value()) == listen_fd)

Unless "value()" already returns an int, then omit the static_cast.

examples/hello_server.cc
jpihl: Sure. I used the void cast here because it's an example and I thought the `UNUSED` macro might confuse new users.

include/pistache/client.h
jpihl: The reason for this change is that ConnectionPool is forward declared as a class - which gives a warning. Alternatively I can revert this and forward declare it as a struct instead.

include/pistache/common.h
jpihl: I just tried removing the cast all together and I don't see any warnings with neither g++ or clang - I guess that's a better fix.

include/pistache/http.h
jpihl: I got warning for this. And then I read this comment also added in that PR:
https://github.com/oktal/pistache/pull/259/files#diff-5a4ac07535107245765a4a8d3f822a83R52

include/pistache/http.h
jpihl: This is addressing the same issue as the `ConnectionPool`.

src/common/http_defs.cc
jpihl: You are right, it looks weird. I just tried removing the return statement and I no warning appeared. I'll remove the return statement.

src/common/os.cc
jpihl: Maybe I'm missing something, but in the link it's specified that `std::thread::hardware_concurrency()` returns an unsigned int?

I'll change the implementation to use `std::thread::hardware_concurrency()`.

src/common/reactor.cc
jpihl: Again same as `ConnectionPool`.

src/server/listener.cc
jpihl: `Tag::value()` returns a `uint64_t`, so the cast is necessary.

examples/hello_server.cc
jpihl: This change has nothing to do with the purpose of this PR. I'll revert it if you like.

I changed the it so that it's possible to stop the execution, this makes it possible to use tool like valgrind to check for memory leaks.

jpihl:
I don't currently have access to GCC-7.3.0 nor Clang-6.0.1.

examples/hello_server.cc
dennisjenkins75: Well, the UNUSED macro is used throughout the rest of the code, and imho, it is kinda self-explanatory.  Please change this to use the macro.  My main goal here is to remove the C-style cast.

examples/hello_server.cc
dennisjenkins75: Please revert this change.

examples/http_server.cc
```diff
@@ -140,6 +140,7 @@ class MyHandler : public Http::Handler {
     }
 
     void onTimeout(const Http::Request& req, Http::ResponseWriter response) {
+        (void) req;
```
dennisjenkins75: Please remove the C-style cast.  Replace w/ UNUSED() macro or c++ style cast.

include/pistache/client.h
dennisjenkins75: Ah, I see.  Either direction (all struct or all class) is fine then.

include/pistache/http.h
dennisjenkins75: I do not want to require c++17 any time soon.  So for the sake of consistency, removing this is fine.  Please find the other references are remove them as well (there is only one other, in that comment).

jpihl:
I hope my latest changes has addressed all your comments. I've tested the compilation of this with the following compilers:
g++ (Ubuntu 7.3.0-16ubuntu3) 7.3.0
clang version 6.0.0-1ubuntu2 (tags/RELEASE_600/final)

PR #153, opened 2017-10-19
Add application/octet-stream to be type-safe

xoac:
Start here https://github.com/oktal/pistache/blob/b66415aaa0dc443b7874acbbb488058b34c2dc01/src/common/http.cc#L710 :
```
auto mime = Mime::MediaType::fromFile(fileName);
        if (mime.isValid())
            setContentType(mime);
```
If we don't know what this file is we can always use application/octet-stream?!

### DIFF
--- a/src/common/http.cc
+++ b/src/common/http.cc
@@ -671,16 +671,16 @@ serveFile(ResponseWriter& response, const char* fileName, const Mime::MediaType&
 
     int fd = open(fileName, O_RDONLY);
     if (fd == -1) {
-        if(errno == ENOENT)
-            throw HttpError(Http::Code::Not_Found, "");
+        std::string str_error(strerror(errno));
+        if(errno == ENOENT) {
+            throw HttpError(Http::Code::Not_Found, std::move(str_error));
+        }
         //eles if TODO 
         /* @Improvement: maybe could we check for errno here and emit a different error
             message
         */
-        else
-        {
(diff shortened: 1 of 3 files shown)


src/common/http.cc
```diff
@@ -671,10 +671,17 @@ serveFile(ResponseWriter& response, const char* fileName, const Mime::MediaType&
...
+            throw HttpError(Http::Code::Not_Found, "");
+        //eles if TODO 
         /* @Improvement: maybe could we check for errno here and emit a different error
             message
         */
-        throw HttpError(Http::Code::Not_Found, "");
+        else
+        {
+            std::string str_error(strerror(errno));
+            HttpError(Http::Code::Internal_Server_Error, str_error);
```
oktal: This expression does not have any effect, I guess you want to `throw` here ?

src/common/http.cc
```diff
@@ -671,10 +671,17 @@ serveFile(ResponseWriter& response, const char* fileName, const Mime::MediaType&
 
     int fd = open(fileName, O_RDONLY);
     if (fd == -1) {
+        if(errno == ENOENT)
+            throw HttpError(Http::Code::Not_Found, "");
```
oktal: We should also provide an error message. `strerror` should be enough

xoac:
Thanks for review! Now should be fine.

PR #154, opened 2017-10-22
[Client] Inflightqueue change from dequeue to queue and add pop()

### LINKED ISSUE #117: std::bad_function thrown when attempting to handle response at client
I was seeing std::bad_function thrown in client.cc.

Catching it in the debugger, it was caused by onDone (callback address) being NULL. Seems to happen when the server side decides to close a connection (perhaps after a period of inactivity?).

On further investigation, the "response" (i.e. the close connection) is happening when inflightRequests is empty.

The current code is doing stuff like:
```C++
Connection::handleResponsePacket(const char* buffer, size_t bytes) {
        ...
        auto req = std::move(inflightRequests.front());
        inflightRequests.pop_back();

        ...

        req.resolve(std::move(parser_.response));
        req.onDone();
        ...
}
```

According to C++ documentation, "Calling front on an empty container is undefined" for std::deque (inflightRequests is a std::deque). What actually seems to happen is that front returns a default-constructed request - which has null onDone.

We could guard all such cases in client.cc by checking for empty inflightRequests and for null onDone, something like this:
```C++
Connection::handleResponsePacket(const char* buffer, size_t bytes) {
        ...
        if (!inflightRequests.empty())
        {
        auto req = std::move(inflightRequests.front());
        inflightRequests.pop_back();

        ...

        req.resolve(std::move(parser_.response));
        if (req.onDone)
                req.onDone();
        ...
        }
}
```
Doing this in the client.cc places where we're pulling from inflightRequests fixes the problem for me.

### LINKED ISSUE #128: Maybe these should be pop_front?
https://github.com/oktal/pistache/blob/b66415aaa0dc443b7874acbbb488058b34c2dc01/src/client/client.cc#L495

The previous line calls deque::front()

Correspondingly pop_front() should be called instead of pop_back() right?

xoac:
Resolve #128
I also add changed proposed in #118 PR that resolve #117.

### DIFF
--- a/include/pistache/client.h
+++ b/include/pistache/client.h
@@ -6,7 +6,7 @@
 
 #pragma once
 #include <atomic>
-#include <deque>
+#include <queue>
 
 #include <sys/types.h>
 #include <sys/socket.h>
@@ -130,7 +130,7 @@ private:
     std::shared_ptr<Transport> transport_;
     Queue<RequestData> requestsQueue;
 
-    std::deque<RequestEntry> inflightRequests;
+    std::queue<RequestEntry> inflightRequests;
 
     TimerPool timerPool_;
     Private::Parser<Http::Response> parser_;
--- a/src/client/client.cc
+++ b/src/client/client.cc
@@ -464,7 +464,7 @@ Connection::handleResponsePacket(const char* buffer, size_t bytes) {
     parser_.feed(buffer, bytes);
     if (parser_.parse() == Private::State::Done) {
         auto req = std::move(inflightRequests.front());
-        inflightRequests.pop_back();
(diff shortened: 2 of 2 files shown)


xoac:
>  QQ: What motivated you to change from a deque to a queue in this PR?

pistache wasn't using any API from dequeu that is not available in queue.

This PR is important in some places:
https://github.com/oktal/pistache/pull/154/files#diff-61294b3181ed7aff7ec6d981a1e8710fR503
https://github.com/oktal/pistache/pull/154/files#diff-61294b3181ed7aff7ec6d981a1e8710fR486

I don't use pistache anymore. Fill free to push to this PR or close.

PR #157, opened 2017-10-28
added an overlaod for match_string

voldyman:
This change makes it simpler to deduce string length when it's possible and void a macro.


voldyman:
Updated, thankfully the tests caught a bug along the way!

PR #159, opened 2017-11-07
Add versionString function at http_defs.h header file

pantuza:
Hi,

The function ```versionString()``` is defined at ```http_defs.cc``` but is not available for use through the ```http_defs.h``` header file. 

My personal use case with ```versionString()``` is for logging purpose.

Regards.


oktal:
Hello,

It perfectly makes sense, I think it was an oversight.

Thanks for the contribution !

PR #162, opened 2017-11-16
Avoid race condition when handling timer timeouts

ywahl:
When using timeouts:
e.g.: response.timeoutAfter(std::chrono::milliseconds(100));
we get sometimes timer already armed error followed by a SEGV+core dump (running run_http_server)

The problem is apparently due to a race condition between the time the timerfd is closed, and the time it is erased from the timers map.

The timerfd is first closed, and then erased from the map,
If we have the following sequence a problem can happen (it happens almost every time in my tests when load gets a bit high)
e.g.:
thread 1 close(timerfd) (http.h:) around line 210
thread 2 void arm(Duration duration) { (http.h)
Async::Promise<uint64_t> p([=](Async::Deferred<uint64_t> deferred) {
timerFd = TRY_RET(timerfd_create(CLOCK_MONOTONIC, TFD_NONBLOCK));

it reuses the closed the timerfd
thread 1: timers.erase(it) (transport.cc line 85)

We should actually close the timerfd only after doing timers.erase(it)

### DIFF
--- a/include/pistache/http.h
+++ b/include/pistache/http.h
@@ -209,7 +209,6 @@ public:
             [=](uint64_t numWakeup) {
                 this->armed = false;
                 this->onTimeout(numWakeup);
-                close(timerFd);
         },
         [=](std::exception_ptr exc) {
             std::rethrow_exception(exc);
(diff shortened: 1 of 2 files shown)


1Hyena:
@ywahl 
I just discovered this awesome lib recently and I have run into a question regarding timers. how do you externally fill a request that already has returned but has a timer running?

I'm making another curl request from the request handler and since the curl request blocks the active thread the timeout will never occur. any ideas for a fix?

PR #165, opened 2017-11-21
fixing hello_server.cc with namespace updates from Net to Pistache

sandakith:


oktal:
Hello,

Thanks for the contribution, merging it in.

PR #172, opened 2017-12-06
Query in client

xoac:
The query is only defined for Server purpose. I added ~`str()`~ `as_str()` method that generate part of url.

U can now create a own Clinet Wraper for some API with easy query parametes (they can by create from initialization list).

If sb has idea how make them compile-time safe in client form don't hesitate.

### DIFF
--- a/src/common/http.cc
+++ b/src/common/http.cc
@@ -500,6 +500,16 @@ namespace Uri {
 
         return Some(it->second);
     }
+    
+    std::string 
+    Query::str() const {
+        std::string query_url;
+        for(const auto &e : params)
+            query_url += "&" + e.first + "=" + e.second;
+        if(not query_url.empty()) // if is not empty
+            query_url[0] = '?';
+        return query_url;
+    }
 
     bool
     Query::has(const std::string& name) const {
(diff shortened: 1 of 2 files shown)


src/common/http.cc
```diff
@@ -500,6 +500,18 @@ namespace Uri {
 
         return Some(it->second);
     }
+    
+    std::string 
+    Query::as_str() const {
+        std::string query_url;
+        for(const auto &e : params) {
+            query_url += "&" + e.first + "=" + e.second;
```
dennisjenkins75: This does not properly / safely escape e.first and e.second.

Pistache::Uri::Query::add() is public and can does not sanitize these values when adding them to the params map.  So it is incumbent on the caller to guarantee that they properly quote/escape 'name' and 'value'.  IMHO, this is not safe.

(I just read all of http.h and http.cc).  It appears that pistache does not URL query parameter sanitization or proper escaping / unescaping.  It seems that pistache requires the caller to pack/unpack quoted URIs.  So this code change is reasonable.  :(

https://en.wikipedia.org/wiki/Query_string

dennisjenkins75: xoac, I will merge this PR.  At first glance I was a bit worried about how the URL was constructed without escaping the values, then I see that Pistache never does that anyway.

PR #180, opened 2018-01-04
Make mime case - insesitive #179

### LINKED ISSUE #179: MIME type should be case-insensitive
I was writing an application and test it with `Content-type: Application/Json` that give me an error of unsupported media type. But `Content-type: application/json`. 

So i was looking for more info:
 * https://stackoverflow.com/questions/7718476/are-http-headers-content-type-c-case-sensitive

>  For example, TEXT, Text, and TeXt are all equivalent. 
from https://www.w3.org/Protocols/rfc1341/4_Content-Type.html

Here is a problem:
https://github.com/oktal/pistache/blob/master/src/common/mime.cc#L122-L130

This should be probably `match_string(s, sizeof s - 1, cursor)`. . .

xoac:
close #179

### DIFF
--- a/src/common/mime.cc
+++ b/src/common/mime.cc
@@ -120,7 +120,7 @@ MediaType::parseRaw(const char* str, size_t len) {
     // Watch out, this pattern is repeated throughout the function
     do {
 #define TYPE(val, s)                              \
-        if (match_raw(s, sizeof s - 1, cursor)) { \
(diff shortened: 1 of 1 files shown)


oktal:
Oups, I did not see that you submitted a Pull Request for this issue. I did the exact same change and also added some unit-tests that validate the change.

Thanks for reporting the issue and attempting it fix it anyway !

PR #192, opened 2018-02-13
make more friendly to embed in cmake projects

tropf:
This approaches two problems that appear when using pistache in another project. The (most easy) way to do so is by adding pistache as a git submodule and then including it to cmake using the following snippet.

> add_subdirectory("${PROJECT_SOURCE_DIR}/extern/path/to/pistache")

When doing so, pistache creates an unnecessary amount of clutter in the form of building tests & examples.

# Building Tests & Examples by default
In the root `CMakeLists.txt` options for not building the tests & examples are introduced. By default these options are disabled so neither tests nor examples are built.

To build them (e.g. for CI), invoke cmake like this.

> $ mkdir build && cd build
> $ cmake -DPISTACHE_BUILD_TESTS=ON -DPISTACHE_BUILD_EXAMPLES=ON ..
> $ make -j4

# Install Target
When invoking `sudo make install` pistache is installed (even when the parent project only links to pistache statically).

In 9/10 cases this behaviour is expected and wanted, but in the other 10% of cases installing pistache can be disabled by overwriting the `PISTACHE_INSTALL` option.

> There is no change in the default `sudo make install` behaviour.


oktal:
Hello.

Thanks for the contribution. I agree with you, pistache's CMake should also be modernized https://pabloariasal.github.io/2018/02/19/its-time-to-do-cmake-right/

I have to confess that I'm very bad with CMake, so I'm very glad you guys are taking some time to help with that.

I'm trusting you on that one, merging the PR.

Issue #160, opened 2017-11-13
Couldn't process http request which header and body in two tcp packet


bxq2011hust:
Recently I use Java send http request, but Pistache doesn't response often. While I use python or node send http request, it response every time. I ues wireshark capture tcp packet, finally I found where is wrong. If a http request's heder and `\r\n` is sent in first tcp packet and it's body is in second tcp packet, `message->body_` will be empty after `BodyStep::apply`, then the code in section-1 will execute twice which makes `available == size - 2` in section-2 , this request will not go to next step. 
- section-1
https://github.com/oktal/pistache/blob/b66415aaa0dc443b7874acbbb488058b34c2dc01/src/common/http.cc#L324-L327
- section-2
https://github.com/oktal/pistache/blob/b66415aaa0dc443b7874acbbb488058b34c2dc01/src/common/http.cc#L349-L361

Kbij:
I'm having the same problem. Did a quick fix (don't know if it is 100% correct)
in http.cc:
    State
    BodyStep::apply(StreamCursor& cursor) {
        if (message->body_.empty()) {
            /* If this is the first time we are reading the body, skip the CRLF */
            if (!cursor.advance(2)) return State::Again;

            bytesSkipped += 2;
        }
And also:
        auto readBody = [&](size_t size) {
            StreamCursor::Token token(cursor);
            const size_t available = cursor.remaining();
            // We have an incomplete body, read what we can
            if (available < (size - bytesSkipped)) {
                cursor.advance(available);
                message->body_.append(token.rawText(), token.size());
            //added
                bytesRead += available;

                return false;
            }
         //changed
            cursor.advance(available);
            message->body_.append(token.rawText(), token.size());
            return true;
        };

bxq2011hust:
@Kbij I think your fix is fine. I fix it by add a bool var to sign if body has been read.
https://github.com/bxq2011hust/pistache/blob/3af3cb4da800b25d682db4b0d105303cf083dd5a/src/common/http.cc#L324

liujoey:
I had the same issue with python3 and requests/http.client library. It was working fine in Postman, but requests/urllib3 got response once in a while only.

Thank you so much @bxq2011hust, your finding and fix just saved me hours or even days!

xoac:
@bxq2011hust  could u create PR? This would be great!

PR #194, opened 2018-02-19
Don't skip CRLF twice in BodyStep::apply

RavenhallP:
#160 

When the request header and body arrive in separate TCP packets, this can cause BodyStep::apply to attempt to skip the CRLF twice. This may result in the message not being processed, as the remaining content is not as long as the Content-Length.

To fix this problem, added a bool variable BodyStep::crlfSkipped to keep track of whether the CRLF has already been skipped.

### DIFF
--- a/include/pistache/http.h
+++ b/include/pistache/http.h
@@ -615,6 +615,7 @@ namespace Private {
             : Step(message)
             , chunk(message)
             , bytesRead(0)
+            , crlfSkipped(false)
         { }
 
         State apply(StreamCursor& cursor);
@@ -647,6 +648,7 @@ namespace Private {
 
         Chunk chunk;
         size_t bytesRead;
+        bool crlfSkipped;
     };
 
     struct ParserBase {
--- a/src/common/http.cc
+++ b/src/common/http.cc
@@ -321,25 +321,33 @@ namespace Private {
 
     State
     BodyStep::apply(StreamCursor& cursor) {
-        if (message->body_.empty()) {
+        if (!crlfSkipped) {
             /* If this is the first time we are reading the body, skip the CRLF */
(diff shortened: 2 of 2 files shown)


oktal:
Hello.

So I took a look at the issue and was able to reproduce it in a unit-test. I also checked your PR which seems to fix the issue.

However, I think I have an other idea that would completely eliminate the need for  the `crlfSkipped`.
The idea would be to only consider the headers parsing phase done when we read the `CRLF` prior to the body. That way, we also do not need to check for the body's empty-ness in the `BodyStep::apply`.

Here is the patch:

```diff
diff --git a/src/common/http.cc b/src/common/http.cc
index 1238b9e..e0e27c6 100644
--- a/src/common/http.cc
+++ b/src/common/http.cc
@@ -315,17 +315,14 @@ namespace Private {
             headerRevert.ignore();
         }
 
+        if (!cursor.advance(2)) return State::Again;
+
         revert.ignore();
         return State::Next;
     }
 
     State
     BodyStep::apply(StreamCursor& cursor) {
-        if (message->body_.empty()) {
-            /* If this is the first time we are reading the body, skip the CRLF */
-            if (!cursor.advance(2)) return State::Again;
-        }
-
         auto cl = message->headers_.tryGet<Header::ContentLength>();
         auto te = message->headers_.tryGet<Header::TransferEncoding>();
```

Could you please try it and let me know if it fixes the issue on your side ? The only downside I see for this solution is that we might have to parse headers again if we receive them in a TCP packet without the final CRLF, which is a little bit pessimistic.

RavenhallP:
Hi. I've tested your patch, and it fixes the issue for me. It also seems like a nicer solution than mine. Thanks for responding so quickly.

oktal:
Patch has been pushed, closing this PR.

PR #198, opened 2018-02-25
Client example : Wait for response and print number of failed requests

VishwanathPatil:


examples/http_client.cc
```diff
@@ -49,12 +49,13 @@ int main(int argc, char *argv[]) {
                std::cout << "Response body = " << body << std::endl;
         }, Async::IgnoreException);
         responses.push_back(std::move(resp));
-    }
-
-    auto sync = Async::whenAll(responses.begin(), responses.end());
-    Async::Barrier<std::vector<Http::Response>> barrier(sync);
+		auto sync = Async::whenAll(responses.begin(), responses.end());
```
xoac: Just formatting issue. This would look better if would be align to left. Like `responses.push_back(std::move(resp));` line above.

VishwanathPatil: Thanks , Done

PR #200, opened 2018-03-04
Getting rid of compiler warnings about initialization order

JenniferBuehler:
Good work on this package!
This PR is pretty trivial, only getting rid of compiler warnings about initialization order.

### DIFF
--- a/include/pistache/async.h
+++ b/include/pistache/async.h
@@ -1,6 +1,6 @@
 /* async.h
    Mathieu Stefani, 05 novembre 2015
-   
+
   This header brings a Promise<T> class inspired by the Promises/A+
   specification for asynchronous operations
 */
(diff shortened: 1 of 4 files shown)


VishwanathPatil:
There is one more place same warning exist,
diff --git a/src/common/peer.cc b/src/common/peer.cc
index 87ba9c6..16fcb05 100644
--- a/src/common/peer.cc
+++ b/src/common/peer.cc
@@ -18,14 +18,14 @@ namespace Tcp {
 using namespace std;
 
 Peer::Peer()
-    : fd_(-1)
-    , transport_(nullptr)
+    : transport_(nullptr)
+    , fd_(-1)
 { }
 
 Peer::Peer(const Address& addr)
-    : addr(addr)
+    : transport_(nullptr)
+    , addr(addr)
     , fd_(-1)
-    , transport_(nullptr)
 { }
 
 Address

JenniferBuehler:
I found a few more after re-compiling, see #201

xoac:
Didn't you duplicate #149?

JenniferBuehler:
Oh I didn't see someone already requested the changes... yeah that probably covers the same.

PR #202, opened 2018-03-08
Added support to get parameters from Query

JenniferBuehler:
I need access to the parameter values which were sent along with the request. Would it be OK to add access to the key values like this?

Alternatively,  access to the whole map would also work, I just thought because the values can already be accessed by get() then it would be at least helpful to get the set of keys as well.

### DIFF
--- a/include/pistache/date.h
+++ b/include/pistache/date.h
@@ -1,0 +1,7878 @@
+#ifndef DATE_H
+#define DATE_H
+
+// The MIT License (MIT)
+//
+// Copyright (c) 2015, 2016, 2017 Howard Hinnant
+// Copyright (c) 2016 Adrian Colomitchi
+// Copyright (c) 2017 Florian Dang
+// Copyright (c) 2017 Paul Thompson
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+//
+// Our apologies.  When the previous paragraph was written, lowercase had not yet
+// been invented (that would involve another several millennia of evolution).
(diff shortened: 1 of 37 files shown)


oktal:
Hello, thanks for the contribution !

I agree with you. I think we could have both. If we don't want to directly expose the `Query`'s underlying map, maybe could we add iterator support ? So that you could at least iterate over the list of all pairs ?

JenniferBuehler:
For sure, that would work too. I already thought that exposing is not desirable, which is why in this PR a copy of the parameters is returned. But providing const iterators would certainly be an alternative solution. Would you like me to modify this PR with iterators instead?

JenniferBuehler:
Yes I will modify the PR, hopefully will get around to it in the next couple of days, will put it on top of my TODO  list :)

JenniferBuehler:
Added access to `begin()` and `end()` iterators, is it something like this you had in mind?

include/pistache/date.h
```diff
@@ -4370,7 +4371,7 @@ scan_keyword(std::basic_istream<CharT, Traits>& is, FwdIter kb, FwdIter ke)
     const unsigned char does_match = '\2';
     unsigned char statbuf[100];
     unsigned char* status = statbuf;
-    unique_ptr<unsigned char, void(*)(void*)> stat_hold(0, free);
+    std::unique_ptr<unsigned char, void(*)(void*)> stat_hold(0, free);
```
JenniferBuehler: Note: It wasn't compiling without `std` for me

include/pistache/date.h
```diff
@@ -52,6 +52,7 @@
 #include <string>
 #include <utility>
 #include <type_traits>
+#include <memory>
```
JenniferBuehler: Note: Required to compile for me

include/pistache/date.h
dennisjenkins75: I have not studied this file much, but no header should ever import a namespace, not even std.  All symbols in headers should be fully qualified. Adding 'std::' here was the correct thing to do.

src/client/client.cc
```diff
@@ -1,6 +1,6 @@
-/* 
+/*
    Mathieu Stefani, 29 janvier 2016
-   
+
```
dennisjenkins75: Please do not make syntax/formatting changes in a PR that is also designed to change functionality.  Please back out all changes that affect whitespace from this PR.

include/pistache/date.h
```diff
@@ -4576,10 +4577,10 @@ to_stream(std::basic_ostream<CharT, Traits>& os, const CharT* fmt,
                         auto d = static_cast<int>(static_cast<unsigned>(fds.ymd.day()));
                         if (d < 10)
                             os << ' ';
-                        os << d << ' ' 
+                        os << d << ' '
```
dennisjenkins75: Please do not change whitespace/formatting in the same PR as a logic change.

include/pistache/http.h
```diff
@@ -98,11 +100,32 @@ namespace Uri {
...
-        
+
         void clear() {
             params.clear();
         }
 
+        // \brief Return iterator to the beginning of the parameters map
+        std::unordered_map<std::string, std::string>::const_iterator
+          parameters_begin() const {
+          return params.begin();
```
dennisjenkins75: (since this is new code), should this line be indented, to follow existing (although undocumented) style?  Same in parameters_end(), below.

JenniferBuehler: Yup, sorry. This happened as I saw last minute that you are using the style with the brackets in the same line as the function declaration. I had it with the bracket in a new line and then the indent was OK. Changing it now.

include/pistache/date.h
JenniferBuehler: Yes that happened unintentionally, my editor removes trailing whitespaces automatically. I thought it would be OK to leave, but will remove it.
Thanks for the review!

JenniferBuehler:
Ok, formatting adjusted.

JenniferBuehler:
Oh no, how silly, this must have snug in with copy and paste :( sorry, I really should have looked more closely.
Since this is already merged, would you like to just update it in your master branch directly, or should I submit a new PR for the fix?

PR #205, opened 2018-03-16
Dynamic buffer size on headers and request body. Ability to use raw headers in return request. Access IP address of request sender

bmr1:
Dynamic buffer size on headers and request body. Starts at 2000 bytes then increments by 40% until the buffer is large enough to acomadate the data.

Ability to use raw headers in return request. Access IP address of request sender

### DIFF
--- a/src/client/client.cc
+++ b/src/client/client.cc
@@ -323,19 +323,21 @@ Transport::handleConnectionQueue() {
 
 void
 Transport::handleIncoming(const std::shared_ptr<Connection>& connection) {
-    char buffer[Const::MaxBuffer];
-    memset(buffer, 0, sizeof buffer);
-
     ssize_t totalBytes = 0;
+    ssize_t currentBytes = 2000;
+
+    std::vector<char> buffer;
+    buffer.resize(currentBytes);
+
     for (;;) {
 
         ssize_t bytes;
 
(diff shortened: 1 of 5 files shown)


include/pistache/stream.h
```diff
@@ -76,53 +76,61 @@ class RawStreamBuf : public StreamBuf<CharT> {
...
+    {
+        buffSize = size;
+        bytes.resize(buffSize);
+        Base::setg(&bytes[0], &bytes[0], &bytes[buffSize]);
     }
 
     bool feed(const char* data, size_t len) {
-        if (size + len >= N) {
-            return false;
+        if (size + len >= buffSize) {
```
xoac: #140

oktal:
Hello, thanks for the Pull Request !

I haven't had time to review it all yet, but I just merged #140 in. Would you like to update your PR accordingly ?

bmr1:
merged

bmr1:
It appears that when a large request comes in with curl the data, in `parseContentLength` http.cc, isnt complete so writes the headers to the body content. It doesn't appear to do it with postman sending the request. the code needs to find a way to skip the header text if the request is incomplete. Ill look into this further tomorrow maybe

bmr1:
@dennisjenkins75  Passes tests and works with unnecessarily long requests. Tested with curl, postman and Firefox

Request: https://pastebin.com/uDSyFHWx

bmr1:
Feel free to close it

PR #209, opened 2018-04-04
Simple update to the README.md to include build info 

jplflyer:
This is a simple update to the README.md to include build info plus notes for building the examples. Most of the notes are from the Getting Started document, but having it in the README.md makes it easier to find.


oktal:
Yep, that's a nice to have in the README. Thanks

PR #216, opened 2018-04-28
Fixed missing virtual destructor of struct Step

dmitry-akimov-dXgjvQjNiMuVYecJ:
Fixed the missing virtual destructor of struct Step. Found by the address sanitizer.


oktal:
Good catch !

Thanks for the contribution

PR #220, opened 2018-05-18
Clear resource handlers at runtime

Fylax:
Added Routes::Clear<Verb> and router.clear_verb to remove all handlers on a given resource for a given verb


include/pistache/router.h
```diff
@@ -154,12 +158,20 @@ class Router {
     void del(std::string resource, Route::Handler handler);
     void options(std::string resource, Route::Handler handler);
 
+    void clear_get(std::string resource);
```
oktal: Pistache uses camel case. Should be `clearGet`

include/pistache/router.h
```diff
@@ -154,12 +158,20 @@ class Router {
...
+    void clear_post(std::string resource);
```
oktal: See above

src/server/router.cc
```diff
@@ -321,6 +351,18 @@ Router::addRoute(Http::Method method, std::string resource, Route::Handler handl
     r.push_back(Route(std::move(resource), method, std::move(handler)));
 }
 
+void Router::clearRoutes(Http::Method method, std::string resource) {
+    auto& r = routes[method];
+    for(int i = 0; i < r.size(); i++) {
```
oktal: `i` sould be `size_t`

Fylax:
I'll fix it as soon as possible

PR #226, opened 2018-05-29
[ENDPOINT] NEW: Now can use SSL encryption on HTTP endpoint

Ne02ptzero:
## Introduction

This patch introduces full SSL support for the pistache endpoint.
The purpose of this patch is to be able to use SSL encryption without
a SSL proxy, which is pretty useful for embedded development.

It has been tested locally with simple SSL support and certificate
connection, my example file is below. I did try to match the
coding style of pistache in order to ease the merge, apologies if
I missed anything. 

## Features

First of all, in order to enable SSL support in the library, one has to
compile with PISTACHE_USE_SSL option ON. One can then configure the
endpoint to accept / use SSL connections only:

    Http::Endpoint server(addr);
    server.useSSL("./server.crt", "./cert/server.key");

With that done, all the connections to the server shall now be in HTTPS.

One can also enable certificate authentication against the server:

    server.useSSLAuth("./rootCA.crt");

The server will now only accept client connection with a certificate
signed with the Certificate Authority passed in the example above.

## Tests
**Full test file**:
```C++
#include <pistache/endpoint.h>

using namespace Pistache;

struct HelloHandler : public Http::Handler {
    HTTP_PROTOTYPE(HelloHandler)

    void onRequest(const Http::Request& request, Http::ResponseWriter writer) {
        writer.send(Http::Code::Ok, "Hello, World!");
    }
};

int main(void) {
    Pistache::Address   addr;
    auto opts = Http::Endpoint::options().threads(1);

    addr = Pistache::Address(Pistache::Ipv4::any(), Pistache::Port(9080));

    Http::Endpoint server(addr);
    server.init(opts);
    server.setHandler(std::make_shared<HelloHandler>());
    server.useSSL("./cert/server.crt", "./cert/server.key");

    server.serve();
    return 0;
}
```

**Compiled with**:
```
g++ main.cpp -L../pistache/build/src/ -lpistache -I../pistache/include/ -o server -lpthread -lssl -lcrypto
```
In order to generate a Certificate Authority and server / client certificate and key, please [refer to this gist](https://gist.github.com/fntlnz/cf14feb5a46b2eda428e000157447309)

### SSL

**Simple CURL**:
```
$> curl https://127.0.0.1:9080/test --cacert CA/rootCA.crt -k                        
Hello, World!% 
```
With `CA/rootCA.crt` as my root Certificate Authority, used to generate the server certificate. The `-k` is to ignore the curl error:
```
curl: (51) SSL: certificate subject name 'Common Name Test Server' does not match target host name '127.0.0.1'
```
Since I did not generate a server certificate for `127.0.0.1`

### SSL authentication
With the following line added in my test file:
```
server.useSSLAuth("./CA/rootCA.crt");
```

**Simple CURL**:
```
$> curl https://127.0.0.1:9080/test --cacert CA/rootCA.crt -k
curl: (35) error:14094410:SSL routines:ssl3_read_bytes:sslv3 alert handshake failure
```

**With proper client cert & key**:
```
$> curl https://127.0.0.1:9080/test --cacert CA/rootCA.crt --cert ./client_cert/client.crt --key client_cert/client.key -k
Hello, World!%
```

Same usage of the `-k` option as above.

## Verbose output
```
$> curl https://127.0.0.1:9080/test --cacert CA/rootCA.crt --cert ./client_cert/client.crt --key client_cert/client.key -k -v
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to 127.0.0.1 (127.0.0.1) port 9080 (#0)
* ALPN, offering h2
* ALPN, offering http/1.1
* successfully set certificate verify locations:
*   CAfile: CA/rootCA.crt
  CApath: none
* TLSv1.2 (OUT), TLS handshake, Client hello (1):
* TLSv1.2 (IN), TLS handshake, Server hello (2):
* TLSv1.2 (IN), TLS handshake, Certificate (11):
* TLSv1.2 (IN), TLS handshake, Server key exchange (12):
* TLSv1.2 (IN), TLS handshake, Request CERT (13):
* TLSv1.2 (IN), TLS handshake, Server finished (14):
* TLSv1.2 (OUT), TLS handshake, Certificate (11):
* TLSv1.2 (OUT), TLS handshake, Client key exchange (16):
* TLSv1.2 (OUT), TLS handshake, CERT verify (15):
* TLSv1.2 (OUT), TLS change cipher, Client hello (1):
* TLSv1.2 (OUT), TLS handshake, Finished (20):
* TLSv1.2 (IN), TLS handshake, Finished (20):
* SSL connection using TLSv1.2 / ECDHE-RSA-AES256-GCM-SHA384
* ALPN, server did not agree to a protocol
* Server certificate:
*  subject: C=FR; ST=Paris; L=Paris; O=XXX; CN=Common Name Test Server
*  start date: May 29 15:59:03 2018 GMT
*  expire date: Oct 11 15:59:03 2019 GMT
*  issuer: C=FR; ST=Paris; L=Paris; O=XXX; CN=Common Name Root Authority; emailAddress=support@xxx.com
*  SSL certificate verify ok.
> GET /test HTTP/1.1
> Host: 127.0.0.1:9080
> User-Agent: curl/7.60.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< Connection: Keep-Alive
< Content-Length: 13
< 
* Connection #0 to host 127.0.0.1 left intact
Hello, World!% 
```


Zenohate:
It would be extremely profitable for the community to have this PR reviewed

Ne02ptzero:
Merge, conflicts and rebasing are now done. I've re-run my originals tests and did not see any regression, should be good to go.

include/pistache/listener.h
```diff
@@ -66,6 +70,11 @@ class Listener {
 
     void pinWorker(size_t worker, const CpuSet& set);
 
+#ifdef PISTACHE_USE_SSL
+    void setupSSL(std::string cert_path, std::string key_path, bool use_compression);
```
snilga: IMO from the performance perspective it's may be better to pass both, cert_path and key_path as 
constant references (const std::string&) rather than copies of the strings (std::string)

include/pistache/endpoint.h
```diff
@@ -49,6 +49,45 @@ class Endpoint {
...
+     *
+     * \note use_compression is false by default to mitigate BREACH[1] and
+     *          CRIME[2] vulnerabilities
+     * \note This function will throw an exception if pistache has not been
+     *          compiled with PISTACHE_USE_SSL
+     *
+     * [1] https://en.wikipedia.org/wiki/BREACH
+     * [2] https://en.wikipedia.org/wiki/CRIME
+     */
+    void useSSL(std::string cert, std::string key, bool use_compression = false);
```
snilga: IMO from the performance perspective it's may be better to pass both, cert and key as 
constant references (const std::string&) rather than copies of the strings (std::string)

snilga:
Ne02ptzero, you've been doing great job! Do you think SSL support can also be added to Http::Client to make it possible to secure communication between small micro-services running in small locally distributed embedded devices?

CMakeLists.txt
```diff
@@ -18,6 +19,11 @@ endif()
 set(CMAKE_CXX_STANDARD_REQUIRED OFF)
 set(CMAKE_CXX_EXTENSIONS OFF)
 
+if (PISTACHE_SSL STREQUAL "ON")
```
mauriciovasquezbernal: It is not needed to test against `"ON"`.  `if (PISTACHE_SSL)` would work fine.

src/server/listener.cc
```diff
@@ -340,5 +389,82 @@ Listener::dispatchPeer(const std::shared_ptr<Peer>& peer) {
...
+    if (!ca_path.empty())
+        __ca_path = ca_path.c_str();
+
+    if (SSL_CTX_load_verify_locations(this->ssl_ctx_, __ca_file, __ca_path) <= 0) {
+        ERR_print_errors_fp(stderr);
+        throw std::runtime_error("Cannot verify SSL locations");
+    }
+
+    SSL_CTX_set_verify(this->ssl_ctx_,
+        SSL_VERIFY_PEER | SSL_VERIFY_FAIL_IF_NO_PEER_CERT | SSL_VERIFY_CLIENT_ONCE, 0);
```
mauriciovasquezbernal: Last parameter is a callback, NULL would be more appropriate than 0.

src/server/listener.cc
```diff
@@ -340,5 +389,82 @@ Listener::dispatchPeer(const std::shared_ptr<Peer>& peer) {
...
+Listener::setupSSLAuth(std::string ca_file, std::string ca_path)
+{
+    const char *__ca_file = NULL;
+    const char *__ca_path = NULL;
+
+    if (!ca_file.empty())
+        __ca_file = ca_file.c_str();
+    if (!ca_path.empty())
+        __ca_path = ca_path.c_str();
+
```
mauriciovasquezbernal: It should check that `Listener::setupSSL` has been already called, if it hasn't the program crashes.

mauriciovasquezbernal:
@Ne02ptzero thanks so much for this PR, it is a big improvement for the library.

I am not an OpenSSL expert, so I cannot give a deep review on that aspect, I just added some comments about the general code structure.

I tested it in our project and I was getting some crashes, after some debugging I found that I have to do `#define PISTACHE_USE_SSL` in our application before including the pistache headers because otherwise  structures in the headers will be different from the ones in the library. 
The example provided above is just working by luck, a more complex application would crash.

I am not sure what is the best way to solve it, leaving it to the user doesn't look like a good solution for me (user will get not compilation warnings, just crashes).

A possibility is to add that definition in https://github.com/oktal/pistache/blob/master/libpistache.pc.in (that by the way is missing -lssl -lcrypto when SSL is enabled).

Another option is to put that define in the headers file automatically.

janhkrueger:
A bit thank you to all who worked toward this!

kiplingw:
> A possibility is to add that definition in https://github.com/oktal/pistache/blob/master/libpistache.pc.in (that by the way is missing -lssl -lcrypto when SSL is enabled).

Yes, pkg-config definitely needs to know about the additional libraries to link against. We want to make it easy for the user. They install a precompiled package for their distro, most likely with SSL encryption enabled, and when it was built CMake automatically substituted those linker flags into the libpistache.pc.in template.

Ne02ptzero:
Many thanks to you all for the kind words and the great reviews.

> Do you think SSL support can also be added to Http::Client to make it possible to secure communication between micro-services running in small locally distributed embedded devices?

Definitely, but that's not in the scope of this patch.

> I tested it in our project and I was getting some crashes, after some debugging I found that I have to do #define PISTACHE_USE_SSL in our application before including the pistache headers because otherwise structures in the headers will be different from the ones in the library.

I've reproduced it, and I think I've found a good solution. The ssl context is now a `void *` pointer, and it is compiled whether or not the library has been compiled with `PISTACHE_USE_SSL`. This way the user do not have to define `PISTACHE_USE_SSL` in his code, since all the variables are public anyway. The comportment is now at runtime; If the user use SSL routes without having compiled the library with SSL, he is now greeted with:
```
./server
  what():  Pistache is not compiled with SSL support.
[3]    12657 abort (core dumped)  ./server
```

> A possibility is to add that definition in https://github.com/oktal/pistache/blob/master/libpistache.pc.in (that by the way is missing -lssl -lcrypto when SSL is enabled).

I've discussed it with KombuchaKip on IRC (He will handle the 'debianization` of pistache), and he will add / handle the OpenSSL dependence.

kiplingw:
> I've discussed it with KombuchaKip on IRC (He will handle the 'debianization` of pistache), and he will add / handle the OpenSSL dependence.

Thanks @Ne02ptzero. I've just submitted a PR. https://github.com/oktal/pistache/pull/351

kiplingw:
@dennisjenkins75, once his unit test is ready I'll also have the Debian maintainer scripts with debhelper's assistance run all the unit tests during package build.

src/server/listener.cc
```diff
@@ -348,5 +392,83 @@ Listener::dispatchPeer(const std::shared_ptr<Peer>& peer) {
...
+    if (SSL_CTX_load_verify_locations((SSL_CTX *)this->ssl_ctx_, __ca_file, __ca_path) <= 0) {
+        ERR_print_errors_fp(stderr);
+        throw std::runtime_error("Cannot verify SSL locations");
+    }
+
+    SSL_CTX_set_verify((SSL_CTX *)this->ssl_ctx_,
+        SSL_VERIFY_PEER | SSL_VERIFY_FAIL_IF_NO_PEER_CERT | SSL_VERIFY_CLIENT_ONCE, NULL);
```
mauriciovasquezbernal: @Ne02ptzero what do you think about exposing this `verify_callback` parameter to the user?
It would be useful for implementing an extended client authentication mechanism, like for instance validating the Common Name on the certificate, checking against a Revocation List, etc.

mauriciovasquezbernal:
@Ne02ptzero do you have any updates on this?

svdHero:
Thanks for all your work guys. I am so looking forward to the SSL feature.

kiplingw:
> > As such (and with all due respect to the contributor) I don't think we should accept this at this time. I I have to point to a red-flag it's that std::string is not an appropriate container for a private key. This is because of the way the memory is released; With std::string it's not possible to reliably overwrite the key data with suitable random data before release of the memory.
> 
> The content of the key is not stored in a `std::string`, the path is.

And even if it was contained, there should be a way to handle its destruction. This could be done either by encapsulating it in a shared_ptr with a lambda destructor or using a custom allocator in the string's constructor. 

I'd also be reluctant to expand the dependencies beyond OpenSSL to other libraries. OpenSSL is a pretty common optional configure / build time dependency for many projects and is packaged on every major distro. It's therefore quite reasonable.

svdHero:
> As it stands it is still more secure to use a reverse proxy (such as nginx) to offload the TLS safely in a dedicated process.

But then again this might not be possible for embedded projects. I am currently working on some IoT stuff where the RAM limitations are just crazy. Adding nginx to the tech stack would probably break the camel's back. :worried: 

I think embedded projects are one of the main consumers of Pistache. All the cool web kids out there probably prefer node.js. It's embedded bit-jugglers like myself who benefit the most from a C++ library like Pistache. Hence, build-in TLS support with OpenSSL as a sole dependency would be awesome.

hydratim:
> The content of the key is not stored in a `std::string`, the path is.

I missed that, the name on line 73 of endpoint.h reads as though it's the key data, I see now that in the implementation of this the name is indeed key_path.


> there should be a way to handle its destruction

There is, but the risk of an overflow makes this potentially vulnerable (we still have known memory leaks are we really prepared to say we don't have other memory issues at this time).


> OpenSSL is a pretty common optional configure / build time dependency

That's a good argument but I'd prefer if we made a more detailed evaluation that allowed the user to select the backing library.


> But then again this might not be possible for embedded projects.

I see your concerns, my point isn't that we shouldn't include it, but that by including it means people will assume it's safe, which means we need to be very confident that we aren't going to have a security hole that exposes their private keys.


At the end of the day it's not my decision, I'm just raising my concerns about including it as things stand.

svdHero:
> At this moment, pistache leaks memory on every handled request. imho, that is a huge blocker to adopting pistache in small devices.

Oh wow. Thanks for mentioning this here. I didn't know that and have only now seen #399.

Ne02ptzero:
Alright @dennisjenkins75, conflicts are resolved, tests are good, commit history looks good. Ready to merge!

PR #231, opened 2018-05-31
Close 2 file descriptors opened from serveFile - avoid crash

### LINKED ISSUE #229: With serveFile(), 2 file descriptors are opened, but not closed, for every file request, until crash with "Too many open files"
Reproduces on Linux with the http_server example on the "/static" url (with README.md in the working directory); crash happens after 1024 file descriptors are open.

One can see all the open file descriptors for a process by name with ``$ ls /proc/`pgrep run_http_server`/fd``.

This issue would lead to an inevitable crash after a certain time for any project that uses Pistache to serve files, so I'm wondering if there's any way this can be fixed. I'm currently playing around with the end of the `serveFile()` function - it looks like `fd` and `sockFd` get created and used there, but I'm having a rough time getting the promise `.then(...)` signature to compile with my tweaks.

jeffvandyke:
Fixes #229, previously caused crash after ~500 files served with `serveFile` because of 2 open file descriptors.

`FileBuffer` was missing a destructor to RAII-style close the `fd_` that the constructor opened. For consistency to make sure FileBuffer always manages a `Fd`, I deleted the default constructor as well. `_fd` is initialized to `-1` to make sure that only valid file descriptors are closed in the destructor.

The other one was a `fd` opened for a call to `::fstat(fd, &sb);` inside `serveFile`, so I closed it as soon as it's finished being used.

The before and after behavior can be examined from the `examples/http_server.cc` program, with the open file descriptors viewable in Linux from ``$ ls /proc/`pgrep run_http_server`/fd`` after using the `/static` url.

### DIFF
--- a/include/pistache/stream.h
+++ b/include/pistache/stream.h
@@ -156,10 +156,11 @@ struct Buffer {
 };
 
 struct FileBuffer {
-    FileBuffer() { }
+    FileBuffer() = delete;
 
     FileBuffer(const char* fileName);
     FileBuffer(const std::string& fileName);
+    ~FileBuffer();
 
     std::string fileName() const { return fileName_; }
     Fd fd() const { return fd_; }
--- a/include/pistache/stream.h
+++ b/include/pistache/stream.h
@@ -156,11 +156,10 @@ struct Buffer {
 };
 
 struct FileBuffer {
-    FileBuffer() = delete;
+    FileBuffer() { }
 
     FileBuffer(const char* fileName);
     FileBuffer(const std::string& fileName);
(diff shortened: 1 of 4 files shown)


jeffvandyke:
Hold on this - apparently this has a rough time with serving large files. I'm still working on it though.

jeffvandyke:
A different pull request #39 is relevant, and appears to fix the same issue - I haven't test it, but I'm concerned over if it would work for large files (> 10k or so)

jeffvandyke:
Thanks for merging, and I'm glad it could help!

PR #233, opened 2018-06-01
Exit after closeListener() when handling SIGINT

jeffvandyke:
Allows the http_server example to exit on <kbd>Ctrl-c</kbd> as well as clean up properly.

It doesn't seem to make sense to close the `fd` created in `Listener::bind(const Address& address)` and then sit there, unbound, but not exiting.

Thoughts or concerns?

### DIFF
--- a/src/server/listener.cc
+++ b/src/server/listener.cc
@@ -40,8 +40,9 @@ namespace {
         }
     }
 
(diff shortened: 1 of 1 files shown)


xoac:
@jeffvandyke I use pistache for embedded (server) so feel free to pm me.

PR #236, opened 2018-06-06
Tree Structure and Removal

Fylax:
* Added C++17 compliant string_view subset
* Moved Routes on a Fragments Tree Structure
* Added ability to remove routes


Fylax:
As you requested, I have added unit tests for string view and a disclaimer on the hashing algorithm.

Fylax:
Hi dennisjenkins75,
despite the documentation stating that both `pos` and `count` have default values, seems like that GCC (even version 8) threats the "no arguments" version as a different and not compiled one.

I have changed the tests to remove that specific one. Now it should work on your machine too.

PR #246, opened 2018-07-04
Allow access to unknown and arbitrary raw headers

milipili:
It may happen that the headers are not know in advance. One possible
scenario is that the caller provide a bunch of arbitrary headers with a
predefined prefix, acting as customer metadata, that the applications
needs to store somewhere for later reuse, in responses for example.

Those arbitrary might be used by the customer for many purposes, like
custom load-balancing or particular routing.


kiplingw:
Hey @milipili. You may also find #551 helpful. This was just pushed to the PPAs last night.

PR #257, opened 2018-07-26
fix memory leak

xqing2003:
fix memory leak, storage's type is the template type T, so the type T's destroy function will not be called.

### DIFF
--- a/include/pistache/async.h
+++ b/include/pistache/async.h
@@ -169,6 +169,7 @@ namespace Async {
             Core(State state, TypeId id)
                 : state(state)
                 , id(id)
+                , is_construct(false)
             { }
 
             State state;
@@ -188,6 +189,7 @@ namespace Async {
             std::mutex mtx;
             std::vector<std::shared_ptr<Request>> requests;
             TypeId id;
+            bool is_construct;
 
             virtual void* memory() = 0;
 
@@ -203,8 +205,12 @@ namespace Async {
                 }
 
                 void *mem = memory();
+                if (is_construct) {
+                    reinterpret_cast<T *>(mem)->~T();
+                };
                 new (mem) T(std::forward<Args>(args)...);
                 state = State::Fulfilled;
+                is_construct = true;
             }
 
         };
@@ -214,7 +220,11 @@ namespace Async {
             CoreT()
                 : Core(State::Pending, TypeId::of<T>())
             { }
-
+            ~CoreT() {
+               if(is_construct) {
+                  reinterpret_cast<T *>(&storage)->~T();
+               }
+            }
             template<class Other>
             struct Rebind {
                 typedef CoreT<Other> Type;


xqing2003:
I'm sorry for not having discribed it clearly. I use async code  in my project. I found if the Promise<T>'s template type T is a struct or class, it use std::aligned_storage to alloc memory,  and use placement new to construct T, but T's destructor will never be called.
As you say, examples/run_http_server has no such problem, because it only use type ssize_t as Promise's template type. I found examples/run_http_client use Promise<Http::Response>, and valgrind reports leak messages :
`=30341== 1,671 (552 direct, 1,119 indirect) bytes in 1 blocks are definitely lost in loss record 35 of 40
==30341==    at 0x4C285FC: operator new(unsigned long) (vg_replace_malloc.c:298)
==30341==    by 0x414456: __gnu_cxx::new_allocator<Pistache::Http::Response>::allocate(unsigned long, void const*) (new_allocator.h:104)
==30341==    by 0x4142BF: std::allocator_traits<std::allocator<Pistache::Http::Response> >::allocate(std::allocator<Pistache::Http::Response>&, unsigned long) (alloc_traits.h:488)
==30341==    by 0x4140A5: std::_Vector_base<Pistache::Http::Response, std::allocator<Pistache::Http::Response> >::_M_allocate(unsigned long) (stl_vector.h:170)
==30341==    by 0x418258: std::_Vector_base<Pistache::Http::Response, std::allocator<Pistache::Http::Response> >::_M_create_storage(unsigned long) (stl_vector.h:185)
==30341==    by 0x4180E8: std::_Vector_base<Pistache::Http::Response, std::allocator<Pistache::Http::Response> >::_Vector_base(unsigned long, std::allocator<Pistache::Http::Response> const&) (stl_vector.h:136)
==30341==    by 0x417FAC: std::vector<Pistache::Http::Response, std::allocator<Pistache::Http::Response> >::vector(std::vector<Pistache::Http::Response, std::allocator<Pistache::Http::Response> > const&) (stl_vector.h:320)
==30341==    by 0x417E65: void Pistache::Async::Private::Core::construct<std::vector<Pistache::Http::Response, std::allocator<Pistache::Http::Response> >, std::vector<Pistache::Http::Response, std::allocator<Pistache::Http::Response> >&>(std::vector<Pistache::Http::Response, std::allocator<Pistache::Http::Response> >&) (async.h:206)
==30341==    by 0x416AFD: bool Pistache::Async::Resolver::operator()<std::vector<Pistache::Http::Response, std::allocator<Pistache::Http::Response> >&>(std::vector<Pistache::Http::Response, std::allocator<Pistache::Http::Response> >&) const (async.h:745)
==30341==    by 0x415AD1: Pistache::Async::Impl::WhenAllRange<Pistache::Http::Response, std::vector<Pistache::Http::Response, std::allocator<Pistache::Http::Response> > >::WhenContinuation<Pistache::Http::Response, void>::operator()(Pistache::Http::Response const&) const (async.h:1434)`

xqing2003:
I have remove the conflict. Thank you.

include/pistache/async.h
```diff
@@ -203,8 +205,12 @@ namespace Async {
                 }
 
                 void *mem = memory();
+                if (is_construct) {
+                    reinterpret_cast<T *>(mem)->~T();
+                };
                 new (mem) T(std::forward<Args>(args)...);
```
milipili: Actually, this is not exception safe. It would be probably better like that:
```cpp
auto new_mem = std::make_unique<T>(std::forward<Args>(args)...); // new may throw here
void* mem = memory();
if (is_construct && mem != nullptr)
    reinterpret_cast<T*>(mem)->~T();
mem = new_mem.release();
state = State::Fulfilled;
is_construct = true;
```

include/pistache/async.h
```diff
@@ -188,6 +189,7 @@ namespace Async {
             std::mutex mtx;
             std::vector<std::shared_ptr<Request>> requests;
             TypeId id;
+            bool is_construct;
```
milipili: Since we have `std::forward` and unless I didn't see something, we can certainly simply write

```cpp
bool is_construct = false;
```

dennisjenkins75: xqing2003: Do you have any thoughts on the feedback from milipili?

Can you author a unit-test the triggers the memory leak, so that I can do a before/after test nf the proposed fix?

I still need to fix the SIGINT handler so that pistache programs that call Pistache::Http::Endpoint::serve() will exit cleanly.  Without that, valgrind will report leaks.

PR #263, opened 2018-08-13
Dynamic arraybuf

iroddis:
- Changed ingestion buffer to std::vector instead of a stack-based array. There will be a performance penalty, but it allows for file uploads
- Added Const::MaxPayload to avoid killing hosts due to massive payloads.


iroddis:
```
$ perl -e 'print "A" x 4000' | curl -s -X POST --data-binary @- http://localhost:9080/echo | wc -c               
4000

$ perl -e 'print "Z" x 1024' | curl -s -X POST --data-binary @- http://localhost:9080/echo | sed -e 's/[^Z]/_/g' | awk -FZ '{ print NF-1}'
1024

$ perl -e 'print "Z" x 1025' | curl -s -X POST --data-binary @- http://localhost:9080/echo | sed -e 's/[^Z]/_/g' | awk -FZ '{ print NF-1}'
1025
```

These are the tests that I saw, and they seem to work correctly. Here's another random test:

```
$ ls -l bigdata.csv
-rw-r--r--. 1 roddis roddis 888888898 Aug 11 19:20 bigdata.csv
$ ls -lh ~/bigdata.csv
-rw-r--r--. 1 roddis roddis 848M Aug 11 19:20 bigdata.csv
$ curl -X POST --data-binary @bigdata.csv http://localhost:9080/echo | wc -c
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 1695M  100  847M  100  847M  28.9M  28.9M  0:00:29  0:00:29 --:--:--  179M
888888898
```
I haven't done any performance regression tests; the use of heap allocated buffer vs stack is sure to have an impact.

iroddis:
Good news so far! I have the fix for the test case (wasn't maintaining the get cursor correctly for the buffer). Making the max payload parameter a runtime config shouldn't be a problem at all, I'll get on that tonight / soon.

Any objection to some name changes? Right now, there's Const::MaxBuffer that's being used to set the max amount of data recv()'d at a time, Const::ChunkSize which isn't used for anything that I can see, and Const::MaxPayload.

ChunkSize seems to more accurately describe what MaxBuffer is being used for, and after making MaxPayload a runtime parameter, I can just remove MaxPayload and MaxBuffer.

iroddis:
Nope, I'm happy to write up the tests. Here's the list of what's outstanding on this PR, then, which I'll work on:

- Make MaxPayload a runtime, not compile-time parameter
- Set MaxPayload to 4096 by default to keep prior behaviour
- Add test to ensure that 413 (Payload Too Large) is properly returned when that constraint is violated
- Add test to ensure that large payloads <= MaxPayload (counting request headers) work as intended.

iroddis:
I'm working on making MaxPayload a runtime parameter, but it's turning out to be less trivial than I thought it'd be. The natural top level interface to set the parameter is in the Handler, but when the reactor calls `clone()` the clones don't have the runtime parameter copied.

I'm playing around with different solutions, but so far, it's not straightforward. The clone() method is being called on a `shared_ptr<Aio::Handler>`, but the object that it's pointing to is a `Tcp::Handler`, meaning that if I override the clone() method on Http::Handler, it doesn't get called properly, and the maxPayloadSize isn't propagated.

Any suggestions on how to implement this?

iroddis:
I'm more than happy to stick to the 4096 limit that's been the norm; I can easily apply a patch during build time with cmake to adjust the constant, or make it a compile-time definition (e.g. -DPAYLOAD_MAX=8192). I do have a use case for multi-GB uploads, but I don't think I'm the norm in this.

I can add the handler tests this weekend (sorry for taking a while, I know you want this merged).

iroddis:
@dennisjenkins75 Ok, it's all done. I've updated the README.md in the root of the project on how to define the max payload size. By default it's 4096 bytes.

One caveat is, since the payload size is a compile-time definition, it's not really possible to write a test case for it, since it's not possible to vary the buffer size dynamically. I tested it manually using a payload size of 8192, and verified that the correct return codes happened (200 and 413).

This last commit also includes the changes required to rebase to the current master.

include/pistache/common.h
```diff
@@ -52,12 +52,20 @@
...
-    static constexpr int MaxBacklog = 128;
-    static constexpr int MaxEvents = 1024;
-    static constexpr int MaxBuffer = 4096;
-    static constexpr int ChunkSize = 1024;
+    static constexpr size_t MaxBacklog = 128;
+    static constexpr size_t MaxEvents  = 1024;
+    static constexpr size_t MaxBuffer  = 4096;
+
+    // Defined from CMakeLists.txt in project root
+    static constexpr size_t MaxPayload = MAX_PAYLOAD;
```
dennisjenkins75: I was super-excited to test this PR.  When Pistache compiles, "Pistache::Const::MaxPayload" is properly set via the -DPISTACHE_MaX_PAYLOAD.  So the installed .so has this feature.

Unfortunately, the installed header files still have the MaxPayload at 4096 bytes.  When I compile my own project that "#include <pistache.h>" (or equiv), MaxPayload is incorrect and my client-side code makes poor decisions.  We need a way to preserve the compile-time size of MaxPayload so that library users can obtain it, or better yet, an API that can be called that returns whatever value was compiled in (instead of just reading a constant from a header file).

Thoughts?

iroddis: Yup, that seems to be the way to go. I'll get it working.

iroddis:
I think this has done it... I had to change the clone() macro in prototype.h to use a copy constructor  of whatever it is prototyping. For the most part, this shouldn't cause any problems, and it allows the handler to be cloned without losing the payload parameter. Fingers crossed on this one :)

Also, it turns out the macro isn't a hack (every day I'm learning something new). There isn't a way to make a nice clone method using normal methods. I tried playing around with decltype, etc, but no luck. The macro is the only way to do it.

iroddis:
Aha, yes, I can definitely do that. I wasn't testing with Swagger. It should be trivial to add. I'll see about adding a test case as well.

I'd also like to update the documentation somewhere, probably as a separate PR. There just seems to be the base README and and the examples, but there is a website that has some examples. Do you have any suggestions on where to start for making the documentation a part of the codebase?

iroddis:
Sorry, that last commit was a checkpoint for my work.

Using your code, I've replicated the issue as a unit test. I'm working on resolving the issue now.

iroddis:
Thanks :) This is a bit painful, the Tcp::Handler and Http::Handlers are getting created all over the place, so it's not obvious where to set the maxpayload. I'm still working on it, though.

iroddis:
Finally have a working test case and reduced a ton of clutter by using a static member variable for ArrayStreamBuf, which is set by the endpoint Options interface. Example is in the test case `tests/payload_test.cc`

iroddis:
I'm working on it. It was working for me on Fedora 28, but after I switched the inner and outer loops of the testPayloads function, I'm getting the same error.

iroddis:
I think this has fixed the use case. Reusing an endpoint seems to cause some issues, so I wrapped the endpoints in forks() to avoid the issue. I'm using gcc 8.1.1 on Fedora 28, and it's working pretty reliably now.

Issue #265, opened 2018-08-15
Server Unable to Parse Multiple Cookies in Request


yisaj:
When parsing the headers of a request, Cookie::fromRaw is used to parse the "Cookie" header. fromRaw, however, appears to be written to parse the "Set-Cookie" header from a server response. As a result, the server-side Request object can never have more than one Cookie in its CookieJar object. Any additional cookies are thrown into Cookie.ext.

yisaj:
Yes, you can see the bug if you run "./build/examples/run_rest_server" and "curl localhost:9080/auth -b "key1=value1; key2=value2; key3=value3"". The server only prints a single cookie instead of the expected three. 

I've got a patch that should fix the problem, so I'll see if I can put together a unit test and submit a PR. It's alright if the test just goes in "tests/cookie_test.cc", correct?

Issue #143, opened 2017-09-15
Is there any way to add HTTPS?


oxoocoffee:
After reading few Issues I see some interests in HTTPS. Sounds like this is considered but it migth take some time. Di any one figured out with current release how to add HTTPS support (opening secure socket externally?) If yes where could I find working example?

oktal:
Hello,

Pistache does not currently support https. For now, as a workaround, I would suggest using a reverse proxy such as `nginx` and then use Pistache downstream.

https is definitely something a lot of users have been requesting so it should be supported but I want to work on stabilizing the framework first.

Keeping your issue around when I start prioritizing https.

dennisjenkins75:
If it is added, it will likely be a compile-time option.  Some users want HTTPS and do not want the complexity of running a separate process as a proxy (hint: small, embedded systems).

annnhj:
Hi, was it added to pistache? I mean the https. Tks

dennisjenkins75:
Yes, PR #226, on 2019-01-28.

PR #266, opened 2018-08-15
Fix Request Cookies Parsing

yisaj:
Fixes the parsing of the "Cookie" header of a request by implementing an "addFromRaw" method that adds cookies to a CookieJar directly from the header string. (Issue #265)

### DIFF
--- a/tests/cookie_test.cc
+++ b/tests/cookie_test.cc
@@ -137,14 +137,14 @@ void addCookies(const char* str, std::function<void (const CookieJar&)> testFunc
 
 TEST(cookie_test, cookiejar_test) {
     addCookies("key1=value1", [](const CookieJar& jar) {
-	ASSERT_EQ(jar.get("key1").value, "value1");
+        ASSERT_EQ(jar.get("key1").value, "value1");
     });
-    
+
     addCookies("key2=value2; key3=value3", [](const CookieJar& jar) {
-	ASSERT_EQ(jar.get("key2").value, "value2");
-	ASSERT_EQ(jar.get("key3").value, "value3");
+        ASSERT_EQ(jar.get("key2").value, "value2");
+        ASSERT_EQ(jar.get("key3").value, "value3");
     });
-    
+
     CookieJar jar;
     ASSERT_THROW(jar.addFromRaw("key4", strlen("key4")), std::runtime_error);
 }
--- a/tests/cookie_test.cc
+++ b/tests/cookie_test.cc
@@ -129,3 +129,22 @@ TEST(cookie_test, invalid_test) {
     ASSERT_THROW(Cookie::fromString("lang=en-US; Max-Age=12ab"), std::invalid_argument);
 }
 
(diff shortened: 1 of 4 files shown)


tests/cookie_test.cc
```diff
@@ -129,3 +129,25 @@ TEST(cookie_test, invalid_test) {
...
+TEST(cookie_test, cookiejar_test) {
+    addCookies("key1=value1", [](const CookieJar& jar) {
+        ASSERT_EQ(jar.get("key1").value, "value1");
+    });
+
+    addCookies("key2=value2; key3=value3; key4=; key5=foo=bar", [](const CookieJar& jar) {
+        ASSERT_EQ(jar.get("key2").value, "value2");
+        ASSERT_EQ(jar.get("key3").value, "value3");
+        ASSERT_EQ(jar.get("key4").value, "");
+        ASSERT_EQ(jar.get("key5").value, "foo=bar");
```
dennisjenkins75: This is valid and what I expected, per https://tools.ietf.org/html/rfc6265#section-4.1

tests/cookie_test.cc
```diff
@@ -129,3 +129,25 @@ TEST(cookie_test, invalid_test) {
...
+    });
+    
+    CookieJar jar;
+    ASSERT_THROW(jar.addFromRaw("key4", strlen("key4")), std::runtime_error);
+    ASSERT_THROW(jar.addFromRaw("key4=value4=value4", strlen("key4")), std::runtime_error);
```
dennisjenkins75: Why is this expected to throw an exception?  "keys4=value4=value4" should be functionally equivalent to the test that you perform on line #143 with "key5=foo=bar".

I added an extra test to see what would happen.  The test passed, which means that it threw an exception.  I'm not sure why, as the syntax looks correct?

`+    ASSERT_THROW(jar.addFromRaw("key4=value4", strlen("key4")), std::runtime_error);
`

per https://tools.ietf.org/html/rfc6265#section-4.1, "<cookie-value>" can be quoted.
`curl localhost:9080/auth -b 'key1=value1; key2=value2; key3="value3=value4"'
`
That works.  Yeah.

However, I'm undecided on if the Cookie parse should be strict like the RFC states, or follows Postel's law and accepts things that violate the spec, such as:

```
curl localhost:9080/auth -b 'key1="foo,bar baz\p";'
Cookies: [
    key1 = "foo,bar baz\p"
]

Thoughts?
```

yisaj: Ah, sorry, that's totally my mistake. That assert must be an artifact left over when I was messing with test statements. Since the length passed is less than the length of the string passed, it's essentially a duplicate of the previous "ASSERT_THROW".

As for the question of what to accept as a parseable cookie, I'm not sure. If the reversion at https://www.fxsitecompat.com/en-CA/docs/2016/whitespaces-are-no-longer-allowed-in-cookie-names/ is any indication, at least some user application seem to want to use whitespace in their cookies. For what it's worth, the current implementation of "Cookie::fromRaw" appears to take any string as a key or value.

PR #270, opened 2018-08-20
Fix perfomance issue: redundant strings copying

knowledge4igor:
In this PR I'm fixing redundant strings copying in `Client` class. It would be better to pass resource name by `const std::string&` instread of creating a lot of template object during chain of calls. You can do perfomance tests with passing plenty of strings to `get` method and see the result.

### DIFF
--- a/include/pistache/client.h
+++ b/include/pistache/client.h
@@ -257,7 +257,7 @@ public:
     friend class Client;
 
     RequestBuilder& method(Method method);
-    RequestBuilder& resource(std::string val);
+    RequestBuilder& resource(const std::string& val);
     RequestBuilder& params(const Uri::Query& query);
     RequestBuilder& header(const std::shared_ptr<Header::Header>& header);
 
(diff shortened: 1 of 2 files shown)


knowledge4igor:
@dennisjenkins75 
First of all, using `const string&` as a parameter of function or method you can pass a string literal without any problems, also you can pass a temporary object as an argument.
Now, when you call a get method you **have to** create a temporary object and copy of this object will take place.
You can notice that my pull request proposes to pass a `const string&` until we really need to copy of string in `RequestBuilder::resource` method. It makes core clear. Do you agree with me?
You can profile the code from this project in this way:
```
    ...
    // TODO: create of vector of string with 7000000 strings
    Http::Client client;

    auto opts = Http::Client::options()
        .threads(1)
        .maxConnectionsPerHost(1);
    client.init(opts);

    auto start = std::chrono::system_clock::now();

    for (size_t i = 0; i < pages.size(); ++i)
    {
        auto resp = client.get(pages[i]);
    }

    auto end = std::chrono::system_clock::now();
    std::cout << "Total time of execution           : "
              << std::chrono::duration_cast<std::chrono::milliseconds>(end - start).count() << "ms" << std::endl;

    client.shutdown();
```

Just run and  compare the time of execution. I have seen almost 2x boost in execution after my changes. Just try and you will see the result. I have prepared a tiny and aritificial [example](https://ideone.com/EuDm4c) for you that simulates two different cases. Yes, I have side effects (`std::cout`) in contructor You can see the difference in number of contructors calls. I have tried it on both gcc and clang compilers with -O3 optimization and get the same result.

PR #272, opened 2018-08-22
Improve string usage in code

knowledge4igor:
In this pull request I have improved `std::string` usage in code

### DIFF
--- a/include/pistache/client.h
+++ b/include/pistache/client.h
@@ -272,6 +272,7 @@ public:
 
     RequestBuilder& cookie(const Cookie& cookie);
     RequestBuilder& body(const std::string& val);
+    RequestBuilder& body(std::string&& val);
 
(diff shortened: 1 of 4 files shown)


knowledge4igor:
@dennisjenkins75 Thank you for the comment. I have added the second `body` method for moving value. For user's point of view having only `RequestBuilder& body(std::string &&val)` is not a good idea. In this case each time user have to move value and cannot use string again later in the code. In original version `RequestBuilder& body(std::string val);` there is a creation of temp object takes place on each call and only futher you move it, it cannot be move in this case because user pass `val` by value and expects that this value can be used later.
I propose to have two methods:
```
RequestBuilder& body(const std::string& val);
RequestBuilder& body(std::string&& val);
```
For passing value that will be copied once and for moving value and user have to explicitly call `std::move` and understand that value will be empty after: 
.body(std::move(value)).

PR #274, opened 2018-08-25
Fix problem with calling send method of RequestBuilder 

knowledge4igor:
In this PR I'm fixing the problem with calling `send` method second time and futher on same `RequestBuilder` object. For example, when we execute the following code:
```
...
auto rb = client.get("localhost:9080");
auto resp1 = rb.send();
resp1.then([&completedRequests](Http::Response response) {
}, Async::IgnoreException);
responses.push_back(std::move(resp1));

auto resp2 = rb.send();
resp2.then([&completedRequests](Http::Response response) {
}, Async::IgnoreException);
responses.push_back(std::move(resp2));

auto resp3 = rb.send();
resp3.then([&completedRequests](Http::Response response) {
}, Async::IgnoreException);
responses.push_back(std::move(resp3));

auto sync = Async::whenAll(responses.begin(), responses.end());
Async::Barrier<std::vector<Http::Response>> barrier(sync);
...
```

We will get runtime exception:
```
  what():  ::getaddrinfo(host.c_str(), port, &hints, &addrs): No such file or directory
Aborted (core dumped)
```

It's caused by the fact that `host` value is empty after moving.
```
  what():  ::getaddrinfo(host.c_str(), port, &hints, &addrs): No such file or directory

Thread 1 "http_client_exa" received signal SIGABRT, Aborted.
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
51	../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
#1  0x00007ffff725f801 in __GI_abort () at abort.c:79
#2  0x00007ffff78b48fb in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#3  0x00007ffff78bad3a in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#4  0x00007ffff78bad95 in std::terminate() () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#5  0x00007ffff78bafe8 in __cxa_throw () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#6  0x00005555555cacd3 in Pistache::Http::Connection::connect (this=0x5555558637b0, addr=...) at /home/pistache/src/client/client.cc:396
#7  0x00005555555cd951 in Pistache::Http::Client::doRequest (this=0x7fffffffdb90, request=..., timeout=...)
    at /home/pistache/src/client/client.cc:873
#8  0x00005555555ccde0 in Pistache::Http::RequestBuilder::send (this=0x7fffffffdc90) at /home/pistache/src/client/client.cc:750
#9  0x0000555555561f04 in main ()
(gdb) f 6
#6  0x00005555555cacd3 in Pistache::Http::Connection::connect (this=0x5555558637b0, addr=...) at /home/pistache/src/client/client.cc:396
396	    TRY(::getaddrinfo(host.c_str(), port, &hints, &addrs));
(gdb) p host
$2 = ""
```

From my point of view, it would be better for plain user to have an opportunity to call `send` method again (maybe after several seconds, it depends) and it's not a good practice to force user to keep in mind details of implemetation (that `std::move` is used). Also, in my opinion, it's ok for each `send` method call to pass copy of `Request` object to `doRequest` method .


knowledge4igor:
@dennisjenkins75 I have tried to write a test and faced with a problem. During test execution I run server in the separate thread:
```
  auto options = Http::Endpoint::options().threads(1);
  _endpoint.init(options);
  _endpoint.setHandler(Http::make_handler<HelloHandler>());
  _endpoint.serve();
```
There is a problem with _endpoint.shutdown() call (I do in desctructor). I have decided to fix this problem and only after that I come back to this PR.

knowledge4igor:
@dennisjenkins75 Dennis, I have added a test for verifying that "send()" method can be called multiple times. Is this test follow test style of this project? I'm ready to change something if it's necessary.

knowledge4igor:
@dennisjenkins75 Ok, I will see it now. What about adding running all tests on CI ?

knowledge4igor:
@dennisjenkins75 I have the following error on this test:
```
10: [ RUN      ] request_builder.multiple_send_test
10: terminate called after throwing an instance of 'Pistache::Http::HttpError'
10:   what():  Encountered invalid HTTP version
1/1 Test #10: http_client_test .................***Exception: Child aborted  0.26 sec
```

I will try to fix it asap.

PR #275, opened 2018-08-25
Stream write

iroddis:
Adds a `ResponseStream::write(char *, std::streamsize)` method, which is helpful for sending chunked binary data (e.g. large gzip'd responses).

Also adds a `/stream_binary` example path to the run_http_server to show the write() call in action.

### DIFF
--- a/include/pistache/http.h
+++ b/include/pistache/http.h
@@ -316,6 +316,14 @@ public:
     friend
     ResponseStream& operator<<(ResponseStream& stream, const T& val);
 
+    std::streamsize write(char * data, std::streamsize sz) {
+        std::ostream os(&buf_);
+        os << std::hex << sz << crlf;
+        os.write(data, sz);
+        os << crlf;
+        return sz;
+    }
+
     const Header::Collection& headers() const {
         return headers_;
     }
--- a/include/pistache/http.h
+++ b/include/pistache/http.h
@@ -100,7 +100,7 @@ namespace Uri {
         bool has(const std::string& name) const;
         // Return empty string or "?key1=value1&key2=value2" if query exist
         std::string as_str() const;
-        
+
         void clear() {
             params.clear();
         }
@@ -321,6 +321,7 @@ public:
         os << std::hex << sz << crlf;
         os.write(data, sz);
         os << crlf;
+        flush();
         return sz;
     }
 
@@ -377,6 +378,7 @@ ResponseStream& operator<<(ResponseStream& stream, const T& val) {
     std::ostream os(&stream.buf_);
     os << std::hex << size(val) << crlf;
     os << val << crlf;
+    stream.flush();
 
     return stream;
 }
--- a/include/pistache/http.h
+++ b/include/pistache/http.h
@@ -321,7 +321,6 @@ public:
         os << std::hex << sz << crlf;
         os.write(data, sz);
         os << crlf;
-        flush();
         return sz;
     }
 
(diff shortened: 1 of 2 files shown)


include/pistache/http.h
```diff
@@ -316,6 +316,14 @@ class ResponseStream : public Message {
     friend
     ResponseStream& operator<<(ResponseStream& stream, const T& val);
 
+    std::streamsize write(const char * data, std::streamsize sz) {
+        std::ostream os(&buf_);
+        os << std::hex << sz << crlf;
```
dennisjenkins75: How would one send raw binary data without the hex size prefix and crlf tail?  Maybe you should rename this method to "writeChunkedBinaryFormat()"?  To me "write()" implies that its not going to modify what is being written.

iroddis: If the encoding weren't chunked, the server could just send the binary data as the body, and it would be up to the client to give it meaning. I'm using this approach now, when gzip'ing a serialized object. Just writing the data into a std::string and sending it as the body of the response works just fine.

I don't object to changing the name, but the ResponseStream is explicitly returning encoded chunks. I didn't think it needed to be made explicit in the write method. If that's your only objection, though, I'm happy to do it.

Since the data passed into the function needn't be binary, it could be any arbitrary data, would `writeChunk()` ok?

iroddis: Sorry, I forgot to mention sending the raw binary data works because the content-length is specified, so the client knows how many bytes to expect, so it replaces the hex size and trailing crlf.

iroddis: I'm changing my mind. The hex size and the trailing crlf are (and should be) hidden implementation details of the stream. Just as `operator<<` doesn't advertise that it's mangling the output, the `write()` call is just responsible for getting the data to the client, which it does. I think that `write()` is the appropriate name.

There's also nice symmetry around the operator<< using the << operator of the buf, whereas the write() call uses the write() of the buf.

iroddis:
Last thing for the night (sorry for the many posts). Would there be value in having input operators to the ResponseStream object flush immediately? What's the advantage of waiting? When I was testing my commit, I found that multiple writes would get accumulated in the buffer until either `flush` or `ends` were sent, then all the writes would be sent as a single chunk.

This doesn't seem to make much sense, as now I'm sending the data in one huge response, with extra size and delimiting characters.

I'd imagine that streaming is desirable in a few cases:

1. It's not known ahead of time the size of the response
2. There's a desire to keep buffers small (as in embedded situations), so sending many small responses instead of staging one massive response is the way to go
3. The time to generate the entire payload takes a long time, but async updates will send chunks as further chunks are generated, cutting down on the entire request -> response complete latency.

(2) and (3) make flushing after every write request make sense, and (1) doesn't really have an issue if that's what happens.

My use case for this is boost::iostream gzip'd data. I've made a sink device for ResponseStreams, and the typical write size is 4096 bytes, which is more than an ethernet packet can handle. In this case, there's no advantage to buffering the stream, and great advantage to flushing immediately.

include/pistache/http.h
dennisjenkins75: I don't like that a "write" method modifies the data that's being written.  I think that it might lead to user confusion in the future.  However, I have no strong argument against your change either.  I will merge your PR after your next response.

iroddis:
@dennisjenkins75 your feedback makes sense, thanks.

The ResponseStream class does have a flush() method that can be used to manually transmit the buffered data.

Right now, all data-writing methods (<< and write()) just append the data (with hex size and crlf tail) to the response buffer. No data is sent to the client until either flush() or ends() is called. In the case where a user just writes to the ResponseStream, then calls ends(), this is equivalent to writing the data into a buffer and sending it as a single response, except now it's got extra bytes (hex sizes, etc, for each chunk written).

At the end of the day, though, I can easily write the flush() calls myself, as I know how much data I'm sending. Adding the write() method as is gets me the behaviour I need, and I don't need to push my flush-happy agenda on anyone else :)

PR #276, opened 2018-08-26
Implement set timeout method in RequestBuilder

knowledge4igor:
In this PR I'm adding the implemetation of method for setting timeout in RequestBuilder

### DIFF
--- a/src/client/client.cc
+++ b/src/client/client.cc
@@ -745,6 +745,12 @@ RequestBuilder::body(std::string&& val) {
     return *this;
 }
 
+RequestBuilder&
+RequestBuilder::timeout(std::chrono::milliseconds timeout) {
+    timeout_ = timeout;
+    return *this;
+}
+
 Async::Promise<Response>
 RequestBuilder::send() {
     return client_->doRequest(std::move(request_), timeout_);


src/client/client.cc
```diff
@@ -745,6 +745,12 @@ RequestBuilder::body(std::string&& val) {
     return *this;
 }
 
+RequestBuilder&
+RequestBuilder::timeout(std::chrono::milliseconds timeout) {
```
dennisjenkins75: I've always treated the std::chrono types as opaque.  I actually do not know if "milliseconds" is a primitive type that fits into a 64-bit cpu register or a struct.  Should timeout be passed by value or const reference?

knowledge4igor: @dennisjenkins75 1) I have read this question on SO: [pass std::chrono::duration by value or by reference to const](https://stackoverflow.com/questions/37918168/pass-stdchronoduration-by-value-or-by-reference-to-const)? 2) I have printed out the size of this type (linux 64 bit):
```
std::cout << "ms size: " << sizeof(std::chrono::milliseconds) << "\n";
```
And got 8 bytes;
3) I have searched `std::chrono::milliseconds` parameters in methods in project and everywhere it's passed by value.
In my opinion it can be considered as a primitive and passed by value, also it will be consistent when other parts of codebase.

PR #279, opened 2018-08-27
Tests and refactoring net utils

knowledge4igor:
In this PR I have added tests and done small refactoring in net utils


src/common/net.cc
```diff
@@ -104,21 +104,21 @@ Address::port() const {
...
     const std::string portPart = addr.substr(pos + 1);
+    if (portPart.empty())
+        throw std::invalid_argument("Invalid port");
     long port = strtol(portPart.c_str(), &end, 10);
-    if (*end != 0 || port > Port::max())
+    if (*end != 0 || port < Port::min() || port > Port::max())
         throw std::invalid_argument("Invalid port");
-
-    host_ = std::move(host);
     port_ = port;
```
dennisjenkins75: You might want to make this "port_ = static_cast<uint16_t>(port);".  "port_" internally is uint16_t, and "port" is long.  You are doing proper range checking, but I can see a compiler throwing a warning about shoving a signed 32-bit int into an unsigned 16-bit int.

knowledge4igor: @dennisjenkins75 You are totally right!

PR #281, opened 2018-08-28
Added Access-Control-Allow-Headers to the set of headers

lcinacio:
Added Access-Control-Allow-Headers to the set of headers.
This is needed when calling endpoint from JavaScript in a browser, otherwise we run into an error in the OPTIONS request.


lcinacio:
Reverted changes to MaxBuffer and added unit test for this feature.

PR #285, opened 2018-09-01
Code refactoring: move port value convertion to string to Port class

knowledge4igor:
In this PR I have moved port value convertion to `std::string` to `Port` class and implemented tests. Also I have moved `*` to `0.0.0.0` convertion to `Address::init`.


xinthose:
Nice!

PR #286, opened 2018-09-01
Router Modification bug fix

Fylax:
Modified the TypedParam and how it is managed with Optional Parameters. Should fix the bug (not completely clear how to reproduce it)


CMakeLists.txt
```diff
@@ -8,9 +8,19 @@ option(PISTACHE_BUILD_TESTS "build tests alongside the project" OFF)
 option(PISTACHE_BUILD_EXAMPLES "build examples alongside the project" OFF)
 option(PISTACHE_INSTALL "add pistache as install target (recommended)" ON)
 
+CHECK_CXX_COMPILER_FLAG("-std=c++17" COMPILER_SUPPORTS_CXX17)
+CHECK_CXX_COMPILER_FLAG("-std=c++14" COMPILER_SUPPORTS_CXX14)
```
milipili: Why not simply rely on CMake for that?

```
set(CMAKE_CXX_STANDARD 17)
set(CMAKE_CXX_STANDARD_REQUIRED ON)
set(CMAKE_CXX_EXTENSIONS OFF)
```

dennisjenkins75: my POV: I do not want to require c++17 yet.  There are users of pistache on older platforms that can only do c++11.  Disclaimer: I do not know cmake; I do not know precisely what milipili's suggestion would do, but it reads to me as if it would begin requiring c++17.  Please correct me if I am incorrect.

Fylax: You are right. On the other hand I have checked CMAKE documentation and seems like that setting CXX standard to 17 but `CMAKE_CXX_STANDARD_REQUIRED` to OFF uses std-17 if available, otherwise decays on higher supported version. Seems like to be what we need.

milipili: you can replace 17 by 11 or 14. by anyway according to the code, staying in older versions doesn’t seem a very good idea to me, even if it sounds appealing for existing users

But the main point here is that we’re doing things manually where cmake has good features for a long time.

Fylax: The main point is to use C++ 17 if available, otherwise rollback on the highest available version.
While forcing C++17 may be tempting, it may cause compiling problems in some embedded systems with not so updated compilers, so we must take care of the problem

PR #287, opened 2018-09-02
issue 283 - runtimeexception when address in use

SnoozeTime:
Please kindly review :)

What do you thing about the strerror(errno)? Would it be better to send a hardcoded message instead?

Also, I sneakily added a fix for another issue (282) in listener.cc:180. I can remove if you need another PR for that.

Reference code: https://beej.us/guide/bgnet/html/multi/clientserver.html

Cheers


SnoozeTime:
Yes you're totally right. I'll push a fix for the FIX when I found what causes the test to fail :)

SnoozeTime:
**Why the test is failing -> use of global state to check whether the socket is bound**
A runtime exception is thrown from the Reaction::impl() method.
```c++
Reactor::Impl *
Reactor::impl() const {
    if (!impl_)
        throw std::runtime_error("Invalid object state, you should call init() before");

    return impl_.get();
}
```
In the test, this method should not have been called. Indeed we never instantiated the reactor because we couldn't even bind the socket... so what happened?

1. Run the first test listener_bind_port_free. In listener.cc, g_listen_fd will be set to whatever file descriptor number the socket is assigned to. (let's assume 4 here).
2. The destructor of listener is called, everything goes well.
3. Run the second test listener_bind_port_not_free_throw_runtime. The global variable is not reset (same process, same global variable...)
4. Exception is thrown as expected in bind but is caught to make sure the test is failing for the correct reason (look at the exception message in the ASSERT_STREQ).
5. Destructor of listener is called. listener::isBound is called BUT returns true instead of false as it is looking at the global variable (which value is still 4).
6. Then we try to shutdown the reactor but it has no implementation so the exception is thrown.

To check this easily, you can print the g_listen_fd in the listener destructor like the following:
```c++
Listener::~Listener() {
    if (isBound()) {
        std::cout << g_listen_fd << std::endl;
        std::cout << "isBound returned true\n";
        shutdown();
    }
    if (acceptThread) acceptThread->join();
}
```
Here, the test results will yield:
```
~/opt/pistache/build(283_runtimeexception_when_address_in_use*) » ./tests/run_listener_test                                          benoit@benoit-VirtualBox 
Running main() from gtest_main.cc      
[==========] Running 2 tests from 1 test case.                                 
[----------] Global test environment set-up.                                   
[----------] 2 tests from listener_test
[ RUN      ] listener_test.listener_bind_port_free                             
4                                      
isBound returned true                  
[       OK ] listener_test.listener_bind_port_free (1 ms)                      
[ RUN      ] listener_test.listener_bind_port_not_free_throw_runtime           
Address already in use                 
4                                      
isBound returned true                  
  what():  Invalid object state, you should call init() before                 
[1]    9701 abort (core dumped)  ./tests/run_listener_test           
```

**Proposed fix**
Two possible fixes:
- I'd rather use the listener fd_listen instead of the global g_fd_listen to check whether the socket is bound. 

From
```c++
bool
Listener::isBound() const {
    return g_listen_fd != -1;
}
```

To
```c++
bool
Listener::isBound() const {
    return listen_fd != -1;
}
```
After doing this change, all the tests are running fine.

- Modify the test to not catch the runtime except and just assert that an exception what thrown. This is not fixing the root cause though so the g_listen_fd problem will have to be addressed eventually.


What do you think?

SnoozeTime:
Cheers.
About last commit, the freeaddrinfo problem is already being fixed in another branch. If these commits are problematic, please let me know and I will create another pull request with a cleaner branch.

About the global variable, I can try to remove it after this pull request is merged. Global variable is also related to the signal handling (SIGINT).

SnoozeTime:
Fixed the conflict with latest master (CMakeLists conflict).

By the way, when running multiple times the test, http_client_test often fails with address already in use. Adding the flag Tcp::Options::ReuseAddr fix this problem..

Issue #282, opened 2018-08-31
Memory leak in Listener::bind


amirnik:
There is a addrinfo addrs struct  which is allocated but never free'd.

possible solution is to put: `freeaddrinfo(addrs);` after for loop @ line 187.

Cheers,
Amir

Issue #267, opened 2018-08-16
Valgrind reports memory leaks in mailbox.h


GMellar:
When I run a simple rest server with just one route and make a call to ist I get the following leaks: 

==20869== 64 bytes in 1 blocks are definitely lost in loss record 5 of 8
==20869==    at 0x4C2C21F: operator new(unsigned long) (vg_replace_malloc.c:334)
==20869==    by 0x2A879E: Pistache::Queue<Pistache::Tcp::Transport::TimerEntry>::Queue() (mailbox.h:178)
==20869==    by 0x2A69BB: Pistache::PollableQueue<Pistache::Tcp::Transport::TimerEntry>::PollableQueue() (mailbox.h:237)
==20869==    by 0x2A351A: Pistache::Tcp::Transport::Transport(std::shared_ptr<Pistache::Tcp::Handler> const&) (transport.cc:23)
==20869==    by 0x2B0A4B: Pistache::Tcp::Listener::bind(Pistache::Address const&) (listener.cc:191)
==20869==    by 0x2B0566: Pistache::Tcp::Listener::bind() (listener.cc:139)
==20869==    by 0x2AF803: void Pistache::Http::Endpoint::serveImpl<void (Pistache::Tcp::Listener::*)()>(void (Pistache::Tcp::Listener::*)()) (endpoint.h:69)
==20869==    by 0x2AF5B4: Pistache::Http::Endpoint::serveThreaded() (endpoint.cc:73)
==20869==    by 0x25D077: main (main.cpp:65)

There are a lot of them but they do not grow depending on the number of requests.

amirnik:
Found a solution just change the code to something like this:
mailbox.h:184

    virtual ~Queue() {
        while (auto *e = pop()) delete e;
        delete tail;
    }

dennisjenkins75:
Are you going to create a PR to implement your fix, and a unit-test to prove that it works?

amirnik:
Hi Dennis,
I am not planning  to create a PR.

PR #290, opened 2018-09-05
Fix memory leak 267 282

amirnik:
Issue #267:
   Freeing first element in the Queue constructor

Issue #282:
   Freeing allocated `addrs` structure after use.

### DIFF
--- a/src/server/listener.cc
+++ b/src/server/listener.cc
@@ -156,7 +156,7 @@ Listener::bind(const Address& address) {
     const auto& host = addr_.host();
     const auto& port = addr_.port().toString();
     struct addrinfo *addrs;
-    TRY(::getaddrinfo(host.c_str(), port.c_str(), &hints, &addrs));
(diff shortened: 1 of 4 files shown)


amirnik:
Hello,
These are small changes which should be easy to review.

Would it be possible for you guys to kindly look at it and discuss / merge if needed?

Best Regards,
Amir

src/server/listener.cc
```diff
@@ -174,6 +174,7 @@ Listener::bind(const Address& address) {
         TRY(::listen(fd, backlog_));
         break;
     }
+    freeaddrinfo(addrs);
```
dennisjenkins75: This works in the common case, but will still fail to clean up resources if the "::listen()" fails 2 lines above (the 'TRY' macro will throw a different exception).

I suggest that you encapsulate 'getaddrinfo' into a RAII class so that it will clean itself up no matter how the method exits.

amirnik:
For some reason build is not triggered on my commit...

PR #303, opened 2018-09-25
Issue 302 - Fix listener_test

SnoozeTime:
If the listener::init, call is not present, the listener object will be in a state where its workers_ member is not initialized (size_t). In clang, this leads to the creation of an absurd number of threads which in turn
will make eventfd call fails in NotifyFd. See issue 302 for more details.


SnoozeTime:
Tests are running happily with both clang and gcc.

PR #304, opened 2018-09-26
Streaming test

iroddis:
This PR resolves some issues where multiple threads write into a streaming response.

- transport.h:34  Removed immediate sending for consumer thread, as it could result in out of order writes (e.g. the chunked encoding termination write occurring before any queued chunks)
- transport.cc: Implemented the std::deque note from transport.h:34 for the toWrite object. This will maintain a deque of writes. Partial writes will be added to the front of the deque to preserve order.
- Added a streaming_writes test case.


iroddis:
I pulled master, and all tests pass on Debian stretch w/gcc 6.3.0.

PR #305, opened 2018-09-27
Initialize members of Listener with default value - Remove shared pointer 

SnoozeTime:
2 changes here:
- The first one is to initialize explicitly the members of Listener. This is to fix the problem that happens when Listener::init is not called before running (see issue 302).
- Replace shared_ptr<Aio::Reactor> by a value owned by Listener. The pointer was never shared with anything else so it makes sense to use a value instead of a shared_ptr for this case.

Please do not merge yet -> Will add a UT where listener is not init, but does not crash.

### DIFF
--- a/tests/listener_test.cc
+++ b/tests/listener_test.cc
@@ -123,6 +123,34 @@ TEST(listener_test, listener_bind_port_free) {
     ASSERT_TRUE(true);
 }
 
+// Listener should not crash if an additional member is added to the listener class. This test
+// is there to prevent regression for PR 303
+TEST(listener_test, listener_uses_default) {
+    uint16_t port_nb;
+
+    // This is just done to get the value of a free port. The socket will be closed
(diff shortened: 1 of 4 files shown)


SnoozeTime:
Added one test to prevent regression for the relevant issue. Please review :)

tests/listener_test.cc
```diff
@@ -123,6 +123,34 @@ TEST(listener_test, listener_bind_port_free) {
...
+    // after the closing curly bracket and the port will be free again (SO_REUSEADDR option).
+    // In theory, it is possible that some application grab this port before we bind it again...
+    {
+        SocketWrapper s = bind_free_port();
+        port_nb = s.port();
+    }
+
+
+    if (port_nb == 0) {
+        FAIL() << "Could not find a free port. Abord test.\n";
```
dennisjenkins75: "Abort", not "Abord"

SnoozeTime: Done!

PR #306, opened 2018-09-28
Improved routing algorithm and ability to add/remove routes at runtime

Fylax:
- Changed routing algorithm from linear search to tree-based
- Ignore multiple consecutive forward slashes
- Added ability to add and remove routes at runtime
- Added two new router tests


Fylax:
I fixed it. The problem was quite straightforward.
Basically, in C++17, there is an implicit conversion between `string` and `string_view`.
Since Clang exposes C++17 features in `std::experimental`, I forced it to use C++14 while GCC uses C++17.
I have redone all the tests forcing C++11 and everything works now.

Fylax:
Hello @janhkrueger 
In #322 I fixed the problem with older CMAKE versions.
Could you please check if in your debian system it works properly without any manual intervention?

PR #309, opened 2018-10-03
Fix for the issue 66

hasankandemir1993:
In the issue in https://github.com/oktal/pistache/issues/66, problem was we could not add cookies with same names, i changed the hashmap to list, so that we were able to see all cookies with same name. Beware though:  System wants to write cookies into list everytime we request cookies using request.cookies() method. I managed to solve this by not letting add cookies with exact same name and value. But this problem still stays in the system;

When we create a cookie with very low maxAge, wait  and request cookies, the cookie is gone, which is correct. But when we request cookies immediately, system adds this cookie to our list(previously hashmap), and this cookie stays forever in this list. This is a rare situation but should be solved.


hasankandemir1993:
> I'm not wild about changing the cookie jar algorithm from O(N*log(N)) to O(N^2). I want to think on this some more.

Yes you are right. What we could is, maybe we can do a hashmap of hashmaps. Like this;

key   ---------            Value
"name" -> ["value" -> CookieA, ( CookieA has the name of "name" and value of "value".)
                  "value2" -> CookieB]    ( CookieB has the name of "name" and value of "value2".)

"name2" ->  ["value" -> CookieC, ( CookieC has the name of "name2" and value of "value".)
                  "value2" -> CookieD]    ( CookieD has the name of "name2" and value of "value2".)


But in order to do this, without changing any signature, we will need to keep a second internal iterator to go through the internal hashmap. So user can get one cookie at a time when he/she requests for request.cookies() .

PR #310, opened 2018-10-04
Code improvements

knowledge4igor:
In this PR have added using 'std::make_shared' instead of reset call (why it's better you can read [here](https://stackoverflow.com/questions/14836691/is-it-better-to-use-shared-ptr-reset-or-operator)). Also I have changed return value of `Listener::bind` methond because returning `bool` value looks obsolete.


SnoozeTime:
Just one thought: right now Listener is using shared pointers but it seems that the transport is owned by the listener (as well as the reactor object that is using the transport as well). What about using unique_ptr for listener and just give a raw pointer to the reactor?

PR #312, opened 2018-10-07
Ghost cookie problem, hashmap of hashmaps solution

hasankandemir1993:
in the pull request 309, i used list instead of previous hashmap, but it was inefficient, but now i implemented hashmap of hashmaps. Outer hashmap's key will be "name" of the cookie. Inner hashmaps' key will be "values" of the cookies. So there can be multiple cookies with same name, but there can only be one cookie with exact same name and value.

I also mentioned about much deeper problems in cookies in pull request 309, please refer to this.

You can change my code as you see fit.


include/pistache/cookie.h
```diff
@@ -41,25 +43,46 @@ struct Cookie {
 
 class CookieJar {
 public:
-    using Storage = std::unordered_map<std::string, Cookie>;
+    typedef std::unordered_map<std::string, Cookie> HashMapCookies; // "value" -> Cookie
```
dennisjenkins75: Please use "using type = def;" instead of "typedef def type;"

include/pistache/cookie.h
```diff
@@ -73,18 +96,22 @@ class CookieJar {
 
     private:
         Storage::const_iterator it_;
+        HashMapCookies::const_iterator it_2;
```
dennisjenkins75: Please use a more descriptive name for these three private member variables.  Suggestion: "iter_storage", "iter_cookie_values", "iter_storage_end".

include/pistache/cookie.h
```diff
@@ -73,18 +96,22 @@ class CookieJar {
...
+        Storage::const_iterator end_; // we need to know where main hashmap ends. 
     };
 
     CookieJar();
 
     void add(const Cookie& cookie);
+    void removeCookie(const std::string& name);  // ADDED LATER
```
dennisjenkins75: The comment should be "Unimplemented", not "[to be] added later".  In fact, I would comment out the entire line until you have an implementation.

include/pistache/cookie.h
```diff
@@ -73,18 +96,22 @@ class CookieJar {
...
+    
     void addFromRaw(const char *str, size_t len);
     Cookie get(const std::string& name) const;
 
     bool has(const std::string& name) const;
 
     iterator begin() const {
-        return iterator(cookies.begin());
+        return iterator(cookies.begin(),cookies.end());
```
dennisjenkins75: Please be consistent with existing code style.  Add a space before the comma.

src/common/cookie.cc
```diff
@@ -103,6 +105,8 @@ Cookie::Cookie(std::string name, std::string value)
 Cookie
 Cookie::fromRaw(const char* str, size_t len)
 {
+    std::cout<<"From Raw initiated with" << str <<" " << endl;
```
dennisjenkins75: Please be consistent with the existing coding style, and consistent within your own line of code.  The "<<" operator should have a space on either side of it.

src/common/cookie.cc
```diff
@@ -163,6 +167,7 @@ Cookie::fromRaw(const char* str, size_t len)
 
 Cookie
 Cookie::fromString(const std::string& str) {
+	std::cout<<"From String initiated with" << str <<" " << endl;
```
dennisjenkins75: Code style consistency.  Space before and after operators.

src/common/cookie.cc
```diff
@@ -207,7 +212,24 @@ CookieJar::CookieJar()
...
+        HashMapCookies hashmapWithFirstCookie;
+        hashmapWithFirstCookie.insert(std::make_pair(cookieValue,cookie));
+        cookies.insert(std::make_pair(cookieName, hashmapWithFirstCookie));
+    } else {
+        it->second.insert(std::make_pair(cookieValue,cookie));
+    }
+
+}
+
+void // ADDED later with fix
```
dennisjenkins75: Comment should be "Unimplemented" (give current state explicitly).

hasankandemir1993:
Hi, thank you for commenting, i will fix the code according to your suggestions.

hasankandemir1993:
I changed the code according to your suggestions.

PR #313, opened 2018-10-07
Add address reuse in http client test

knowledge4igor:
This is the first part of changes for fixing http client test. We definitelly should create server with `Tcp::Options::ReuseAddr` option. Now I'm researching the roots of the following problem:
```
10: [ RUN      ] request_builder.multiple_send_test
10: terminate called after throwing an instance of 'Pistache::Http::HttpError'
10:   what():  Encountered invalid HTTP version
1/1 Test #10: http_client_test .................***Exception: Child aborted  0.50 sec
```
This problem was raised after my [commit](https://github.com/oktal/pistache/commit/11bba3e994d33f1e9b476b15a2f8f3e8c3d1bd65) when I added http client test for checking multiple send method calls.

### DIFF
--- a/src/server/endpoint.cc
+++ b/src/server/endpoint.cc
@@ -14,6 +14,8 @@ namespace Http {
 
 Endpoint::Options::Options()
     : threads_(1)
+    , flags_()
+    , backlog_(Const::MaxBacklog)
     , maxPayload_(Const::DefaultMaxPayload)
 { }
 
--- a/src/server/endpoint.cc
+++ b/src/server/endpoint.cc
@@ -14,6 +14,7 @@ namespace Http {
 
 Endpoint::Options::Options()
     : threads_(1)
+    , maxPayload_(Const::DefaultMaxPayload)
 { }
 
 Endpoint::Options&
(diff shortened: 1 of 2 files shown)


knowledge4igor:
Also we must initialize `maxPayload_` field of `Options` class with default value

src/server/endpoint.cc
```diff
@@ -14,6 +14,7 @@ namespace Http {
 
 Endpoint::Options::Options()
     : threads_(1)
+    , maxPayload_(Const::DefaultMaxPayload)
```
dennisjenkins75: I thought that we submitted a PR a few weeks ago to fully initialize ALL members of Endpoint::Options.  Oh well.

Igor, I know that this request is slightly beyond the scope of this PR, but please init all 4 members to sane defaults, and in the same order that they are declare in the struct (or "-Weffc++" and cppcheck will complain about the init order).

knowledge4igor: @dennisjenkins75 I have done it. Now I'm exploring the situation with `http_client_test` test and I have already found the root of problem. Hope in the next PR I will fix it.

PR #315, opened 2018-10-09
Improve Client and Endpoint init API and tiny changes in tests

knowledge4igor:
In this PR I'm proposing tiny improvements in `init` method of `Client` and `Endpoint` classes. I have some doubts about setting `InstallSignalHandler` and `ReuseAddr` flags as defaults ones on server, but it looks reasonable:
1) It's ok by default to reuse address for listening;
2) It's ok to catch `SIGINT` by default.

### DIFF
--- a/src/server/endpoint.cc
+++ b/src/server/endpoint.cc
@@ -14,7 +14,7 @@ namespace Http {
 
 Endpoint::Options::Options()
     : threads_(1)
-    , flags_()
+    , flags_(Tcp::Options::InstallSignalHandler | Tcp::Options::ReuseAddr)
     , backlog_(Const::MaxBacklog)
     , maxPayload_(Const::DefaultMaxPayload)
 { }
--- a/include/pistache/client.h
+++ b/include/pistache/client.h
@@ -306,7 +306,6 @@ public:
            : threads_(Default::Threads)
            , maxConnectionsPerHost_(Default::MaxConnectionsPerHost)
            , keepAlive_(Default::KeepAlive)
-
        { }
 
        Options& threads(int val);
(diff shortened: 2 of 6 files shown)


src/server/endpoint.cc
```diff
@@ -14,7 +14,7 @@ namespace Http {
 
 Endpoint::Options::Options()
     : threads_(1)
-    , flags_()
+    , flags_(Tcp::Options::InstallSignalHandler | Tcp::Options::ReuseAddr)
```
dennisjenkins75: This is a behaviour change.  I have mixed feelings about it.  I will likely merge this PR, but be prepared if others want this reverted.  However, IMHO, users should be explicitly setting flags.  I do know that at least one user of Pistache has their own SIGINT handler, and this change will break their code.  IDR who they were though.

What is your rationale for making this change?

knowledge4igor: @dennisjenkins75 The rationale for making this change is the following:

1. I have seen several servers and most of time `SO_REUSEADDR` option is set when listen socket is configured., because you can stop server and run it immediately after that and it would be better have no problems with starting again by **default**, without this option.

2. I consider `Endpoint` as a high level (it means that we have made some desicion for users) building block in Pistache and it ok to handle by **default** CTRL + C to free resources. It looks natual. For instance, If I'm not mistaken in [Twisted](https://twistedmatrix.com/) reactor loop has default `SIGINT` handler. On the other hand, if you know users that use their own `SIGINT` handler, it would be better to back out my change.

Ok, I will revert the part of commit that is related to `Endpoint` soon.

knowledge4igor:
@dennisjenkins75 I have reverted setting `Endpoint` default flags

PR #320, opened 2018-10-13
Fix server crashes: abort on epoll_ctl and multiple memory deletion

knowledge4igor:
In this PR I'm trying to fix server crashes: abort on epoll_ctl and multiple memory deletion. Also I have added several server tests.


tests/http_server_test.cc
```diff
@@ -0,0 +1,225 @@
...
+            std::this_thread::sleep_for(std::chrono::seconds(delay_));
+        }
+        writer.send(Http::Code::Ok, "Hello, World!");
+    }
+
+    int delay_;
+};
+
+TEST(http_server_test, client_disconnection_on_timeout_from_single_threaded_server) {
+    const std::string address = "localhost:9095";
```
dennisjenkins75: This is fine for now, but we should generate these port numbers dynamically.  Maybe set the port to 0, let the kernel choose it, add an api to pistache (if there is not one already) to ask what port number is in use, and pass that to the client (in the unit test).

These unit tests will fail if the port number is already bound on the test system, which makes these tests non-hermetic.

knowledge4igor: @dennisjenkins75 Yes, you are totally right, I will think about this problem.

PR #340, opened 2018-10-26
Fix promise

mnottale:
Fix multiple issues encountered when chaining promises.

### DIFF
--- a/include/pistache/async.h
+++ b/include/pistache/async.h
@@ -549,11 +549,7 @@ namespace Async {
 
                 void doReject(const std::shared_ptr<CoreT<T>>& core) {
                     reject_(core->exc);
-                    for (const auto& req: core->requests) {
-                        req->reject(core);
-                    }
                 }
-
 
                 template<
                     typename Promise,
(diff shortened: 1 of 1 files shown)


xurxoham:
The example below does not work with neither master nor with the pull request's branch, when `USE_VOID` is defined:
```
#include <pistache/async.h>
#include <iostream>
#include <future>

#if USE_VOID
typedef void Placeholder;
#define PH_Value
#else
struct Placeholder {};
#define PH_Value Placeholder{}
#endif

using namespace Pistache;

int main() {
    Async::Promise<Placeholder> a(
        [](Async::Resolver& success, Async::Rejection& failure) {
            std::cout << "0\n";
            return success(PH_Value);
        });

    Async::Promise<Placeholder> b = a
            .then(
                [](Placeholder){
                    std::cout << "1\n";
                    return PH_Value;
                },
                Async::IgnoreException)

            .then(
                [](Placeholder){
                    std::cout << "2\n";
                    return PH_Value;
                },
                Async::IgnoreException);


    std::async(
        [&]() {
            a.resolved(PH_Value);
        });

    Async::Barrier<Placeholder> barrier(b);
    std::cv_status status = barrier.wait_for(std::chrono::seconds(5));

    if( status == std::cv_status::timeout ) {
        std::cout << "Failure: barrier timed out\n";
        return 1;
    }
    return 0;
}
```

PR #345, opened 2018-11-07
Improve error handling on accept call

knowledge4igor:
In this PR I'm proposing inprovement of `accept` call on listening socket. In my opinition not all accept errors should cause server shutdown.

### DIFF
--- a/include/pistache/errors.h
+++ b/include/pistache/errors.h
@@ -1,0 +1,23 @@
+#pragma once
+
+#include <stdexcept>
+
+namespace Pistache {
+namespace Tcp {
+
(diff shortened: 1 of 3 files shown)


knowledge4igor:
@dennisjenkins75 I have made a change in http_server_test because it seems that there is an error in server that makes this test unstable. I will explore and try to fix this problem in the near future.

knowledge4igor:
@dennisjenkins75 I have fixed problem with http_server_test in this [PR](https://github.com/oktal/pistache/pull/347) and decided to revert temporary commit here.

PR #355, opened 2018-11-18
Fix cookies sending and getting on client side

knowledge4igor:
I this PR I'm trying to fix two problem that I have found in Pistache:
1) On client request cookies should be sent in the following way:
```
...
Cookie: key1=value1; key2=value2
...
```
2) Client cannot get cookies from server response:
```
...
Set-Cookie: key1=value1
Set-Cookie: key2=value2
...
```

I'm not an expert in this topic, please check me carefully.

Reference: [HTTP cookies](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies)


knowledge4igor:
@dennisjenkins75 Hello, can you take a look at this PR, is it ok or there are doubts that I'm right?

knowledge4igor:
@dennisjenkins75 Thank you so much

PR #362, opened 2018-11-26
Client code refactoring

knowledge4igor:
In this PR I have made the following things that make the code better (from my point of view):
1) Replace `Buffer` to `std::string`. I think it would be better to use STL containers in **this case** instead of implementing your own.
2) Replace `DynamicStreamBuf` to `std::stringstream` and checking this stream in one place;
3) Remove from client explicit `delete[]` call. In modern C++ it would be better to avoid using raw pointers, explicit `new `and `delete` calls.
3) Fix several warnings

### DIFF
--- a/include/pistache/client.h
+++ b/include/pistache/client.h
@@ -5,11 +5,6 @@
 */
 
 #pragma once
-#include <atomic>
-#include <deque>
-
-#include <sys/types.h>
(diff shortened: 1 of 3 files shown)


xinthose:
I agree with using `std::string` and smart pointers.  Avoid `new` and `delete` at all costs in C++.

PR #366, opened 2018-12-03
Added IPv6 support

hydratim:
In this commit I've added IPv6 support and done some minor clean-up of the socket & address stack.

### DIFF
--- a/src/common/http_header.cc
+++ b/src/common/http_header.cc
@@ -340,17 +340,32 @@ Host::Host(const std::string& data)
 
 void
 Host::parse(const std::string& data) {
-    auto pos = data.find(':');
-    if (pos != std::string::npos) {
-        std::string h = data.substr(0, pos);
-        int16_t p = std::stoi(data.substr(pos + 1));
-
-        host_ = h;
-        port_ = p;
+    unsigned long pos = data.find(']');
+    unsigned long s_pos = data.find('[');
+    if (pos != std::string::npos && s_pos != std::string::npos) {
+        //IPv6 address
+        host_ = data.substr(s_pos+1, pos-1);
+        pos++;
     } else {
-        host_ = data;
-        port_ = HTTP_STANDARD_PORT;
+        //IPv4 address
+        pos = data.find(':');
+        if (pos == std::string::npos) {
+            host_ = data;
+            port_ = HTTP_STANDARD_PORT;
+        }
+        host_ = data.substr(0, pos);
+        if (host_ == "*") {
+            host_ = "0.0.0.0";
+        }
     }
+    char *end;
+    const std::string portPart = data.substr(pos + 1);
+    if (portPart.empty())
+        throw std::invalid_argument("Invalid port");
+    long port = strtol(portPart.c_str(), &end, 10);
+    if (*end != 0 || port < Port::min() || port > Port::max())
+        throw std::invalid_argument("Invalid port");
+    port_ = static_cast<uint16_t>(port);
 }
 
 void
--- a/src/common/net.cc
+++ b/src/common/net.cc
@@ -52,12 +52,71 @@ Ipv4::any() {
 
 std::string
 Ipv4::toString() const {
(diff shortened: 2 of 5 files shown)


src/common/net.cc
```diff
@@ -52,12 +52,71 @@ Ipv4::any() {
...
+    inet_ntop(AF_INET, &addr, buff, INET_ADDRSTRLEN);
+    
+    return std::string(buff);
+}
 
-    char buff[MaxSize];
-    snprintf(buff, MaxSize, "%d.%d.%d.%d", a, b, c, d);
+void Ipv4::toNetwork(in_addr_t *addr) const {
+    // Bitshift the bytes into an in_addr_t (a single 32bit unsigned int);
+    *addr = (uint32_t)(d<<24) | (uint32_t)(c<<16) | (uint32_t)(b<<8) | (uint32_t)a;
```
dennisjenkins75: Will this behavior be correct on a big-endian cpu?

src/common/http_header.cc
```diff
@@ -340,15 +341,33 @@ Host::Host(const std::string& data)
 
 void
 Host::parse(const std::string& data) {
```
dennisjenkins75: Please add a set of unit tests (including sending in garbage data) for Host::parse(), or extend any existing unit test to cover IPV6 addresses and mal-formed IPV4 and IPV6 addresses.

src/common/net.cc
```diff
@@ -52,12 +52,71 @@ Ipv4::any() {
...
+void Ipv6::toNetwork(in6_addr *addr6) const {
+    uint16_t temp_ip6[8] = {a, b, c, d, e, f, g, h};
+    uint16_t remap_ip6[8] = {0};
+    uint16_t x, y;
+    
+     // If native endianness is little-endian swap the bytes, otherwise just copy them into the new array
+    if ( htonl(1) != 1 ) {
+        for (uint16_t i = 0; i<8; i++) {
+            x = temp_ip6[i];
+            y = __builtin_bswap16 (x);
```
dennisjenkins75: Is there a reason why you did not use "htons()" here?  Under gcc and clang "htons()" might be an alias for "__builtin_bswap16()".  Since this is networking code, "htons()" seems more natural.  Just curious.  Since pistache uses Linux's eventfd, it is unlikely that this code will ever be compiled by something other than gcc or clang (I highly douct that msvc has this builtitn).

src/common/net.cc
hydratim: That's a good question, the short answer is that I'm not certain. I'll double check.

src/common/net.cc
hydratim: htons is also fine I'll swap that in to avoid any potential confusion, but AFAIK this is supported in current MSVC (at least that seems to be indicated by a similar discussion on ardupilot).

src/common/http_header.cc
hydratim: Will do

src/common/net.cc
hydratim: I haven't gotten around to validating if this is actually a problem as I don't have a big-endian system at home, however something like:
`*addr = htonl( (uint32_t)(a<<24) | (uint32_t)(b<<16) | (uint32_t)(c<<8) | (uint32_t)d );`
Should ensure that the resulting number is compatible across all architectures by making it the compiler's responsibility (notice the change in the ordering of the bytes). 
If that's an acceptable solution I'll put it in a commit, otherwise I'll see if I can come up with a cleaner way.

hydratim:
Still need to write unit tests.

knowledge4igor:
@dennisjenkins75 There is a flaky problem with Client with handling response that I have already been discovering for 3 weeks (one problem with handling I have  fixed).

hydratim:
I've still got more tests to write anyway, so don't worry too much about it yet. I just had five minutes last night so I got the bare minimum together.

hydratim:
Hi @dennisjenkins75,

I've written all the tests I can think of for now, please review and merge.

Thanks,
Tim

PR #385, opened 2018-12-20
Ephemeral ports

dennisjenkins75:
Adds ability to use an ephemeral TCP port for pistache.  Just specify "Port(0)" when creating an Endpoint.  The socket is created inside Listener::run() (like normal).  However, after the socket is created, you can obtain the actual port by calling Endpoint::getPort().

Caveat: Listener::run() creates the socket AND runs the event loop until Endpoint::shutdown() is called.  Endpoint::serve() directly calls Listener::run(), so your calling thread cannot get the port (since its all the same thread).  If your application runs the endpoint if a different thread (due to calling Endpoint::serveThreaded()), and port is not available until AFTER the socket is created, you will need to poll or have some other way to know when the socket is created.

IMHO, "Endpoint::serveThreaded()" should not return until the endpoint is fully initialized and ready to go, OR there should be another method in Endpoint that blocks until the worker thread and socket are ready.  I might add that in a future PR.

### DIFF
--- a/include/pistache/endpoint.h
+++ b/include/pistache/endpoint.h
@@ -53,6 +53,10 @@ public:
         return listener.isBound();
     }
 
+    Port getPort() const {
+        return listener.getPort();
+    }
+
     Async::Promise<Tcp::Listener::Load> requestLoad(const Tcp::Listener::Load& old);
 
     static Options options();
--- a/include/pistache/listener.h
+++ b/include/pistache/listener.h
@@ -53,6 +53,7 @@ public:
     void bind(const Address& address);
 
     bool isBound() const;
+    Port getPort() const;
 
     void run();
     void runThreaded();
--- a/src/server/listener.cc
+++ b/src/server/listener.cc
@@ -184,7 +184,7 @@ Listener::bind(const Address& address) {
         TRY(::listen(fd, backlog_));
         break;
     }
-    
+
     // At this point, it is still possible that we couldn't bind any socket. If it is the case, the previous
     // loop would have exited naturally and addr will be null.
     if (addr == nullptr) {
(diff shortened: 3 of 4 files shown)


kiplingw:
To anyone else in the meantime who would like to have serveThreaded() attempt to bind() and fully initialize before returning, as well as attempt to re-bind if the address is already in use, this is a work around I came up with last night:

```
// Attempt to bind daemon to port, trying again intelligently if its still in
//  use after a period of time...
void Bind()
{
    // Timeout value to use between rebind attempts...
    size_t RetryTimeout(30);

    // Attempt to query kernel for the length of time an orphaned (no longer
    //  referenced by any application) connection will remain in the FIN_WAIT_2
    //  state before it is aborted at the local end...
    //  <https://www.kernel.org/doc/Documentation/networking/ip-sysctl.txt>
    ifstream KernelTCPWaitTimeoutFile("/proc/sys/net/ipv4/tcp_fin_timeout");
    if(KernelTCPWaitTimeoutFile)
        KernelTCPWaitTimeoutFile >> RetryTimeout;

    // Flag on whether we bound successfully...
    bool Successful = false;

    // Try several times if the address is already in use until giving up...
    for(size_t Retries = 0; Retries < 3; ++Retries)
    {
        try
        {
            // Try to bind...
            m_Endpoint->serveThreaded();

            // Port still isn't bound...
            while(!m_Endpoint->isBound())
            {
                // Wait a second before checking again or Pistache throws an
                //  exception on an EADDRINUSE...
                this_thread::sleep_for(chrono::seconds(1));
            }

            // We bound successfully...
            Successful = true;

            // No need to try again...
            break;
        }

        // An exception occurred...
        catch(const runtime_error &Error)
        {
            // Address already in use...
            if(errno == EADDRINUSE)
            {
                // Log it...
                P_WARNING_MSG(_("the port is already in use, will retry momentarily"))

                // Wait a moment...
                this_thread::sleep_for(chrono::seconds(RetryTimeout));

                // Try again...
                continue;
            }

            // Some other kind of problem, pass the exception up the chain...
            else
                throw Error;
        }
    }

    // If we didn't succeed, raise an error...
    if(!Successful)
        throw runtime_error(_("unable to bind to port, even after retrying"));
}
```

PR #386, opened 2018-12-21
Improve use of enumeration sets (flags)

xurxoham:
Removes the necessity of declaring enumerators with single
bit set values (e.g. 1<<X) and the requirement of a "None"
special value (there is no such a thing as TCP option None).

Instead, operations that depend on a set of these
enumerators take a Flags<Enum> argument, which is now
implemented using a std::bitset.

Future work may include specialization of Flags such as
NotifyOn, where the value and operations make use of the
underlying C macros instead of an intermediate
bitset representation.

Possible optimizations would consist in turning the constructors
and other simple operations constexpr, although this is restricted
to C++14 or later (no loops in C++11 constexpr functions).

Type trait Pistache::Underlying_Type removed since GCC 4.6 does
not support C++11 (4.8.1 or greater).

### DIFF
--- a/include/pistache/flags.h
+++ b/include/pistache/flags.h
@@ -7,145 +7,90 @@
 #pragma once
 
 #include <type_traits>
+#include <bitset>
 #include <climits>
 #include <iostream>
 
 namespace Pistache {
 
-// Looks like gcc 4.6 does not implement std::underlying_type
-namespace detail {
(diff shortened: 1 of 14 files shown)


xurxoham:
Hi, no worries. I understand the burden and time it takes to maintain a repository!
I changed projects a while back, so I won't have time for this unfortunately.
I'll close the PR myself then.

PR #388, opened 2018-12-26
IPv6 Bug Fix

hydratim:
I missed something in yesterday's commit, it only showed up when I tried to actually use the merged library. 
I've fixed it an added more tests that will catch this kind of bug.

### DIFF
--- a/src/server/listener.cc
+++ b/src/server/listener.cc
@@ -161,8 +161,10 @@ Listener::systemSupportsIpv6(){
                 continue;
 
             family = ifa->ifa_addr->sa_family;
-            if (family == AF_INET6)
+            if (family == AF_INET6) {
+                supportsIpv6 = true;
                 continue;
+            }
     }
     freeifaddrs(ifaddr);
     return supportsIpv6;


tests/listener_test.cc
```diff
@@ -123,6 +123,60 @@ TEST(listener_test, listener_bind_port_free) {
...
+
+    Pistache::Tcp::Listener listener;
+    Pistache::Flags<Pistache::Tcp::Options> options;
+    listener.init(1, options);
+    listener.setHandler(Pistache::Http::make_handler<DummyHandler>());
+    listener.bind(address);
+    ASSERT_TRUE(true);
+}
+
+TEST(listener_test, listener_bind_v6_port) {
```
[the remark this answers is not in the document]
dennisjenkins75: I would be fine if pistache itself had a utility method, maybe "static bool Listener::systemSupportsIpV6()", and we use that in relevant unit tests.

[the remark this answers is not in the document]
hydratim: Sounds good, I'll add that in the next commit.

tests/listener_test.cc
```diff
@@ -123,6 +123,60 @@ TEST(listener_test, listener_bind_port_free) {
...
+    uint16_t port_nb;
+
+    // This is just done to get the value of a free port. The socket will be closed
+    // after the closing curly bracket and the port will be free again (SO_REUSEADDR option).
+    // In theory, it is possible that some application grab this port before we bind it again...
+    {
+        SocketWrapper s = bind_free_port();
+        port_nb = s.port();
+    }
+
```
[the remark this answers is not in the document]
hydratim: I'll take a look at your new tests and adjust accordingly.

hydratim:
@dennisjenkins75  All done and rebased to include your changes to the test systems, please review and merge at your discretion.

src/server/listener.cc
```diff
@@ -145,6 +147,27 @@ Listener::pinWorker(size_t worker, const CpuSet& set)
...
+    for (ifa = ifaddr, n = 0; ifa != NULL; ifa = ifa->ifa_next, n++) {
+            if (ifa->ifa_addr == NULL)
+                continue;
+
+            family = ifa->ifa_addr->sa_family;
+            if (family == AF_INET6)
+                continue;
+    }
+    freeifaddrs(ifaddr);
+    return supportsIpv6;
```
dennisjenkins75: You never actually set supportsIpv6 to true, and that's the most important part!

tests/listener_test.cc
```diff
@@ -192,3 +194,20 @@ TEST(listener_test, listener_bind_ephemeral_port) {
...
     ASSERT_TRUE(bound_port > (uint16_t)0);
 }
+
+
+TEST(listener_test, listener_bind_ephemeral_v6_port) {
+    Pistache::Port port(0);
+    Pistache::Address address(Pistache::Ipv6::any(), port);
+
+    Pistache::Tcp::Listener listener;
+    if (not listener.systemSupportsIpv6()) {
```
dennisjenkins75: TIL: https://en.cppreference.com/w/cpp/language/operator_alternative
I had no idea that this was valid c++.

dennisjenkins75: No need to remove it, its valid.  Just a funny comment...

src/server/listener.cc
hydratim: Doh! Looks like that got dropped during some minor refactoring before submission.
Fixed now.

tests/listener_test.cc
hydratim: It's something I've carried across from my heavy use of Python in other projects, I find it makes it easier to read when skimming over a large block.

hydratim:
@dennisjenkins75 you should have an e-mail from tim@plain... now.

That does look on the surface like an IPv6 issue, but it might be originating from the host libc implementation - If there is an issue open tag me in it and I'll dig into it tomorrow.

PR #401, opened 2019-01-01
Fixes memory leak in Listenter::bind()

dennisjenkins75:
Listener::bind() calls "getaddrinfo()", which allocates memory on the heap.  Later, Listenter::bind() has an exit path that fails to call "freeaddrinfo()".  I solved this by wrapping the "addr info" in a class that frees the addrinfo on destruction and provides read-only access to the same during its lifetime.

Addresses one memory leak from issue #399.

### DIFF
--- a/include/pistache/net.h
+++ b/include/pistache/net.h
@@ -26,6 +26,42 @@
 
 namespace Pistache {
 
+// Wrapper around 'getaddrinfo()' that handles cleanup on destruction.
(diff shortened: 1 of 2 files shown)


knowledge4igor:
@dennisjenkins75 I think your fix it pretty ok, only, it would be better to add this change to client.cc file too, as I have already written.

PR #413, opened 2019-01-21
Macro for simple HTTP headers

tete17:
This PR creates a Macro `CUSTOM_HEADER` for very simple HTTP Headers.

This has the benefit of still preserving the type safety of the library design as well as providing a quick and easy way to add custom headers without the hassle of writing the whole class structure.

I thought this PR is appropriate since I have seen many of this simplifications across the library like the function `Pistache::Http::listenAndServe` which improves the empty editor to hello world time.

I would recommend to review each commit on its own since the small changes may pollute a bit the editor. If you think the small changes should not be there by any reason please let me know and I can take them out.

### DIFF
--- a/include/pistache/http_header.h
+++ b/include/pistache/http_header.h
@@ -552,6 +552,25 @@ private:
     std::string ua_;
 };
 
+#define CUSTOM_HEADER(header_name) \
+    class header_name : public Pistache::Http::Header::Header {     \
+    public:                                                         \
+		    NAME(#header_name)                                          \
+                                                                    \
+		    header_name() = default;                                    \
+		                                                                \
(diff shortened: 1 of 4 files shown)


tete17:
I have kept the feature commit and the typos fixes.
Let me know if you need more explaining

Sorry for the previous pollution, I will be creating another PR for the `clang-tidy` fixes.


PS: Even Though is not relevant anymore the changes related with r-values and `std::move` were because the structs that affected where trivially copyable or the receiving functions took argument as `const &` and no move was happening.

tete17:
Silly mistake by my side to not test for extensively

Tell me if you think more unit tests are required.

PR #417, opened 2019-01-25
Implementation of Date::write

arthurafarias:
Hello

I was using Date::write. So I contributed implementing it. Maybe all methods that aren't implemented should emit a warning.

Arthur

### DIFF
--- a/tests/headers_test.cc
+++ b/tests/headers_test.cc
@@ -5,6 +5,7 @@
 
 #include <algorithm>
 #include <chrono>
+#include <iostream>
 
 using namespace Pistache::Http;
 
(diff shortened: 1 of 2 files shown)


arthurafarias:
This is the result

image

arthurafarias:
👌

tests/headers_test.cc
```diff
@@ -238,6 +239,18 @@ TEST(headers_test, date_test_asctime) {
     ASSERT_EQ(dd3, expected_time_point);
 }
 
+TEST(headers_test, date_test_ostream) {
+
+    using namespace std::chrono;
+
+    std::ostringstream os;
```
dennisjenkins75: I think that "std::chrono: namespace is unused here?

arthurafarias: I've removed this namespace in next commit.

PR #429, opened 2019-02-03
Fix memory defects in net

knowledge4igor:


src/common/net.cc
```diff
@@ -238,8 +238,8 @@ Address::init(const std::string& addr) {
         family_ = AF_INET6;
         try {
             in6_addr addr6;
-            char buff6[INET6_ADDRSTRLEN+1];
-            memcpy(buff6, host_.c_str(), INET6_ADDRSTRLEN);
+            char buff6[INET6_ADDRSTRLEN + 1] = {0, };
+            std::copy(&host_[0], &host_[0] + host_.size(), buff6);
```
dennisjenkins75: Just curious: Is there a reason why you did not use "host_.begin()" and "host_.end()" ?

knowledge4igor: @dennisjenkins75 Dennis, thank you so much for this remark, you are totally right. It's better to to use iterators - C++ way :) I have made yet another commit. Also I will refactor code in the next PR to reduce code duplication (net.cc and http_header.cc)

PR #430, opened 2019-02-03
Fix issues in test_memcheck

AdrianCX:
Remaining issues in test_memcheck

Follow up for: https://github.com/oktal/pistache/issues/399
No issues remaining in 'test_memcheck' after this. (note: hopefully test coverage is sufficient)

1. include/pistache/async.h
- Memory leak - data allocated in 'Core' was never freed.

2.include/pistache/client.h
- Leak of ' const struct sockaddr* addr' - if the leak is fixed then this becomes an 'invalid read'
- Solution - make a local storage in the class and copy data over.

3. include/pistache/optional.h
- Uninitialized data read - data() is uninitialized if nothing was constructed there - the condition was dangerous - copied logic from normal operator= in same class.

4. src/client/client.cc
- Fix continuation for item 2.
- Fix actual leak source -> getaddrinfo leaked the 'addrs' - that needs to be freed. Good thing we have 'AddrInfo' class in pistache to do auto-cleanup when out of scope

5. src/common/net.cc
- Invalid read - buf may be smaller then 'INET6_ADDRSTRLEN' or the v4 version

6. src/common/transport.cc
- Dangling pointer - this is really scary, hopefully there aren't any others in the code.
- wq.pop_front(); - destroyes buffer which is just a reference to a reference: 
```
        auto & entry = wq.front();
        const BufferHolder &buffer = entry.buffer;
```
- code after that accessed buffer which was deleted.

### DIFF
--- a/include/pistache/async.h
+++ b/include/pistache/async.h
@@ -171,13 +171,15 @@ namespace Async {
 
         struct Core {
             Core(State _state, TypeId _id)
-                : state(_state)
+                : allocated(false)
+                , state(_state)
                 , exc()
                 , mtx()
                 , requests()
                 , id(_id)
             { }
 
+            bool allocated;
             State state;
             std::exception_ptr exc;
 
@@ -210,11 +212,18 @@ namespace Async {
                 }
 
                 void *mem = memory();
+
+                if (allocated) {
+                    reinterpret_cast<T*>(mem)->~T();
+                    allocated = false;
+                }
+
                 new (mem) T(std::forward<Args>(args)...);
+                allocated = true;
                 state = State::Fulfilled;
             }
 
-            virtual ~Core() { }
+            virtual ~Core() {}
         };
 
         template<typename T>
@@ -224,13 +233,17 @@ namespace Async {
                 , storage()
             { }
 
+            ~CoreT() {
+                if (allocated) {
+                    reinterpret_cast<T*>(&storage)->~T();
+                    allocated = false;
+                }
+            }
+            
             template<class Other>
             struct Rebind {
                 typedef CoreT<Other> Type;
             };
-
-            typedef typename std::aligned_storage<sizeof(T), alignof(T)>::type Storage;
-            Storage storage;
 
             T& value() {
                 if (state != State::Fulfilled)
@@ -241,9 +254,14 @@ namespace Async {
 
             bool isVoid() const override { return false; }
 
+        protected:
             void *memory() override {
                 return &storage;
             }
+
+        private:
+            typedef typename std::aligned_storage<sizeof(T), alignof(T)>::type Storage;
+            Storage storage;
         };
 
         template<>
@@ -254,6 +272,7 @@ namespace Async {
 
             bool isVoid() const override { return true; }
 
+        protected:
             void *memory() override {
                 return nullptr;
             }
--- a/src/common/net.cc
+++ b/src/common/net.cc
@@ -238,9 +238,7 @@ Address::init(const std::string& addr) {
         family_ = AF_INET6;
         try {
             in6_addr addr6;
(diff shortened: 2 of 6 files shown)


include/pistache/async.h
```diff
@@ -241,9 +254,14 @@ namespace Async {
 
             bool isVoid() const override { return false; }
 
+        protected:
             void *memory() override {
                 return &storage;
             }
+
+        private:
+            typedef typename std::aligned_storage<sizeof(T), alignof(T)>::type Storage;
```
knowledge4igor: Maybe it would be better to write:
```
using Storage = typename std::aligned_storage<sizeof(T), alignof(T)>::type;
```

src/common/net.cc
```diff
@@ -238,9 +238,7 @@ Address::init(const std::string& addr) {
         family_ = AF_INET6;
         try {
             in6_addr addr6;
-            char buff6[INET6_ADDRSTRLEN + 1] = {0, };
-            std::copy(host_.begin(), host_.end(), buff6);
-            inet_pton(AF_INET6, buff6, &(addr6.s6_addr16));
+            inet_pton(AF_INET6, host_.c_str(), &(addr6.s6_addr16));
```
knowledge4igor: Please, merge net.cc with master

src/common/net.cc
```diff
@@ -257,9 +255,7 @@ Address::init(const std::string& addr) {
         }
         try {
             in_addr addr;
-            char buff[INET_ADDRSTRLEN + 1] = {0, };
-            std::copy(host_.begin(), host_.end(), buff);
-            inet_pton(AF_INET, buff, &(addr));
+            inet_pton(AF_INET, host_.c_str(), &(addr));
```
knowledge4igor: Please, merge net.cc with master

include/pistache/async.h
```diff
@@ -171,13 +171,15 @@ namespace Async {
...
-                : state(_state)
+                : allocated(false)
+                , state(_state)
                 , exc()
                 , mtx()
                 , requests()
                 , id(_id)
             { }
 
+            bool allocated;
```
knowledge4igor: Can this variable be accessed from different threads?

include/pistache/async.h
AdrianCX: I just moved the code around:
 - I want it 'private' 
- This is to make sure nobody does any funny business (that would result in dangling pointers)

If you want something prettier that can be done as a separate commit.

src/common/net.cc
AdrianCX: merged - we now do:
inet_pton(AF_INET6, host_.c_str(), &(addr6.s6_addr16));

i.e: just pass the 'host_' to inet_pton. 
I don't know why a copy was done before that (just wasted cycles + valgrind issue)

src/common/net.cc
AdrianCX: merged - we now do:
inet_pton(AF_INET6, host_.c_str(), &(addr6.s6_addr16));

i.e: just pass the 'host_' to inet_pton.
I don't know why a copy was done before that (just wasted cycles + valgrind issue)

include/pistache/async.h
AdrianCX: If it can be accessed from different threads then that means that: 
- 'storage' can be accessed as well from different threads

That means that object in storage may be used before it is fully constructed...
So, there may be bigger issues if that is the case.

I can definitely change it to an 'atomic_bool' if you want but there are bigger issues if that's the case.

knowledge4igor: Ok, let it be so, we can check it using thread sanitizer further

AdrianCX: ok, thanks, I can help debug any further issues anyway.

PS: This whole file seems a bit over-engineered, simplicity is better when doing multi-threaded.

PR #433, opened 2019-02-04
Refactoring in Address handling and validation

knowledge4igor:


include/pistache/net.h
```diff
@@ -65,6 +64,7 @@ class AddrInfo {
 class Port {
 public:
     Port(uint16_t port = 0);
```
dennisjenkins75: All single-argument constructors should be explicit.  Please fix this one also.

knowledge4igor: @dennisjenkins75 Yes, you are right, but in some cases it's ok.
Adding `explicit` to this contructor doesn't permit us to write the following code:
```
port_ = HTTP_STANDARD_PORT;
port = std::stol(argv[1]);
```
What do you think about it?

dennisjenkins75: hmm... idk.  I typically add it b/c cppcheck will complain if I don't. :)

knowledge4igor:
@hydratim Thank you so much for this feedback

PR #435, opened 2019-02-05
Fix buffer memory leak

arthurafarias:
https://github.com/oktal/pistache/issues/399#issuecomment-459224155

### DIFF
--- a/include/pistache/stream.h
+++ b/include/pistache/stream.h
@@ -121,49 +121,12 @@ private:
 template<typename CharT>
 size_t ArrayStreamBuf<CharT>::maxSize = Const::DefaultMaxPayload;
 
-// struct Buffer {
-//     Buffer()
-//         : data(nullptr)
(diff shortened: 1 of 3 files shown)


arthurafarias:
Oh, this version of GCC doesn't supports incomplete type in template parameter. It builds in my machine.

image

I will try to avoid this type of declaration std::shared_ptr<char[]>.

arthurafarias:
I've probed it just works with g++ 7. Closing this merge request.

arthurafarias:
I think I fixed this problem. I will commit and see if works.

arthurafarias:
> At some point I would like to have the travis-ci test run against all "supported" platforms. I have a different open bug to remind me to work on this. Sadly, I'm not a travis-ci guru, but it looks simple enough; I just have not devoted time to it yet.

I could try to solve it. What is the issue?

PR #440, opened 2019-02-08
Coverage Report and Continuous Integration in Multiple Platforms

arthurafarias:
I've added coverage test and continuous integration new features. Compiling to diverse targets with multiple compilers. Someone should open an account in codecov and put the correct token to keep track of the correct repository because right now, I'm using my fork at arthurafarias/pistache.


arthurafarias:
I should make more improvements before trying to merge.

PR #442, opened 2019-02-09
c++ 11 override, whitepace, RFC 8297

lillypad:
As per the C++ override specifier in version 11:
https://en.cppreference.com/w/cpp/language/override

I added this to the examples and removed a little whitespace.

Not much but should help with the pure C++ 11 specification.

I've added the http status code `103` for HTTP Hints as per:
https://tools.ietf.org/html/rfc8297

And really nice library! must say this is fantastic work!

Cheers! 😄


Makefile
```diff
@@ -0,0 +1,19 @@
+all:
+	mkdir -p build/
+	cd build/ && \
+		cmake -G "Unix Makefiles" -DCMAKE_BUILD_TYPE=Release ../ && \
+		make
+
+debug:
+	mkdir -p build/
+	cd build/ && \
+		cmake -G "UnixMakefiles" -DCMAKE_BUILD_TYPE=Debug ../ && \
```
dennisjenkins75: You have a typo on this line.  "UnixMakefiles" != "Unix Makefiles"

Makefile
```diff
@@ -0,0 +1,19 @@
+all:
+	mkdir -p build/
+	cd build/ && \
+		cmake -G "Unix Makefiles" -DCMAKE_BUILD_TYPE=Release ../ && \
```
dennisjenkins75: Building pistache is more nuanced than this.  Some users will want to enable SSL, or compile the examples or the unit tests even.  I field ~1 question per week that could be answered by the user on their own if they had examined the examples.

lillypad:
@dennisjenkins75 Should be good now, sorry about the `Makefile` :smile:

PR #445, opened 2019-02-11
Add coverage test and integration with codecov.io

arthurafarias:


arthurafarias:
Interesting, here something strange is happening here:

> Processing tests/CMakeFiles/run_rest_server_test.dir/rest_server_test.cc.gcda
Out of memory!

this week I will investigate what is happening and I gonna put here a pull request.

arthurafarias:
Cool I think It is working. Not in all cases. because of different compiler's versions. But when it matches gcov version with the compiler's one It generates the reports. Please, it is possible to trigger a Travis Build in master branch to make codecov.io able to do comparisons between PRs and current code?

PR #450, opened 2019-02-15
Fixing case where const was cast away

iroddis:
Small fix to allow compilation with gcc's -Wcast-qual warning, and treating all warnings as errors.


iroddis:
Compiles under g++ 8.2.0 on Ubuntu, passes all tests.

Issue #457, opened 2019-02-19
Compilation fails if cppcheck not installed (Ubuntu)


gnyiri:
hi,
compilation fails without cppcheck installed. I could only get it work with following update in CMakeLists.txt:

if(CMAKE_CXX_CPPCHECK)
..
else()
  set(CMAKE_CXX_CPPCHECK "")
endif()

dennisjenkins75:
gnyiri@, thank you for the bug report.  It was not our intention to require cppcheck to compile pistache.

PR #458, opened 2019-02-19
Fixes Issue #457

arthurafarias:


arthurafarias:
@dennisjenkins75 Could you trigger the failing test again?

hydratim:
The test is only triggered by a new commit.
Your patch looks acceptable so it's probably a flake. 
Just add a comment or a newline and re-submit.

Issue #333, opened 2018-10-22
Simple "Hello World" example just hangs


adrianmercieca:
Hi,

I tried the "Hello World" program but unfortunately, the program just hangs when stress tested with Apache's ab tool.

The code is:
```C++
#include <pistache/endpoint.h>
#include <pistache/net.h>
#include <thread>

using namespace Pistache;

class HelloHandler : public Http::Handler {
 public:
  HTTP_PROTOTYPE(HelloHandler)

  void onRequest(const Http::Request& request, Http::ResponseWriter response) {
    response.send(Http::Code::Ok, "Hello, World");
  }
};

int main() {
  Address addr(Ipv4::any(), Port(9080));

  auto opts =
      Http::Endpoint::options().threads(std::thread::hardware_concurrency());
  Http::Endpoint server(addr);
  server.init(opts);
  server.setHandler(std::make_shared<HelloHandler>());
  server.serve();
}
```
![image](https://user-images.githubusercontent.com/37391885/47281713-a9050500-d5dc-11e8-83e7-8878cd5e46f1.png)

strace shows that the process is at epoll_wait; whilst the client has just received a 200 OK and trying to make another request.

Has anyone encountered this? Any clues please?

Thanks.

dennisjenkins75:
I attempted to reproduce the problem using your source with 2 small modifications (log to stderr in the handler and main should return 0).

"ab" locks up after sending one request, but the pistache program continues to respond when I access via "wget", "curl" and a regular web browser ("chrome" via apache reverse proxy to handle SSL termination).

```
djenkins@ostara ~/tmp $ g++ -ggdb3 -Wall pistache-bug-333.cc -lpistache -lpthread
djenkins@ostara ~/tmp $ gdb ./a.out 
GNU gdb (Gentoo 8.1 p1) 8.1
Reading symbols from ./a.out...done.
(gdb) run
Starting program: /home/djenkins/tmp/a.out 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".
[New Thread 0x7ffff6e87700 (LWP 7866)]
[New Thread 0x7ffff6686700 (LWP 7867)]
[New Thread 0x7ffff5e85700 (LWP 7868)]
[New Thread 0x7ffff5684700 (LWP 7869)]
request
request
request
request
request
request
request
request
request
request
request
request
request
request
```

The first "request" is from "ab".  The others from manual testing while "a.out" was still running.

```
[...]

PR #462, opened 2019-02-22
Attempt to fix HTTP connection header handling.

arthurafarias:
When client requests

keep-alive: the answer from the server is keep-alive and the connection remains open
close: the answer from the server is close and connection is closed
ext: the answer from the server is ext and connection is closed (I don't really know what exactly should be the behaviour of the server when connection is like this; so this follows default: close connection).

When no Connection header is specified the server closed the connection by default.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,8 @@ build
 *.exe
 *.out
 *.app
+
+# Ignoring IDE files
+
+.vscode
+.idea
--- a/include/pistache/http.h
+++ b/include/pistache/http.h
@@ -85,10 +85,13 @@ protected:
     Version version_;
     Code code_;
 
-    Header::Collection headers_;
+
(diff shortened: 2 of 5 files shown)


arthurafarias:
That's interesting. I've checked here and was passing. I gonna investigate latter. Closing this.

arthurafarias:
I submitted an attempt to fix issue #333. But I really appreciate if someone could make a correctness analysis. It's fixed, but I don't know if this is the appropriate solution.

arthurafarias:
@dennisjenkins75 could you trigger a new test for those who failed?

arthurafarias:
Apache Benchmark run without keep-alive connection.

image

Apache Benchmark run with keep-alive connection.

image

arthurafarias:
I think this closes #333.

Issue #16, opened 2016-10-21
rest_server.cc  example issue


MilleniumBPO:
no compile errors in rest_server.cc  example  but the example crashes on run and return munmap_chunk(): invalid pointer: 0x00007ffdf9f23ff0 *** error

oktal:
Hello,

Could you please provide more details on the issue ? Which operating system are you running on ? Which compiler are you using ? What exactly are you doing to make it crash ? Is it crashing right after being started ?

Also, could you please send the whole stack trace ?

Thanks.

MilleniumBPO:
Oktal thanks for your quick response, I am very interested in a rest pistache server for their compatibility with swagger. I hope that soon the documentation about create swagger definition for rest service is updated. 
The  above error occurs at runtime both the example and rest_description.cc rest_server.cc.  I use Ubuntu 14.04 LTS , Eclipse CDT (Luna and Mars2.0) with gcc 5.4.1 (2016 09 04) , I simply add the include directories to my c++ project: /usr/local/include/pistache, /mnt/san/rapidjson_master/include and /mnt/san/pistache/include , link the library path : /usr/local/lib and libs: net and net_static  , dialect ISO C++11 , copy and paste the example code without modifications.  The hello_server, http_server and http_client examples run well. The error(*** Error in `/mnt/san/workspacecpp/pistache_test/Debug/pistache_test': munmap_chunk(): invalid pointer: 0x00007fffffffde90 ***) occurs at    stats.init(thr) line , I also use Valgrind tool (valgrind --leak-check=full -v ./pistache_test)  but the result is confusing when describing the cause of memory leak:  thanks for your help.

The output of Valgrind tool:


--441-- REDIR: 0x569f9b0 (libc.so.6:strchrnul) redirected to 0x4c31a60 (strchrnul)
Now listening on http://0.0.0.0:9080
--441-- REDIR: 0x568f130 (libc.so.6:calloc) redirected to 0x4c2c931 (calloc)
==441== Thread 7:
==441== Conditional jump or move depends on uninitialised value(s)
[...]

PR #463, opened 2019-02-23
Use STL algorithm instead of for loop in FdSet CNTR

knowledge4igor:


knowledge4igor:
@dennisjenkins75 Flaky test again:
```
16: terminate called after throwing an instance of 'Pistache::Async::Error'
16:   what():  Resolve must not be called more than once
16/23 Test #16: payload_test .....................***Exception: Other  0.52 sec
``

knowledge4igor:
@dennisjenkins75 Maybe next week I will find free time to dive into this problem.

PR #470, opened 2019-02-28
Fix data races in Async whenAll and flaky tests

knowledge4igor:
In this PR I'm fixing data races in the case when we use `Async::whenAll` and as a result `data->resolved` can be increased simultaniosly in different threads and we have two `data->resolve` calls. I decided to use mutex to protect shared data and cannot understand how to use atomics in this case. Maybe someone proposes better solution. I think this PR will fix for the following problem in tests that happens from time to time:
```
13: terminate called after throwing an instance of 'Pistache::Async::Error'
13:   what():  Resolve must not be called more than once
```


include/pistache/async.h
```diff
@@ -1446,22 +1450,25 @@ namespace Async {
...
                     if (data->rejected) return;
 
                     data->results[index] = val;
-                    data->resolved.fetch_add(1);
-
-                    if (data->resolved == data->total) {
+                    data->resolved++;
+                    if (data->resolved == data->total)
+                    {
                         data->resolve(data->results);
```
dennisjenkins75: Is it appropriate to call data->resolve() while holding this mutex?

tests/http_client_test.cc
```diff
@@ -86,12 +88,15 @@ TEST(http_client_test, one_client_with_multiple_requests) {
     int response_counter = 0;
 
     auto rb = client.get(server_address);
-    for (int i = 0; i < RESPONSE_SIZE; ++i) {
-        auto response = rb.header<Http::Header::Connection>(Http::ConnectionControl::KeepAlive).send();
-        response.then([&](Http::Response rsp) {
-            if (rsp.code() == Http::Code::Ok)
-                ++response_counter;
-        }, Async::IgnoreException);
```
dennisjenkins75: Please try to avoid reformatting this code when making logic changes, it adds to the cognitive burden of trying to decipher what has changed and the broader implications of the change.

include/pistache/async.h
knowledge4igor: @dennisjenkins75 To be honest, I don't see any problems (for instance, dead locks).
In the `data->resolve(data->results);` line we hold mutex from `Data` object and just call `operator()(Arg&& arg)` from `Resolver` object. Do you see any problems? Maybe I don't understand something.

tests/http_client_test.cc
knowledge4igor: @dennisjenkins75 I see, I'm sorry, should I revert change in this commit related to this comment?

include/pistache/async.h
dennisjenkins75: I did not look at what data->resolve() did (sorry, was busy).  Was just asking a general question. I should have looked first.  Its ok.

tests/http_client_test.cc
dennisjenkins75: You do not need to change this PR.  But please avoid reformatting code not directly related to the logic changes that you are making.

Does your editor do this automatically?

It would be a problem if different contributor's editors constantly fought over style.  Arthur asked me about adopting clang-tidy as an enforced formatting style.  I don't want to do that just yet; forceably reformatting all of the code will break the "blame" for each line, and break everyone's inflight PRs.  The community as a whole should decide to do this.  Discussion for another day.

We really need a useable web forum and not a disjoint IRC and Slack channels.

knowledge4igor: @knowledge4igor I caught your point. I hope one day I will install IRC or Slack and join to discussion, maybe at the weekend. Is it a good idea to discuss formatting style in issues here (as we were discussing C++14 and C++17) ?

knowledge4igor: @dennisjenkins75 I caught your point. I hope one day I will install IRC or Slack and join to discussion, maybe at the weekend. Is it a good idea to discuss formatting style in issues here (as we were discussing C++14 and C++17) ?

knowledge4igor:
@dennisjenkins75 Sorry for my English, "here" - I meant github issues in this project.

PR #476, opened 2019-03-02
Delete obsolete code in http files

knowledge4igor:
It seems that this code is used and obsolete.


knowledge4igor:
@dennisjenkins75 Ok, I see, you are right, this method can be used for debugging purposes

PR #479, opened 2019-03-05
Fix data races in Async: whenAll and whenAny with input Args&&

knowledge4igor:


knowledge4igor:
@dennisjenkins75 Hello, please, can you re-run tests, it seems that there is yer another flaky test:
```
13: [ RUN      ] http_client_test.multiple_clients_with_one_request
13: terminate called after throwing an instance of 'std::runtime_error'
13:   what():  Connection error
13: Server address: localhost:38311
```

knowledge4igor:
@dennisjenkins75 I totally agree with you, I think should run on CI building Pistache with different sanitizers.

PR #481, opened 2019-03-08
Fix memory defects in http server tests

knowledge4igor:
In this PR I'm fixing memory defects in http server tests. At the current time we should explicitly set timeout to avoid memory defects on client. I will continue investigation of possible memory defects in client.


arthurafarias:
Why not put a default timeout in every request method instead of just the test case. This leak will occur when you repeat the same pattern try to get without being explicit and put a timeout in your request.

PR #485, opened 2019-03-08
Fix flaky tests caused by timeout problem

knowledge4igor:


knowledge4igor:
@dennisjenkins75 I don't know what's the problem (maybe yet another flaky tests, haha), cannot reproduct it locally:
```
16: [ RUN      ] payload.manual_construction
16: terminate called after throwing an instance of 'std::system_error'
16:   what():  Invalid argument
16/23 Test #16: payload_test .....................***Exception: Other  0.49 sec
```

PR #489, opened 2019-03-10
Improving coverage for src/common/http_header.cc

arthurafarias:


src/common/http_header.cc
```diff
@@ -204,7 +203,6 @@ CacheControl::write(std::ostream& os) const {
             case CacheDirective::SMaxAge:
                 return "s-maxage";
             case CacheDirective::Ext:
-                return "";
```
dennisjenkins75: Don't rely on fall-through here.
I am also surprise that there is a default, since directive.directive() is an enum (or I thought that it was?).

tests/headers_test.cc
```diff
@@ -10,7 +10,7 @@
 using namespace Pistache::Http;
 
 TEST(headers_test, accept) {
-    Header::Accept a1;
+    Pistache::Http::Header::Accept a1;
```
dennisjenkins75: I'm curious: why is the fully qualified symbol needed here?  line 10 has 'using namespace'.  Is there a symbol clash with 'Header' meaning multiple things?

arthurafarias: I am using Visual Studio Code and Clion to edit this project. The code completion feature wasn't resolving symbols correctly in both editors. I just added fully qualified symbols to avoid ambiguity when resolving the symbols.

src/common/http_header.cc
```diff
@@ -32,26 +48,9 @@ const char* encodingString(Encoding encoding) {
     case Encoding::Chunked:
         return "chunked";
     case Encoding::Unknown:
+    default:
```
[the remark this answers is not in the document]
arthurafarias: I see. I gonna remove this in a new commit.

src/common/http_header.cc
arthurafarias: I am surprised that there is a lambda function instead of a map. =P

arthurafarias: But this is not the point. The effort wasn't in refactoring the code just improving coverage of the test case.

tests/headers_test.cc
arthurafarias: I prefer fully qualified symbols to avoid ambiguities between:

- namespace Pistache::Http::Header
- class Pistache::Http::Header::Header

dennisjenkins75: Please do not make syntactically unneeded changes b/c an editor is broken.  I wonder if there is something weird in the header files in pisatche that is tripping up your editors?  In any event, please back these changes out.

This would be a good thing to note in any style guide and enforced with a linter and/or clang-tidy thing: use of "using namespace" in a .cc file.  But changing code to conform to any style guide should be done in its own PR, and after the code is stable and we cut a 0.1 release for the c++11 folks (so they can fork from that as we move into c++17-only land).

I do not want to create an undo hardship on editing code in pistache; but I also do not want the style code the code to shift back and forth as different people edit it with no rules / style guides.  I'm always open to discussion.

If you really want to keep these changes, then please remove the "using namespace" from the top of the .cc file; having both the "using namespace" and the fully-qualified symbols in that namespace in the same file can cause confusion in future (human) code readers / editors.

arthurafarias: Do you see that if you use

There is a symbol

Pistache::Http::Header::Accept
Pistache::Http::Header::Header

When 

using namespace Pistache::Http::Header;

and after

Header::Accept

Where the linter gonna look?

in the class Header::Accept

or in the namespace Header::Accept?

That was the problem so by simplicity I opted to use fully qualified names. But I know that if I use:

using namespace Pistache::Http

Members of the namespace are accessible just by Header:: prefix. So what do you prefer? Keep the fully qualified namespace or change it to the abbreviated form?

dennisjenkins75: Ah, I see.   There are namespaces and classes with conflicting names.  Yucky.

Remove the "using namespace" and then adjust the code as needed.  Need to keep this in mind for possibly renaming some of these things at a later date.  non-class namespaces and real classes should not clash like this.

arthurafarias: I've removed.

src/common/http_header.cc
arthurafarias: Reverted unnecessary modifications to http_header.cc (just to coverage point of view).

Issue #493, opened 2019-03-11
Plans to release Pistache 0.1


knowledge4igor:
Hello,
I think it's a good time to talk about planes to Pistache 0.1 release.

1) Do we have now serious bugs (blockers) that stop 0.1 release? If they are, then can we write down them?

2) I have run unit tests several time and now I don't see any flaky tests. Maybe you have different results.

3) Also I have run memcheck several times:
```
$ cmake --build . --target test_memcheck
```
several times
and have the following results:
```
...
100% tests passed, 0 tests failed out of 23

Total Test time (real) = 355.91 sec
-- Processing memory checking output:
MemCheck log files can be found here: ( * corresponds to test number)
/home/ws184/Devel/Opensource/pistache/build/Testing/Temporary/MemoryChecker.*.log
Memory checking results:
Built target test_memcheck
```
Don't see any defects. Maybe on your hardware and Linux OS you have different results, let's discuss it too.

Now I propose to stop (until release) to accept PR (I mean merge them to master) with **new features**: only bug fixes, code improvements and unit tests for features that have already been implemented.

What do you think about it? Any ideas?

arthurafarias:
Now we can add memory check as a required CI component.

dennisjenkins75:
Did we fix the denial of service bug where someone can hold a socket open and not flush the full HTTP header and essentially consume a pistache thread?

Does the pistache "echo" test crash when under sustained load from apache "ab"?

We should enable memory leak checking on the presubmit tests.  And possibly "tsan".

Its not a bug in pistache, but the travis-ci framework is very frustrating.  The tests often fail due to the test system itself not coming up (failure to fetch packages).

I need to check if github makes it easy to "tag" bugs or group them.  sadly, I am very busy right now and don't have a lot of time to do much other than rubber-stamp PRs. :(

dennisjenkins75:
We might also want to tackle issue #471, agree upon a code-formatting style (I'm flexible, so long as the indenting tool can be ran from the Linux CLI and the config is checked into github), then reformat the code (after final bugs are fixed, before cutting the release).

knowledge4igor:
@dennisjenkins75 

> Did we fix the denial of service bug where someone can hold a socket open and not flush the full HTTP header and essentially consume a pistache thread?

and

> Does the pistache "echo" test crash when under sustained load from apache "ab"?

Sounds like serious bugs. Can you tell me link to issue or describe steps to reproduce these problems? I will try to find time to dive into them.

[...]

Issue #471, opened 2019-03-01
RFP: Implement a standard C++ formatting style, use tooling to enforce it.


dennisjenkins75:
We all have different code formatting styles (potentially).  I've seen some PRs that adjust style while also making logic changes (please don't do this).  Worse is when a style oscillated back and forth.  Arthur suggested that the project adopt "clang-tidy" style and have a presubmit check that forcibly reformats all code.  Personally, I do not want to do that just yet; the churn would be heavy and break everyone's open PRs and muck up the "blame" history.

I was thinking that once a style is agreed upon by the major contributors, we would reformat the code one or two cfiles at a time (like a related .h and .cc file).  Ideally, we would do this during a certain time frame so that everyone has a chance to NACK b/c they are actively working on a PR for a targeted file.

Thoughts?  Discussions?  Bikeshedding (mine is green)?  Flame wars over style guides?

dennisjenkins75:
For consideration, PR #661, using "clang-format" with default options to reformat one .cc file.

dennisjenkins75:
There are lots of formatting options available [1].  I'm not a huge fan of the clang-format llvm defaults.  I might tinker with a custom style file.

[1] https://clang.llvm.org/docs/ClangFormatStyleOptions.html

dennisjenkins75:
I have successfully reformatted every file except "include/pistache/async.h" via "clang-format" with default options.  clang-format will happily reformat async.h, but cppcheck-1.88 then segfaults when trying to evaluate it.  cppcheck-1.88 is the current version on my Gentoo dev box.  However, I installed cppcheck-1.90 from github HEAD, and that works fine.

IDK what version of cppcheck is used by travis, but I'll try a pull request and see if a reformatted async.h crashes in travis or not.

PR #495, opened 2019-03-12
[SUGGESTION] Use overcommit to manage git hooks and add git pre-hook to format the code.

arthurafarias:
In order to solve issues #493 and #471, I've added overcommit to manage git hooks and added a formatting pre-commit hook.

Overcommit is a tool to manage git hooks: https://github.com/brigade/overcommit
It depends on ruby and it is easy to install.

I've added a pre-commit script that checks if clang-format exists: if it doesn't exist, the tool will not let the contributor commit the changes.

If clang-format exists it will format the code and add the formatted code to current commit.

What do you think about?


arthurafarias:
Formatting changes where not added to commit.

.githooks/pre-commit/format.sh
```diff
@@ -0,0 +1,20 @@
...
+if [ ! -x $CLANG_FORMAT ]; then
+    echo "Failed to format clang-format wasn't found."
+    exit -1
+fi
+
+#finding root directory
+ROOT_DIRECTORY=${ROOT_DIRECTORY:-$(realpath $(pwd))}
+
+# formatting all code
+clang-format -i -style=Mozilla $(find $ROOT_DIRECTORY -type d -name third-party* -prune -o -print | grep -E '.*\.(h|hpp|cc|cpp|cxx)$')
```
dennisjenkins75: https://developer.mozilla.org/en-US/docs/Mozilla/Developer_guide/Coding_Style

Style looks ok, but will clang-format force the renaming of existing tokens that violate this style guide?

IMHO, we should keep this PR at the ready, but methodically reformat the code, a few files at a time (like a mated pair of .h/.cc files).  Once everything is close to the official format, then we commit this PR and make it official/formal/enforced.

It would be nice if there were a CMake create makefile target that reformats code for us.  "make clang-tidy" or something, that we can run on our local workspaces after editing code and before creating a pull request (so we can see the canonical diff first).  Maybe git has a comment to re presubmit hooks but not create a PR that I am unaware of?

One last thing: I don't want to stomp on any pending PRs that Kip, Tim or Igor (or anyone else) is working on.  But I have the contact info for all 4 of you, so I can email you all to get signoff before submitting this PR.

I am excited for this.  The code is a mess and a uniform style will improve readability.  I just want to proceed with caution.

arthurafarias: 1 - Now it is using overcommit to check, not to format.
2 - I created a new target format that can be called as "make format".
3 - I modified the name of test_memcheck to memcheck
4 - I created a new cmake option called PISTACHE_CREATE_OSS_CONTRIB_TARGETS
 - Turns Valgrind, cppcheck and clang-format as required dependencies and configure them
 - Creates format target
 - Create test and docs targets
 - this option is enabled by default when using debug target

What do you think about these modifications?

arthurafarias:
Could you trigger a new check on travis build #784.7?

PR #499, opened 2019-03-16
Fix warnings

knowledge4igor:


knowledge4igor:
@dennisjenkins75 Some problems in payload test, I will try to fix it asap.

src/common/http_header.cc
```diff
@@ -50,7 +50,10 @@ const char* encodingString(Encoding encoding) {
         return "chunked";
     case Encoding::Unknown:
     	return "unknown";
+    default:
```
dennisjenkins75: Please do not add a "default" for switch statements that switch on enums.  Add the actual missing values.  The compiler warning about missing values from the switch statement is useful, and adding a default just masks the check.

Issue #520, opened 2019-04-07
Client not appending query to path string when writing request


matt1795:
Hey guys, I'm setting up testing for my REST API, and I noticed that the client wasn't writing the query to the request. I know the client doesn't get much love, so I'll send up the pull request soon.

Oh and thanks everyone for working on this library, I absolutely adore it.

PR #521, opened 2019-04-07
added query to path string

matt1795:
Simple fix for #520.


matt1795:
@dennisjenkins75 how's this?

Issue #343, opened 2018-11-03
Using C++14 and C++17 features


knowledge4igor:
I have noticed several changes in CMakeLists.txt related to C++14 and C++17. If I'm not mistaken we can write C++ code in Pistache based at least on C++14 standard (even more simply, we can use features from C++14). Am I right?

dennisjenkins75:
As far as I know, the answer you your question is "no".  Pistache should still compile on an older Debian distro that only supports c++11.  Another contributor added a "string_view" that compiles on c++11 and implements the c++14/17 spec.

Unfortunately, I don't recall who is using the older compiler.  We should probably announce a deprecation date for pure c++11, and allow contributions that only support c++17 and beyond.

I've been holding off on making an official release b/c we still have lots of bugs and interface-level (eg, things exposed via header files) code changes.  But before we drop c++11 support, we should probably have an official release that does support c++11 so that if someone really needs it, we can point them to it.

I don't want to be tied down by c++11 forever, but I don't have a solid plan w/ firm dates to move past it.  I'm open to suggestions!

janhkrueger:
I think it was me with the issue of an older cmake on debian stable.
For us it would be ok of you move forward, we will adjust.

knowledge4igor:
@dennisjenkins75 Hello, do we have some updates or news about moving to C++14? I have started to fix yet another memory leak in Client related to `getaddrinfo` (see your PR #401). For my solution I need generalized lambda capture feature from C++14. If there are no plans I will think about another solution.

dennisjenkins75:
Sadly, I do not know which compilers are in which older distros and which levels of c++ they support.  jahnkrueger@ mentioned Debian Stable (I assume this is a moving target)?  How easy is it to get g++ 7.3.0 (supports c++17) installed onto older Ubuntu, Debian and Redhat systems?

We should add the oldest distro that users want to our travis.yaml file.  But we have no easy way to sample that.  There is a nearly idle slack channel for pistache, and an IRC channel with only 4 ppl on it.

knowledge4igor@ Proceed with fixing the bugs.  Keep the getaddrinfo fix separate from the others, in case it needs to get rolled back.

janhkrueger:
Debian Stable currently ships with a 6.3:
```
⌁ insulae  ~   g++ --version
g++ (Debian 6.3.0-18+deb9u1) 6.3.0 20170516
```

[...]

PR #529, opened 2019-04-16
Remove std::regex in server code

### LINKED ISSUE #507: rest_server crashes with 'std::out_of_range'
run_rest_server craches with message
terminate called after throwing an instance of 'std::out_of_range'

The problem happens on the line 61 of rest_server.cc which calls:
Routes::Post(router, "/record/:name/:value?", Routes::bind(&StatsEndpoint::doRecordMetric, this));

Swagger-codegen version: taken from git on March 23

pistache and its example and tests compiled with
cmake (3.6.3)
make (G++ 4.8.2)

Launched with the command
./run_rest_server

reyoung1110:
std::regex is not implemented until GCC 4.9. By replacing it by
a handcrafted code can make pistache server compatible with GCC 4.8.

Fix #507

### DIFF
--- a/include/pistache/router.h
+++ b/include/pistache/router.h
@@ -6,7 +6,6 @@
 
 #pragma once
 
-#include <regex>
 #include <string>
 #include <tuple>
 #include <unordered_map>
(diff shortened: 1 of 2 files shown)


hydratim:
We decided to move to C++14 back in January and only support GCC >= 5 since then.
See #343 
Rolling back a change we made since that decision is not likely something that we are going to merge.

reyoung1110:
OK. Maybe we need an update in `README.md` later and point out which version of `GCC` is supported.

zhanb:
hello, Why not consider keeping a branch compatible with gcc4.8 ？

kiplingw:
Because as @hydratim pointed out, we moved to C++14. It is very complicated to maintain backwards compatibility. It makes a mess of the code base with various preprocessor directives, configure time checks, and so on.

Issue #501, opened 2019-03-18
bad file descriptor error


jacksu:
![QQ截图20190318213901](https://user-images.githubusercontent.com/6895190/54533918-40bd2280-49c6-11e9-9d8a-0b023f208fc0.png)

jacksu:
https://github.com/oktal/pistache/blob/master/src/common/os.cc#L187

knowledge4igor:
@jacksu Do you have this problem on latest source? What are the steps for reproducing this uncaught exception?

albertochm95:
I have the same problem. I updated pistache and the problem continues

dmorton-spirent:
My recent PR #536 added a unit test that triggers this error message. It made the build flaky and I spent a day digging further into it. Results of my investigation are included in this comment: 
https://github.com/oktal/pistache/pull/536#issuecomment-491460208.

tl;dr version: if the HTTP parser is called more than once and the first call contains text that causes a parser exception to be thrown (i.e. invalid header value) and the second call contains a HTTP header (even a valid one), then the above error is thrown when the second call tries to send back a second error to the client.

dmorton-spirent:
Further investigation has, I believe, uncovered the root cause. Continuing on from the thread above. When the first HTTP parser exception is thrown Pistache catches it and sends an error response to the client. The code that handles transmitting the error response defaults to closing the connection after transmitting the reply. Connection is kept open if and only if the request contained the `Connection: Keep-Alive` header. 
https://github.com/oktal/pistache/blob/4355267e20a912992006b2ca914147a8cb49c3fa/src/common/http.cc#L651-L655
and
https://github.com/oktal/pistache/blob/4355267e20a912992006b2ca914147a8cb49c3fa/src/common/http.cc#L689-L692
In the error condition from the unit test the `Connection: Keep-Alive` header is either not sent at all or sent after the invalid header.

When Pistache tries to send the second error reply the `send()` call returns `-1` on account of the file descriptor being closed. The current error checks don't check for `EBADF` in that case. Instead, control passes to the default error path which discards the reply and reconfigures the file descriptor. It's this reconfigure step that causes the crash.

https://github.com/oktal/pistache/blob/4355267e20a912992006b2ca914147a8cb49c3fa/src/common/transport.cc#L262-L275

PR #536, opened 2019-05-10
Add support for HTTP 405 responses from the REST server

dmorton-spirent:
If the REST router cannot find an exact match for the requested method
and resource, it will search for other methods that support the
resource. If another method supporting the resource is found, router
will send a HTTP 405 response to the client.

Also added a unit test to exercise REST server response code support.

### DIFF
--- a/include/pistache/http.h
+++ b/include/pistache/http.h
@@ -486,6 +486,12 @@ public:
      * - movedPermantly -> 301
      * - moved() -> 302
      */
+    Async::Promise<ssize_t> sendMethodNotAllowed(const std::vector<Http::Method>& supportedMethods) {
+        code_ = Http::Code::Method_Not_Allowed;
+        headers_.add(std::make_shared<Http::Header::Allow>(supportedMethods));
+        std::string body = codeString(Pistache::Http::Code::Method_Not_Allowed);
+        return putOnWire(body.c_str(), body.size());
+    }
 
     Async::Promise<ssize_t> send(Code code) {
         code_ = code;
--- a/include/pistache/router.h
+++ b/include/pistache/router.h
@@ -70,7 +70,7 @@ class Request;
 struct Route {
     enum class Result { Ok, Failure };
 
-    enum class Status { Match, NotFound };
+    enum class Status { Match, NotFound, NotAllowed };
 
     typedef std::function<Result(const Request, Http::ResponseWriter)> Handler;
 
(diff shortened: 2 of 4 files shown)


dmorton-spirent:
Hi. I was able to reproduce the issue on my Ubuntu 18.04 box with gcc-6. I can see what's happening, but am not sure of the best way to fix it.

For starters it's the "HTTP 415" test I added that's triggering the issue. Commenting that out removes the flakiness. Based on the following investigation I believe this case has uncovered an existing issue, and that the changes in this PR did not cause it.

From what I can see the issue is the result of Pistache trying to send multiple error responses to a single client request. The second response is, I think, trying to write to a socket that's already been closed and that's where you get the "Bad file descriptor" from.

The multiple responses are the result of a unique interaction between https://github.com/oktal/pistache/blob/4355267e20a912992006b2ca914147a8cb49c3fa/src/common/http.cc#L785 and https://github.com/oktal/pistache/blob/4355267e20a912992006b2ca914147a8cb49c3fa/src/common/transport.cc#L126
For the issue to happen, `Handler::onInput` is called twice and an `HttpError` is thrown in the first call. The `HttpError` is caught here and the parser object gets reset. https://github.com/oktal/pistache/blob/4355267e20a912992006b2ca914147a8cb49c3fa/src/common/http.cc#L818-L823 
The issue happens when the second `Handler::onInput` call is passed a `char * buffer` that begins with an HTTP header. When this happens the parser thinks it's seeing a new request and throws an exception when the headers it sees don't look valid. The exception eventually triggers the same catch block from above, and it dies when it tries to send the second error response.

For reference here's what the `char * buffer` variable looks like in the error case.
First call:
``POST /read/function1 HTTP/1.1\r\nAccept: */*\r\nConnection: close\r\nContent-Length: 38\r
\nContent-Type: invalid\r\nHost: localhost:49819\r\n``
Second call:
```User-Agent: cpp-httplib/0.2\r\n.........```

As far as why this is an intermittent issue, I did notice that the `recv()` call in `Transport::handleIncoming` will usually return the whole request in a single call. But sometimes it will return the request in multiple calls.

As far as fixes I'm wondering if it would be best to stop processing a request once an HTTP parser error is encountered?

dmorton-spirent:
It looks like there's an existing issue #501 that might be related (same error message). I've added the research from above as a comment there. And I'll see what I can do :).

PR #541, opened 2019-05-20
Add error handling in cases of writing to an invalid file descriptor

### LINKED ISSUE #501: bad file descriptor error

dmorton-spirent:
Related to recent unit test intermittent failures.

Handles a condition where the file descriptor for a connection is closed
before all processing is done, leading to an error. This is usually the
result of multiple exceptions being thrown by the HTTP parser without a
connection being set to `Keep-Alive`.

In the error case the HTTP parser code closes the file descriptor but there are multiple replies in flight internally. Previously when the `send()` call failed, Pistache would still try to modify the file descriptor which lead to an error from the `epoll` code. This change modifies the cleanup code when a bad file descriptor is detected during send to skip trying to modify the file descriptor.

Closes #501 (same error message) and makes the unit tests reliable again.

### DIFF
--- a/src/common/transport.cc
+++ b/src/common/transport.cc
@@ -269,6 +269,11 @@ Transport::asyncWriteImpl(Fd fd)
                     wq.push_front(WriteEntry(std::move(deferred), bufferHolder, flags));
                     reactor()->modifyFd(key(), fd, NotifyOn::Read | NotifyOn::Write, Polling::Mode::Edge);
                 }
+                else if (errno == EBADF) {
+                    wq.pop_front();
+                    toWrite.erase(fd);
+                    stop = true;
+                }
                 else {
                     cleanUp();
                     deferred.reject(Pistache::Error::system("Could not write data"));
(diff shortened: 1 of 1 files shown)


src/common/transport.cc
```diff
@@ -269,6 +269,11 @@ Transport::asyncWriteImpl(Fd fd)
                     wq.push_front(WriteEntry(std::move(deferred), bufferHolder, flags));
                     reactor()->modifyFd(key(), fd, NotifyOn::Read | NotifyOn::Write, Polling::Mode::Edge);
                 }
+                else if (errno == EBADF) {
```
dennisjenkins75: Please add a comment referencing the URL for the bug in github.  Code that ignored EBADF looks fishy (or it will in a year or so when someone else is spelunking here).  i will merge PR as soon as you've done this.  TYVM for the work-around.

dmorton-spirent: You are very welcome! Happy to be able to give something back :).

PR #542, opened 2019-05-20
Some clean-up of the net primitives

hydratim:
Currently untested, but should work and be backwards compatible-ish.
* Unifies Ipv4 and Ipv6 classes into IP class
* Swaps out std::string host_ for the new IP class as the backing store for Address class

### DIFF
--- a/include/pistache/net.h
+++ b/include/pistache/net.h
@@ -83,45 +83,35 @@ private:
     uint16_t port;
 };
 
-class Ipv4 {
-public:
-    Ipv4(uint8_t a, uint8_t b, uint8_t c, uint8_t d);
 
-    static Ipv4 any();
-    static Ipv4 loopback();
+class IP {
+private:
+    int port;
+    int family;    
+    union {
+        struct sockaddr_in addr;
+        struct sockaddr_in6 addr6;
+    };
+public:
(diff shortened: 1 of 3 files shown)


hydratim:
This is part 1 of my proposed clean-up.

Currently this fails the `Ipv6::any()` net test because I made it so that if you want an `Ipv6::any()` you have to instead supply `Ipv6::any(true)` - This is because the now (propsed) legacy `Ipv6` and `Ipv4` are actually aliases of `IP`

I'd rather adjust the test and document the change than make a child class (instead of the alias), but this is just my preference.

Any thoughts? @dennisjenkins75 @knowledge4igor

hydratim:
I wasn't expecting you to merge it until I'd fixed the tests. 
At this point I was just asking for feedback on the PR and explaining why I broke the test for now.

hydratim:
All tests passing.
It would be nice to see a benchmark pre-merge and post-merge as this should (in theory) be faster as we aren't converting to / from `std::string` every time we need an address in network format.

hydratim:
Thanks @dennisjenkins75 I've got some thoughts for some more clean-up to follow, but my next PR will probably be updates to documentation and some new examples.

Issue #221, opened 2018-05-24
InstallSignalHandler does not work as expected


1Hyena:
```C++
Pistache::Address addr(Pistache::Ipv4::any(), Pistache::Port(options.get_port()));
auto opts = Pistache::Http::Endpoint::options().threads(2).flags(Pistache::Tcp::Options::InstallSignalHandler);;

Pistache::Http::Endpoint server(addr);
server.init(opts);
server.setHandler(Pistache::Http::make_handler<HttpHandler>());

log("Serving the API on %s:%d.", addr.host().c_str(), static_cast<uint16_t>(addr.port()) );
server.serve();
server.shutdown();
```

Sending CTRL+C to the terminal does nothing although I would expect `server.serve()` to return and process to properly shut down. What purpose does the `InstallSignalHandler` serve? Also, isn't it a bad design choice that a multi-threaded lib attempts to handle signals? Signals in a multi-threaded process are already tricky enough, so having a lib to have its own signal handler only adds to the complexity with questionable benefits.

1Hyena:
Anyway, this is how I solved it:

First before any threads are created I have this code:
```C++
    if (sigemptyset(&signals)        != 0
    ||  sigaddset(&signals, SIGTERM) != 0
    ||  sigaddset(&signals, SIGINT)  != 0
    ||  sigaddset(&signals, SIGQUIT) != 0
    ||  sigaddset(&signals, SIGPIPE) != 0
    ||  sigaddset(&signals, SIGALRM) != 0
    ||  pthread_sigmask(SIG_BLOCK, &signals, nullptr) != 0) return false;
```

It should ensure that no thread will be able to get any signals. Since Pistache is running in threaded mode I can just keep polling for any signals from the main thread.

```C++
Pistache::Address addr(Pistache::Ipv4::any(), Pistache::Port(options.get_port()));
auto opts = Pistache::Http::Endpoint::options().threads(2);
Pistache::Http::Endpoint server(addr);
server.init(opts);
server.setHandler(Pistache::Http::make_handler<HttpHandler>());

log("Serving the API on %s:%d.", addr.host().c_str(), static_cast<uint16_t>(addr.port()) );
server.serveThreaded();

bool terminate = false;
while (!terminate) {
    int number = 0;
    int status = sigwait(&signals, &number);
    if (status != 0) {
        log("sigwait() returned %d, shutting down.", status);
        break;
    }

    switch (number) {
[...]

PR #547, opened 2019-06-08
remove broken Tcp::Options::InstallSignalHandler

aarchetype:
(see discussion in #221)
Also remove unreachable shutdown() from examples and add
http_server_shutdown example with signal handling (thanks to 1Hyena).

### DIFF
--- a/examples/http_server_shutdown.cc
+++ b/examples/http_server_shutdown.cc
@@ -19,6 +19,7 @@ int main() {
     if (sigemptyset(&signals) != 0
             || sigaddset(&signals, SIGTERM) != 0
             || sigaddset(&signals, SIGINT) != 0
+            || sigaddset(&signals, SIGHUP) != 0
             || pthread_sigmask(SIG_BLOCK, &signals, nullptr) != 0) {
         perror("install signal handler failed");
         return 1;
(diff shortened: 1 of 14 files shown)


include/pistache/tcp.h
```diff
@@ -25,9 +25,7 @@ enum class Options : uint64_t {
     Linger               = NoDelay << 1,
     FastOpen             = Linger << 1,
     QuickAck             = FastOpen << 1,
-    ReuseAddr            = QuickAck << 1,
-    ReverseLookup        = ReuseAddr << 1,
```
dennisjenkins75: You removed ReverseLookup (unintentional)?

examples/http_server_shutdown.cc
```diff
@@ -0,0 +1,48 @@
...
+        UNUSED(request);
+        response.send(Pistache::Http::Code::Ok, "Hello World\n");
+    }
+};
+
+int main() {
+    sigset_t signals;
+    if (sigemptyset(&signals) != 0
+            || sigaddset(&signals, SIGTERM) != 0
+            || sigaddset(&signals, SIGINT) != 0
```
dennisjenkins75: I suggest also trapping SIGHUP for this example program.

include/pistache/tcp.h
aarchetype: Actually it was intentional because ReverseLookup isn't being used anywhere.

examples/http_server_shutdown.cc
aarchetype: Well, it's make a sense.

PR #550, opened 2019-06-12
Request and response limits

iroddis:
- Renamed maxPayload* to maxRequestSize
- Added maxResponseSize
  - Added Endpoint::setMaxResponseSize option
- Renamed payload_test to request_size_test


iroddis:
I'm not sure how to check that a send().then() throws an error when trying to return a response larger than maxResponseSize.

include/pistache/endpoint.h
```diff
@@ -21,13 +21,15 @@ class Endpoint {
         Options& threads(int val);
         Options& flags(Flags<Tcp::Options> flags);
         Options& backlog(int val);
-        Options& maxPayload(size_t val);
```
dennisjenkins75: For now, please keep the old method as well.  Have it just call the new method (inline is fine), but give the old method the "__deprecated__" attribute, as documented:

https://gcc.gnu.org/onlinedocs/gcc-7.3.0/gcc/Common-Function-Attributes.html#Common-Function-Attributes

src/server/endpoint.cc
```diff
@@ -38,8 +39,14 @@ Endpoint::Options::backlog(int val) {
 }
 
 Endpoint::Options&
-Endpoint::Options::maxPayload(size_t val) {
```
dennisjenkins75: Please retain the older method for now, mark as __deprecated__.

iroddis:
It looks like the minimum standard is C++14, so I went with the newer [[deprecated]] attribute.

The TravisCI failed due to networking issues? The error was that it could not install cppcheck, due to a host lookup issue.

Issue #438, opened 2019-02-07
How to access (basic auth) authorization header?


alexe100:
Hi
How can I get the basic auth authorization header?

Thanks a lot

dennisjenkins75:
Header management in Pistcahe is annoying.  One of the original design goals (Oktal's) was that all headers should be strongly typed.  This helps ensure that code using the headers is less error-prone, but makes it difficult to use headers that are not coded for.  There is an interface to use non-defined headers (raw headers), and to define custom headers.  But once a header is formally defined, it disappears from the "raw headers" list (src/common/http.cc, HeadersStep::apply()).

Anyway, I was able to access the Authorization header using the raw headers interface (details below).

If you were to author a pull request that implements the "authorization" header formally (don't forget to add unit tests), I would happily merge it.  Look in include/pistcahe/http_header.h for details.


```
$ curl -X POST -d "hello world\n"H "Authorization: Basic SDFHKDHFKDJFK"  \
   http://localhost:9080/v1/system/echo



Server ready: http://localhost:9080/
Authorization Basic SDFHKDHFKDJFK
2019-02-07 11:16:42 200 postEchoTest 380



void SystemService::postEchoTest(
  const Rest::Request& request,
  Http::ResponseWriter response
) {
  Http::Mime::MediaType mime;
  const auto mime_ptr = request.headers().tryGet<Http::Header::ContentType>();
  if (mime_ptr) {
    mime = mime_ptr->mime();
  }

  response.send(Http::Code::Ok, request.body(), mime);

  for (const auto &hdr: request.headers().rawList()) {
    std::cerr << hdr.first << " " << hdr.second.value() << "\n";
  }
}
```

alexe100:
Ok, done! The 'raw' does it.

(Now the question is answered here but I thing that would be very useful an authorization header (at least BASIC) sample or chunk of code in the web page :-)

Thanks a lot
Pistache seems great!!

dennisjenkins75:
It should be fairly easy to formally add an Authorization header class (with handy methods to parse the base64 payload), and document it via a few unite tests.  Please let this issue open to serve as a reminder and someone will implement it sooner or later.  :)  note that once Authorization is a formally supported header, that it will disappear from the "raw" array.  imho, "raw" should include all headers, even those with a real class + parser.  Maybe we'll make that change too.

kiplingw:
[...]

PR #551, opened 2019-06-12
All raw headers are now preserved

kiplingw:
Part of Oktal's original design for Pistache as @dennisjenkins75 has pointed out in #438 was to ensure that all headers are strongly typed to reduce errors. 

This is a good idea, but the drawback is that there is no simple way to enumerate / print all of the client request's headers unless you could always anticipate exactly the type and handle each type manually. This is not helpful for debugging situations where one needs to examine what a client is sending a server. Consider the situation where the server has the expected headers registered with the Registry, but the client sends the wrong headers (perhaps a typo), or perhaps you want to log all client headers.

Pistache::Rest::Request::headers() is a vector of all of the client's strongly typed headers if they were registered with the Registry. A "raw" copy is stored in the rawList() unordered_map only if the header was not registered in the Registry.

This patch ensures that all client request headers, regardless of whether they are registered or not, are still accessible via the rawList() unordered_map interface.


src/common/http.cc
```diff
@@ -297,15 +297,19 @@ namespace Private {
...
                 message->headers_.add(header);
             }
-            else {
-                std::string value(cursor.offset(start), cursor.diff(start));
-                message->headers_.addRaw(Header::Raw(std::move(name), std::move(value)));
-            }
+
+            // But also preserve a raw header version too, regardless of whether
+            //  its type was known to the Registry...
+            std::string value(cursor.offset(start), cursor.diff(start));
```
dennisjenkins75: Please add a unit tests that verifies that a registered header is also in the raw header's list.

I am surprised that we do not have a unit test that (would now fail) that verifies the inverse.  Oh well.

README.md
```diff
@@ -16,9 +16,24 @@ Pistache was originally created by Mathieu Stefani, but he is no longer actively
 
 For those that prefer IRC over Slack, the rag-tag crew of maintainers idle in `#pistache` on Freenode. Please come and join us!
 
-# Precompiled packages
+# Precompiled Packages
```
dennisjenkins75: Should your Debian edits be in a different PR?  What if the raw header logic needs to get rolled back?

kiplingw: Sorry, for some reason I thought I had already submitted a PR. I just did as #552.

src/common/http.cc
kiplingw: Done.

PR #560, opened 2019-06-15
Change version.h and libpistache.pc install source location to use CMAKE_CURRENT_BINARY_DIR

### LINKED ISSUE #559: Change version.h and libpistache.pc install source location to use CMAKE_CURRENT_BINARY_DIR instead of CMAKE_BINARY_DIR
I have developed a Conan recipe for Pistache and ran into a problem in a couple of places in the root `CMakeLists.txt`.  For example:
```
    # Install version header...
    if(PISTACHE_INSTALL)
        install (
            FILES
            ${CMAKE_BINARY_DIR}/include/pistache/version.h
            DESTINATION
            ${CMAKE_INSTALL_INCLUDEDIR}/pistache/
        )
    endif()
```
My Conan recipe `CMakeList.txt` includes the Pistache source directory like `include(pistache)`.  When I `make install` using my `CMakeLists.txt` it can't find `version.h` or `libpistache.pc`.   The fix is to change `CMAKE_BINARY_DIR` to `CMAKE_CURRENT_BINARY_DIR`, for example:
```
    # Install version header...
    if(PISTACHE_INSTALL)
        install (
            FILES
            ${CMAKE_CURRENT_BINARY_DIR}/include/pistache/version.h
            DESTINATION
            ${CMAKE_INSTALL_INCLUDEDIR}/pistache/
        )
    endif()
```
This is a more flexible `CMakeLists.txt` for Pistache because it allows it to be included by parent CMake projects.

sourcedelica:
Fixes #559

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -114,7 +114,7 @@ include_directories (${CMAKE_CURRENT_SOURCE_DIR}/include)
     if(PISTACHE_INSTALL)
         install (
             FILES
-            ${CMAKE_BINARY_DIR}/include/pistache/version.h
+            ${CMAKE_CURRENT_BINARY_DIR}/include/pistache/version.h
             DESTINATION
             ${CMAKE_INSTALL_INCLUDEDIR}/pistache/
         )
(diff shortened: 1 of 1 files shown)


sourcedelica:
Thanks!

Issue #337, opened 2018-10-24
crash when trying to chain promises


mnottale:
Hi.

The following snippet crashes with a stack overflow, invoking the exception callback in a loop:

```
TEST(Promise, Fail2)
{
    bool ok = false;
    std::unique_ptr<Pistache::Async::Rejection> rejecter;
    Pistache::Async::Promise<int> promise([&](Pistache::Async::Resolver& resolve, Pistache::Async::Rejection& reject) {
        	rejecter.reset(new Pistache::Async::Rejection(std::move(reject)));
    });
    promise.then(
    	[](int v) -> Pistache::Async::Promise<int> {
    		std::cerr << "SHOULD NOT BE THERE" << std::endl;
    		return Pistache::Async::Promise<int>::resolved(v);
    	},
    	[&](std::exception_ptr e)->Pistache::Async::Promise<int>
    	{ 
    		std::cerr << "REJECTION CALLBACK" << std::endl;
    		ok = true;
    		return Pistache::Async::Promise<int>::rejected(e);
    	});
    
    ASSERT_FALSE(ok);
    (*rejecter)(std::runtime_error("foo"));
    ASSERT_TRUE(ok);
}
```

dennisjenkins75:
Thank you for the bug report.  +2 points for the repro.

Unfortunately, I don't know when I'll have time to dig into this.  Maybe this coming weekend?  Others might jump in with a fix.

Have you tried running the unit test directly under valgrind?  This will be my first course of action if/when I attempt to fix this.

Would you consider creating a PR that adds this unit test (yes, it will fail now, but I would have something to work from for a fix).

mnottale:
Hi.
I managed to fix this issue, opened a PR there: https://github.com/oktal/pistache/pull/340

PR #561, opened 2019-06-16
Recursive rejects

iroddis:
This is trying to solve the same problem as #337 
- Added the test from that issue
- Changed the chain handling behaviour to immediately increment the reject counter, and to just return if an attempt is made to reject the promise already (as it's already been rejected).

I'm not 100% sure that this is the right approach, but it works for my workload (hundreds of short and very long (>4 minute) requests, with clients disconnecting mid-request. It also passes all the unit tests.

I just think this needs a code review from someone who understands the promise system more deeply than I do.

### DIFF
--- a/include/pistache/async.h
+++ b/include/pistache/async.h
@@ -288,16 +288,19 @@ namespace Async {
 
             void resolve(const std::shared_ptr<Core>& core) override {
                 if (resolveCount_ >= 1)
-                    throw Error("Resolve must not be called more than once");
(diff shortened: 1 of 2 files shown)


hydratim:
Sorry for the late response on this one, I couldn't see any obvious issues with it, but I'm not an active user of the promises system.

In short this looks like a reasonable fix.

I'll add the promises system to my todo list for proper documentation / examples.

PR #570, opened 2019-07-02
Adding hostname resolution to peer

iroddis:
Adds automatic peer hostname resolution, as `Peer::hostname_` is not currently populated by any means.

This removes the const-ness of ::hostname() and the ostream printer. I thought about adding a '::resolveHostname()' method, but if someone asks for the hostname, resolution will need to happen regardless.

Added a unit test, but it may fail as it assumes that the resolution sequence ("localhost" -> ip -> hostname), hostname = "localhost".

### DIFF
--- a/include/pistache/peer.h
+++ b/include/pistache/peer.h
@@ -41,7 +41,7 @@ public:
     ~Peer();
     
     const Address& address() const;
-    const std::string& hostname() const;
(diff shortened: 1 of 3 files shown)


iroddis:
I think I've been hit by the Travis bug again. Is it possible to resubmit the CI builds again, please?

PR #575, opened 2019-07-03
Set Thread name for running HTTP server threads

mohsenomidi:
Setting the thread name on the server mode, it can be used to set a unique name for a thread, which can be useful for debugging  multi-threaded applications, if you are starting more than one server on different ports (the threads on each port cab be set by one name) .

**.threadsName("MyHTTPServer")**

Usage : 
```
std::shared_ptr<Http::Endpoint> mHttpEndpoint_;

auto opts = Http::Endpoint::options()
          .threads(3)
          .threadsName("MyHTTPServer")
      mHttpEndpoint_->init(opts);
```

### DIFF
--- a/include/pistache/endpoint.h
+++ b/include/pistache/endpoint.h
@@ -19,7 +19,7 @@ public:
         friend class Endpoint;
 
         Options& threads(int val);
-        Options& threadsName(const char* val);
+        Options& threadsName(std::string val);
         Options& flags(Flags<Tcp::Options> flags);
         Options& backlog(int val);
         Options& maxRequestSize(size_t val);
@@ -30,7 +30,7 @@ public:
 
     private:
         int threads_;
-        char threadsName_[16];
+        std::string threadsName_;
         Flags<Tcp::Options> flags_;
         int backlog_;
         size_t maxRequestSize_;
--- a/include/pistache/endpoint.h
+++ b/include/pistache/endpoint.h
@@ -19,7 +19,7 @@ public:
         friend class Endpoint;
 
         Options& threads(int val);
-        Options& threadsName(std::string val);
+        Options& threadsName(const std::string& val);
         Options& flags(Flags<Tcp::Options> flags);
         Options& backlog(int val);
         Options& maxRequestSize(size_t val);
--- a/include/pistache/endpoint.h
+++ b/include/pistache/endpoint.h
@@ -19,6 +19,7 @@ public:
         friend class Endpoint;
 
         Options& threads(int val);
+        Options& threadsName(const char* val);
         Options& flags(Flags<Tcp::Options> flags);
         Options& backlog(int val);
         Options& maxRequestSize(size_t val);
@@ -29,6 +30,7 @@ public:
 
     private:
         int threads_;
+        char threadsName_[16];
         Flags<Tcp::Options> flags_;
         int backlog_;
         size_t maxRequestSize_;
--- a/include/pistache/listener.h
+++ b/include/pistache/listener.h
@@ -52,7 +52,7 @@ public:
     void init(
             size_t workers,
             Flags<Options> options = Options::None,
-            std::string workersName = "",
+            const std::string& workersName = "",
             int backlog = Const::MaxBacklog);
     void setHandler(const std::shared_ptr<Handler>& handler);
 
--- a/include/pistache/listener.h
+++ b/include/pistache/listener.h
@@ -52,7 +52,7 @@ public:
     void init(
             size_t workers,
             Flags<Options> options = Options::None,
-            const char * workersName = "",
+            std::string workersName = "",
             int backlog = Const::MaxBacklog);
     void setHandler(const std::shared_ptr<Handler>& handler);
 
@@ -66,7 +66,7 @@ public:
     void runThreaded();
 
     void shutdown();
- 
+
     Async::Promise<Load> requestLoad(const Load& old);
 
     Options options() const;
@@ -88,7 +88,7 @@ private:
     std::thread acceptThread;
 
     size_t workers_;
-    char   workersName_[16];
+    std::string workersName_;
     std::shared_ptr<Handler> handler_;
 
     Aio::Reactor reactor_;
--- a/include/pistache/listener.h
+++ b/include/pistache/listener.h
@@ -52,6 +52,7 @@ public:
     void init(
             size_t workers,
             Flags<Options> options = Options::None,
+            const char * workersName = "",
             int backlog = Const::MaxBacklog);
     void setHandler(const std::shared_ptr<Handler>& handler);
 
@@ -65,7 +66,7 @@ public:
     void runThreaded();
 
     void shutdown();
-
+ 
     Async::Promise<Load> requestLoad(const Load& old);
 
     Options options() const;
@@ -87,6 +88,7 @@ private:
     std::thread acceptThread;
 
     size_t workers_;
+    char   workersName_[16];
     std::shared_ptr<Handler> handler_;
 
     Aio::Reactor reactor_;
--- a/include/pistache/reactor.h
+++ b/include/pistache/reactor.h
@@ -176,9 +176,10 @@ public:
 
 class AsyncContext : public ExecutionContext {
 public:
-    explicit AsyncContext(size_t threads)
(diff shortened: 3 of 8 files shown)


include/pistache/listener.h
```diff
@@ -65,7 +66,7 @@ class Listener {
     void runThreaded();
 
     void shutdown();
-
+
```
dennisjenkins75: Please remove no-op formatting changes.

include/pistache/listener.h
```diff
@@ -87,6 +88,7 @@ class Listener {
     std::thread acceptThread;
 
     size_t workers_;
+    char   workersName_[16];
```
dennisjenkins75: Please conform to the existing code style.  Only one space between type ("char") and name ("workersName").

include/pistache/reactor.h
```diff
@@ -176,9 +176,10 @@ class SyncContext : public ExecutionContext {
 
 class AsyncContext : public ExecutionContext {
 public:
-    explicit AsyncContext(size_t threads)
+    explicit AsyncContext(size_t threads,const char* threadsName = "")
```
dennisjenkins75: Space after comma, before "const"

include/pistache/reactor.h
```diff
@@ -176,9 +176,10 @@ class SyncContext : public ExecutionContext {
...
         : threads_(threads)
-    { }
+
```
dennisjenkins75: Remove empty line here.

include/pistache/reactor.h
```diff
@@ -176,9 +176,10 @@ class SyncContext : public ExecutionContext {
...
+    { strcpy(threadsName_,threadsName);}
```
dennisjenkins75: Make this 3 lines:
{
  strcpy(foo, bar);
}

include/pistache/listener.h
dennisjenkins75: Why is this a char[16]?  Please use the std::string type.

include/pistache/endpoint.h
```diff
@@ -29,6 +30,7 @@ class Endpoint {
 
     private:
         int threads_;
+        char threadsName_[16];
```
dennisjenkins75: Please use std::string, not char[].

include/pistache/reactor.h
```diff
@@ -188,6 +189,7 @@ class AsyncContext : public ExecutionContext {
 
 private:
     size_t threads_;
+    char threadsName_[16];
```
dennisjenkins75: Please use std::string instead of char[].

src/common/reactor.cc
```diff
@@ -153,6 +153,8 @@ class SyncImpl : public Reactor::Impl {
     void run() override {
         handlers_.forEachHandler([](const std::shared_ptr<Handler> handler) {
             handler->context_.tid = std::this_thread::get_id();
+
```
dennisjenkins75: Remove this un-needed change.

src/common/reactor.cc
```diff
@@ -334,11 +336,11 @@ class AsyncImpl : public Reactor::Impl {
 
     static constexpr uint32_t KeyMarker = 0xBADB0B;
 
-    AsyncImpl(Reactor* reactor, size_t threads)
+    AsyncImpl(Reactor* reactor, size_t threads,const char *threadsName)
```
dennisjenkins75: One space after the comma.  Here and in the rest of this PR.

include/pistache/reactor.h
dennisjenkins75: Also, don't use strcpy().

mohsenomidi:
i will review your comments, amd fix the issues at weekend.
thanks

mohsenomidi:
Hello Dennis, 

I made the changes, you mentioned above about the code style and using std::string instead of char*.
about your last comment 

> Do not assume that all threading implementations use the same max length thread name. iirc, Windows lets you set a 10-char thread name. IDK about FreeBSD or Solaris. Someday we'll port pistache to an OS beyond Linux. Just use an arbitrary length string.

i dont know how can i handle multi OS support for this PR, i am not familiar with Windows nor Mac, BSD systems. 

if you don't specify the length for thread name and this string length is more than 16 chars including '\0'  (in Linux), the pthread_setname_np function will return the error (code:34), it must be have in the specify length.

Thanks.

include/pistache/endpoint.h
```diff
@@ -19,6 +19,7 @@ class Endpoint {
         friend class Endpoint;
 
         Options& threads(int val);
+        Options& threadsName(std::string val);
```
dennisjenkins75: const std::string& (pass by ref, not pass by value, since we are not using X-value/move semantics).

include/pistache/listener.h
```diff
@@ -52,6 +52,7 @@ class Listener {
     void init(
             size_t workers,
             Flags<Options> options = Options::None,
+            std::string workersName = "",
```
dennisjenkins75: const std::string&

include/pistache/reactor.h
```diff
@@ -176,9 +176,12 @@ class SyncContext : public ExecutionContext {
 
 class AsyncContext : public ExecutionContext {
 public:
-    explicit AsyncContext(size_t threads)
+    explicit AsyncContext(size_t threads, std::string threadsName = "")
```
dennisjenkins75: const std::string&

src/common/reactor.cc
```diff
@@ -334,11 +334,11 @@ class AsyncImpl : public Reactor::Impl {
 
     static constexpr uint32_t KeyMarker = 0xBADB0B;
 
-    AsyncImpl(Reactor* reactor, size_t threads)
+    AsyncImpl(Reactor* reactor, size_t threads, std::string threadsName)
```
dennisjenkins75: const std::string&

src/server/endpoint.cc
```diff
@@ -26,6 +26,12 @@ Endpoint::Options::threads(int val) {
     return *this;
 }
 
+Endpoint::Options&
+Endpoint::Options::threadsName(std::string val) {
```
dennisjenkins75: const std::string&

src/server/listener.cc
```diff
@@ -105,7 +107,9 @@ Listener::~Listener() {
 void
 Listener::init(
     size_t workers,
-    Flags<Options> options, int backlog)
+    Flags<Options> options,
+    std::string workersName,
```
dennisjenkins75: const std::string&

mohsenomidi:
Dear Dennis, 

All requested changes has been done.

Thanks

mohsenomidi:
Dear Dennis, 

i add unit test for initializing the endpoint with EXPECT_NO_THROW.

tests/threadname_test.cc
```diff
@@ -0,0 +1,30 @@
...
+{
+    std::shared_ptr<Http::Endpoint> mHttpEndpoint_;
+    Address addr(Ipv4::any(), Pistache::Port(0));
+    mHttpEndpoint_ = std::make_shared<Http::Endpoint>(addr);
+
+    auto test_options = Http::Endpoint::options()
+          .threads(2)
+          .threadsName(pThreadName);
+    mHttpEndpoint_->init(test_options);
+}
```
dennisjenkins75: The actual call to pthread_setname_np() is inside Worker::run().  Is that code path exercised via this test?  Hint: temporarily add a call to "abort()" right after pthread_setname_np(), then run this test.  If the test does not crash, then you're not testing the call all the way down to the bottom of the stack.  You might need to extend this test to actually start the server.  Of course, it is also possible that your code does already test your change - I'm only eye-balling your PR, I've not actually cloned it to my local workstation and tested it myself (yet).

mohsenomidi:
I tested this PR code in my local machine and it works perfectly, without any exception in different name lengths, but you need the full test case to prove this, lets make the full scenario the same as  http_server_test.cc with setting name in thread with your preferred values. 

i will do it in few days.

mohsenomidi:
Dear Dennis,

i create the sever thread with name and client example to send & receive data, this is actual and real test case that show full functionality of HTTP server and client with setting the name in server side app.

kiplingw:
Pushed to unstable PPA.

Issue #239, opened 2018-06-12
How to determine the IP address of the requesting client?


CarPetBay:
When in onRequest of an own Handler, how could one determine the IP address of the requesting client?

kiplingw:
I was wondering the same. The Request class has a method peer(), but it appears to be ommitted at compile time in http.h:

`#ifdef LIBSTDCPP_SMARTPTR_LOCK_FIXME
    std::shared_ptr<Tcp::Peer> peer() const;
#endif`

dennisjenkins75:
There are several PRs which address this issue.  I'm hoping to merge #205 within a few days.

lnihlen:
Pinging this open issue. It would still be super helpful to be able to determine IP address. Thanks!

dennisjenkins75:
You need to rebuild your local pistache with the above mentioned symbol defined.  The original author of pistache left this undefined because it caused a performance regression and his use-case did not require the server knowing the client IP address.  Maybe someday someone will do some extensive performance tests with the option enabled and disabled, post the results and maybe we'll change it.  Or anyone could submit a PR that makes a deep copy of the peer's IP/Port (but not the entire Peer object).

You have another option.  If your pistache process sits behind any web proxy (squid, nginx, etc...), you can likely get the client's original IP address and port from the HTTP request headers.

```
$ find src/ include/ -name "*.cc" -o -name "*.h" | sort | xargs egrep "LIBSTDCPP_SMARTPTR_LOCK_FIXME"
include/pistache/http.h:#ifdef LIBSTDCPP_SMARTPTR_LOCK_FIXME
include/pistache/http.h:#ifdef LIBSTDCPP_SMARTPTR_LOCK_FIXME
include/pistache/http.h:#ifdef LIBSTDCPP_SMARTPTR_LOCK_FIXME
src/common/http.cc:#ifdef LIBSTDCPP_SMARTPTR_LOCK_FIXME
src/common/http.cc:#ifdef LIBSTDCPP_SMARTPTR_LOCK_FIXME
```

daljitmahi:
I still see this issue in server side, I get peer from Pistache::Http::ResponseWriter class (response.peer()) and then get peer->hostname() or peer->address(), I get  I get a string with value "::".

dennisjenkins75:
I use this code, but I also have ipv6 disabled.  The "::" suggests to me that you have ipv6 enabled.

```
// Returns Client IP Address, either from "X-Forward-For" (if present),
[...]

PR #593, opened 2019-08-16
adds address copying to request, and test

lnihlen:
For issue #239, this copies the Address from the requesting Peer into the Request.


lnihlen:
Thanks for review and merge!

Issue #566, opened 2019-06-25
Http::serveFile doesn't work when using HTTPS


tonylin0826:
Hi, I'm currently working on a HTTPS server using Pistache and found out that when using HTTPS, the function Http::serveFile will not send using SSL. It only send SSL packet when buffer type is raw. Here is what I found.

```c++
if (buffer.isRaw()) {
    auto raw = buffer.raw();
    auto ptr = raw.data().c_str() + totalWritten;

#ifdef PISTACHE_USE_SSL
    auto it = peers.find(fd);

    if (it == std::end(peers))
        throw std::runtime_error("No peer found for fd: " + std::to_string(fd));

    if (it->second->ssl() != NULL) {
        bytesWritten = SSL_write((SSL*)it->second->ssl(), ptr, len);
    }
    else {
#endif /* PISTACHE_USE_SSL */
        bytesWritten = ::send(fd, ptr, len, flags);
#ifdef PISTACHE_USE_SSL
    }
#endif /* PISTACHE_USE_SSL */
}
else {
    auto file = buffer.fd();
    off_t offset = totalWritten;
    bytesWritten = ::sendfile(fd, file, &offset, len);
}
```

dlinten:
I do not know if this will help, but you could force a raw buffer by using ::send, this way you can fiddle with Mime types and also get all the functionality of a proper response object.  I like the stream approach too, because at some point you're going to want to inspect the stream, e.g. to add tracking info, auto copyright stamp images etc.  all becomes possible, therefore ::sendFile for me if of very little use.

Example for stream approach to load a *simple* resource from disk e.g. and use the response object:

```c++
std::ifstream file_stream(res_path);
response.send(Pistache::Http::Code::Ok, std::string((std::istreambuf_iterator<char>(file_stream)),
    std::istreambuf_iterator<char>()));
```

kiplingw:
Hey @tonylin0826. I've noticed the same problem. I just discovered your issue after posting https://github.com/oktal/pistache/issues/549#issuecomment-541979096. I can concur that all other client-server transactions seem to work fine, except for any that involve a code path to serveFile. My guess is the way it is implemented it isn't using OpenSSL correctly.

Issue #549, opened 2019-06-11
std::runtime error at second curl request when using ssl


timijntema:
I might have found an issue with the Pistache https/ssl support. I used the following code:

```C++
#include <pistache/endpoint.h>
#include <iostream>

using namespace Pistache;

struct HelloHandler : public Http::Handler {
    HTTP_PROTOTYPE(HelloHandler)

    void onRequest(const Http::Request& request, Http::ResponseWriter writer) {
        writer.send(Http::Code::Ok, "Hello, World!");
    }
};

int main(void) {
    std::cout << "Starting server\n";
    Pistache::Address addr;
    auto opts = Http::Endpoint::options().threads(1);

    addr = Pistache::Address(Pistache::Ipv4::any(), Pistache::Port(9999));

    Http::Endpoint server(addr);
    server.init(opts);
    server.setHandler(std::make_shared<HelloHandler>());
    server.useSSL("./cas/server/server.crt", "./cas/server/server.key");
    server.useSSLAuth("./cas/rootCA/rootCA.crt");
    std::cout << "SSL enabled\n";

    server.serve();
    return 0;
}
```

The certificates where generated using the terminal script below. WARNING: The certificates generated by the terminal script might not be fully save. I am no expert in SSL and I chose to remove the password from the certificate generation. Make sure you check if the certificates are save before you use them.

```sh
#!/bin/bash

# This script was executed from inside the cas directory seen in the C++ code above

# Generate root CA
echo "Generating rootCA"
mkdir rootCA
cd rootCA
openssl genrsa -out rootCA.key 4096
openssl req -x509 -new -nodes -key rootCA.key -subj "/C=NL/ST=UTRECHT/O=Some company/CN=rootca" -sha256 -days 1024 -out rootCA.crt
cd ..

# Generate server certificate
echo "\nGenerating server certificate"
mkdir server
cd server
openssl genrsa -out server.key 2048
openssl req -new -sha256 -key server.key -subj "/C=NL/ST=UTRECHT/O=Some company/CN=Some ip" -out server.csr
openssl x509 -req -in server.csr -CA ../rootCA/rootCA.crt -CAkey ../rootCA/rootCA.key -CAcreateserial -out server.crt -days 500 -sha256
cd ..

# Generate client certificate
echo "\nGenerating client certificate"
mkdir client
cd client
openssl genrsa -out client.key 2048
[...]

PR #617, opened 2019-10-15
Added basic_tls_request_with_servefile unit test to try and reproduce…

kiplingw:
I've added a `basic_tls_request_with_servefile` unit test. This is to try and reproduce issues #549 and #566. It is expected to hopefully fail until those issues (which are likely related) are patched.

The call to `curl_easy_perform` fails with CURLE_RECV_ERROR (56).

```
Expected equality of these values:
  res
    Which is: 56
  CURLE_OK
    Which is: 0
```

I may be doing something incorrect in setting up the unit test, or this could be curl's way of tripping over the SSL error. If the latter, then the test is failing as expected. Otherwise if someone could help me (e.g. @hydratim or @dennisjenkins75)  ensure the environment setup is correct, that would be helpful.


kiplingw:
Hey @dennisjenkins75. They were renumbered in the sense that the port each test uses after the one I inserted was incremented to prevent trying to bind to an already in use port. I inserted `basic_tls_request_with_servefile` where I did based on what I subjectively thought was logical proximity. 

The one before it is basic TLS without any authentication. The new one is also basic TLS without any authentication, but doing something more complex over the connection than its preceding test. All the tests after that involve some form of authentication.

kiplingw:
Try that.

kiplingw:
Done. Thanks @dennisjenkins75 and @Ne02ptzero. Pushed to unstable PPA.

PR #626, opened 2019-10-28
Raw and Cookie header case-insensitive

### LINKED ISSUE #604: Headers case-sensitive (or at least the cookie header)
Maybe relevant? https://github.com/oktal/pistache/issues/86

Headers should be case-insensitive but at least the cookie header seems to be case sensitive, trust me I know it is because I found out after 3 days ;)

adamburgess:
fixes #604

I'm using pistache with nginx, with HTTP2 -> nginx proxy -> Pistache. As HTTP2 lowercases all headers, Cookie and other raw headers weren't working.

This PR:

* Adds case-insensitive matching for the Cookie and Set-Cookie header
* Adds case-insensitive lookups/hashing for rawHeaders
* Refactors the case-insensitive comparing to use less memory (_how is this actually not in c++ officially, omg_)
* Adds tests for raw headers and cookies

### DIFF
--- a/tests/headers_test.cc
+++ b/tests/headers_test.cc
@@ -713,3 +713,29 @@ TEST(headers_test, raw_headers_are_case_insensitive)
     ASSERT_FALSE(request.headers().tryGetRaw("Custom-Header").isEmpty());
     ASSERT_FALSE(request.headers().tryGetRaw("custom-header").isEmpty());
 }
+
+
+TEST(headers_test, cookie_headers_are_case_insensitive)
+{
+    {
+        std::string line = "cookie: x=y\r\n";
+        Pistache::RawStreamBuf<> buf(&line[0], line.size());
+        Pistache::StreamCursor cursor(&buf);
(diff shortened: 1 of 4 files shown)


tests/headers_test.cc
```diff
@@ -700,3 +700,42 @@ TEST(headers_test, registered_header_in_raw_list)
...
+
+    ASSERT_FALSE(request.headers().tryGetRaw("Custom-Header").isEmpty());
+    ASSERT_FALSE(request.headers().tryGetRaw("custom-header").isEmpty());
+}
+
+
+TEST(headers_test, cookie_headers_are_case_insensitive)
+{
+    {
+        std::string line = "cookie: x=y\r\n";
```
dennisjenkins75: Please add tests where the 'line' is in mixed or upper case.

adamburgess: Done.

adamburgess:
👍 

thanks!

PR #632, opened 2019-11-01
Added overload for ResponseWriter::send for raw byte stream...

kiplingw:
Added overload for ResponseWriter::send for raw byte stream. I also migrated some of the code that was still using char pointers to represent raw bytes to use std::byte. The former is deprecated now.


Ne02ptzero:
LGTM :+1:

kiplingw:
@dennisjenkins75, @Ne02ptzero and I were talking about this tonight in \#pistache. This solution works but one concern I have, and you probably will too, is std::byte wasn't introduced until C++17. We have two choices. We can either require applications to be C++ >= 17, or we can keep using deprecated char pointer for raw binary data.

kiplingw:
Yeah it's ready for merge. I removed the std::byte C++ 17 dependency for now until we're ready to move to 2017 standard. All unit tests passed along with autopkgtest. I've also pushed to unstable PPA.

PR #636, opened 2019-11-11
Fix for segmentation fault caused by incorrect lambda capture in promise

DmtKats:
I am experiencing an issue with a single worker thread pistache server, debug build. I am serving a simple web page. I connect to the web page and repeatedly press refresh (F5) on my browser (Firefox). After a couple of resets I get a segfault.

Looking through the call stack and after some analysis, it seems that there is a promise created that captures its local scope. In edge cases, the promise would resolve after the object was deleted. The internal "Core" shared pointer would be cleared and this would lead to the segmentation fault.

The trigger for this case is getting an EPIPE or ECONNRESET, which should probably not be issuing a response anyhow since the fd is broken.

I can see that lambdas capturing the entire local scope is common in the code base. I would suggest that you avoid doing so to avoid referencing invalidated objects.

Also the error case for the asyncWriteImpl code by default tries to respond to the remote with a failed write error, which seems counter intuitive since not being able to write to the remote is probably an indication that you can't send data to him any more. I think it would be better if you explicitly specified the cases where it makes sense to send the error as a response. However, since I do not have much experience with the code I opted to take the least drastic approach.

### DIFF
--- a/include/pistache/async.h
+++ b/include/pistache/async.h
@@ -570,15 +570,18 @@ namespace Async {
                 template<typename P>
                 void finishResolve(P& promise) {
                     auto chainer = makeChainer(promise);
-                    promise.then(std::move(chainer), [=](std::exception_ptr exc) {
-                        auto core = this->chain_;
-                        core->exc = std::move(exc);
-                        core->state = State::Rejected;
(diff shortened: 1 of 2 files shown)


kiplingw:
Pushed to unstable PPA.

PR #641, opened 2019-11-26
optional::operator== improvements

kuzkry:
1st commit : I don't think we need the extra type U which is only provided to has_equalto_operator<U> metafunction.
2nd commit : when a user provides a non-comparable type, then we state what they've done wrong instead of letting a compiler complain about the lack of operator== in optional<T>.
3rd commit : previous didn't test optional

However, now has_equalto_operator<Optional<not_comparable>>::value will return a different value (but it's still technically correct) :D @muttleyxd what do you think?

### DIFF
--- a/include/pistache/optional.h
+++ b/include/pistache/optional.h
@@ -240,13 +240,15 @@ public:
         }
     }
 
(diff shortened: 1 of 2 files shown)


kiplingw:
Thanks @kuzkry. Pushed to unstable PPA.

PR #648, opened 2020-01-01
Fixed stack corruption when Aio::AsyncImpl constructed with large number of worker threads.

kiplingw:
I've been performing some R&D porting an experimental software product to high end hardware. It is an IBM Romulus system with a version 2.2 (pvr 004e 1202) POWER9 processor.

The system has two physical CPUs, each with 18 physical cores, and each in turn with 4 logical cores ("threads" in the IBM parlance). This means that when an application queries the platform with `std::thread::hardware_concurrency()`, it will learn of 144 logical cores. But the application may (as in my case) multiply this by a constant factor (2) to account for some threads being i/o bound from time to time and to therefore account for overlapping i/o. 

That means a lot of file descriptors if you manage to create that many worker threads. Pistache needs to construct six file descriptors for each worker thread. Five `eventfd`'s and one `eventpoll` for each thread. Putting aside any `RLIMIT_NOFILE` process limit on the maximum number of file descriptors (this isn't Pistache's problem), which was 1024 on my system, there's still another problem.

Pistache cannot handle any more than 255 worker threads when using the asynchronous reactor implementation. The reason is due to an architectural design flaw in the reactor which uniquely identifies each worker thread by a fixed width of 8 bits. Understandably few sane people would have run into this limitation until now without having this strange hardware. From `src/common/reactor.cc`:

```
    struct HandlerList {

        // We are using the highest 8 bits of the fd to encode the index of the handler,
        // which gives us a maximum of 2**8 - 1 handler, 255
        static constexpr size_t HandlerBits = 8;
        static constexpr size_t HandlerShift = sizeof(uint64_t) - HandlerBits;
        static constexpr uint64_t DataMask = uint64_t(-1) >> HandlerBits;

        static constexpr size_t MaxHandlers = (1 << HandlerBits) - 1;
```

`MaxHandlers` is 255 and is the upper limit on the number of worker threads. As long as file descriptors with only 8 vacant leading bits are used to track handlers, there's not really any way around this. As a result if you tried to create more than 255 threads, and if you were lucky, your process would only crash. If you were unlucky your stack would become corrupted as `AsyncImpl::addHandler` would cause a buffer overflow in overwriting the end of `keys`.

What I've done is gone ahead and fixed this by preventing the possibility in `AsyncImpl`. It now performs bounds checking on the number of requested worker threads. I've also added some additional bounds checking because using the subscript operator on an STL container is dangerous since there is no automatic bounds checking. Use `at()` whenever possible.

Since @dennisjenkins75  would have asked for it anyways, I've gone ahead and also added a `reactor_exceed_max_threads` test to the `reactor_test` test suite.


xinthose:
Are you doing stock trading or something?

kiplingw:
Hah, almost. No, it's [music](https://www.heliosmusic.io) software. But nearly did go the high frequency trading software domain back in 2016.

Issue #164, opened 2017-11-20
[Question] Endpoint restart


goldstar2154:
Maybe it is not correct behavior, but i need to restart endpoint from time to time during program runs.
I use code like this

```
my_endpoint->shutdown();
// delete my_endpoint
// create new my_endpoint
// init new endpoint
// install new handler
my_endpoint->serveThreaded();

```
and got "epoll_ctl(epoll_fd, EPOLL_CTL_ADD, fd, &ev): Bad file descriptor". This happen because epoll epoll_fd do not close when calling shutdown()?

If i restart entire child process, where endpoint live, then all fine, cause Pistache can handle terminate signal with `Pistache::Tcp::Options::InstallSignalHandler`.

As i sad, maybe my case is not correct for pistache, but why does shutdown dont close epoll fd when calling shutdown?

P.S. i also try use `Pistache::Tcp::Options::ReuseAddr` flag, but unsuccessfully.

Thx & best regards.

xoac:
Hi,
Maybe u overwrite `options` instead of add new one? Haw u saw #137 ?

goldstar2154:
Hello @xoac . Thanks for responce but problem is not in flags. Problem is in non-closed fd (as usually :)

first: Poller class (os.cc) i add code
```
    Epoll::~Epoll()
    {
        if (epoll_fd > 0)
        {
            close(epoll_fd);
        }
    }
```
second: Listener class (listener.cc) add code
```
Listener::~Listener() {
    if (isBound()) shutdown();
    if (acceptThread) acceptThread->join();
   
    // added lines
    if (listen_fd > 0)
    {
        close(listen_fd);
        g_listen_fd = 0;
        listen_fd = 0;
    }
}
```
third: i found problematic semantics in Listener::bind(const Address& address) (listener.cc):
code:
```
    int fd = -1;
    for (struct addrinfo *addr = addrs; addr; addr = addr->ai_next) {
        fd = ::socket(addr->ai_family, addr->ai_socktype, addr->ai_protocol);
        if (fd < 0) continue;

        setSocketOptions(fd, options_);

        if (::bind(fd, addr->ai_addr, addr->ai_addrlen) < 0) {
            close(fd);
            continue;
        }

        TRY(::listen(fd, backlog_));
        break;
    }
    make_non_blocking(fd);
    poller.addFd(fd, Polling::NotifyOn::Read, Polling::Tag(fd));
    listen_fd = fd;
    g_listen_fd = fd;
```
[...]

Issue #150, opened 2017-10-11
http.cc : line 672. opened file descriptor


goldstar2154:
Hello everyone.
For first thx for good and really simple http library.
For second:
There is bug for Pistache::Http::serveFile(...) behavior i think;
in http.cc line 672
and in stream.h line 159
code http.cc:
```
    int fd = open(fileName, O_RDONLY);
    if (fd == -1) {
        /* @Improvement: maybe could we check for errno here and emit a different error
            message
        */
        throw HttpError(Http::Code::Not_Found, "");
    }

    int res = ::fstat(fd, &sb);
    close(fd);   // added!
    if (res == -1) {
        throw HttpError(Code::Internal_Server_Error, "");
    }
```
code stream.h
```
struct FileBuffer {
    FileBuffer() { }
    ~FileBuffer();   // added1!

    FileBuffer(const char* fileName);
    ....
```
code stream.cc
```
FileBuffer::~FileBuffer()
{
    close(fd_);
}
```
This bug lead to `std::runtime_error() : Too many open files` exception when server serving big numbers of static files (number of requests must be higher than system open files limit)

Environment:
Ubuntu 16.04.03
standart open file limit - 1024 (enough for testing)
nginx as https proxy (https=>http)
pistache as http backend
testing with ab: ab -n 5000 *my api route to png image*

After rebuilding pistache lib with `close(fd);` ab runs good (with 100k request and 100 concurency)
I can write minimal example for reproducing bug if you need.
Best regards.

Upd1:
Also problem can be fixed in Transport::asyncWriteImpl (transport.cc line 186) in lambda cleanup, like this:
```
    auto cleanUp = [&]() {
        if (buffer.isRaw()) {   // We also have implementation for raw buffers, but no for files
            auto raw = buffer.raw();
            if (raw.isOwned) delete[] raw.data;
        }
        else
        if (buffer.isFile())   // This is not really correct fix just for testing
        {
            close(buffer.fd());
        }
```

Issue #43, opened 2017-01-24
epoll_ctl(epoll_fd, EPOLL_CTL_ADD, fd, &ev): Bad file descriptor error caught in a multithreaded application


nabilasalmi:
Hi,

I had an error on my application following the "Getting started example", implementing a Rest Server in a multithreaded application. Sometimes, an exception is caught . It's description is:

epoll_ctl(epoll_fd, EPOLL_CTL_ADD, fd, &ev): Bad file descriptor.

What lets this error occur ??

Thanks a lot for your help !

Nabila

mic90:
Hi,

Just went here to look for answer to similar problem. 
Here is sample code to reproduce: 

```
#include <iostream>
#include <atomic>
#include <endpoint.h>
#include <signal.h>

using namespace std;
using namespace Net;

bool stop(false);

void signalHandler(int signal)
{
    cout << "Signal: " << signal << endl;
    stop = true;
}

class HelloHandler : public Http::Handler
{
public:

    HTTP_PROTOTYPE(HelloHandler)

    void onRequest(const Http::Request& request, Http::ResponseWriter response)
    {
        cout << "Received request " << request.resource() << endl;
        response.send(Http::Code::Ok, "Hello, World");
    }
};

int main()
{
    signal(SIGINT, signalHandler);
    signal(SIGILL, signalHandler);
    signal(SIGKILL, signalHandler);

    cout << "Starting rest server" << endl;
    Net::Address addr(Net::Ipv4::any(), Net::Port(9080));

    auto opts = Http::Endpoint::options().threads(1);
    Http::Endpoint server(addr);
    server.init(opts);
    server.setHandler(std::make_shared<HelloHandler>());
    server.serveThreaded();
    cout << "Server started" << endl;
    while(!stop)
    {
    }
    server.shutdown();
    cout << "Exiting" << endl;
    return 0;
}
```
If application is started and no request were made - I can close the application and start it again without problems.
If application is started and one or more requests were made - when I close the application and immediately start it again I got the same error as Nabila:
```
terminate called after throwing an instance of 'std::runtime_error'
  what():  epoll_ctl(epoll_fd, EPOLL_CTL_ADD, fd, &ev): Bad file descriptor
Aborted (core dumped)
```
Given error pops out for about a minute, and then I can start application without problems again.

Best regards,
Michal

DrLex0:
[...]

PR #650, opened 2020-01-07
close non-closed file descriptor will resolve : #38, #43, #150, #164, #221 - shutdown will work properly 

mohsenomidi:
This PR introduces changes which resolve these issues (#38, #43, #150, #164, #221).

the file descriptor will stay open at the end of the Listener and Epoll life-time, there for the shutdown will not work properly, and may get the Bad file descriptor in multi threading and also Too many open files. 
it just close the listener port and FD and close FD in epoll deconstructor.

### DIFF
--- a/src/common/os.cc
+++ b/src/common/os.cc
@@ -147,6 +147,14 @@ namespace Polling {
        epoll_fd = TRY_RET(epoll_create(Const::MaxEvents));
     }
 
+    Epoll::~Epoll()
+    {
+        if (epoll_fd > 0)
+        {
+            close(epoll_fd);
+        }
+    }
+        
     void
     Epoll::addFd(Fd fd, Flags<NotifyOn> interest, Tag tag, Mode mode) {
         struct epoll_event ev;
--- a/src/common/os.cc
+++ b/src/common/os.cc
@@ -149,7 +149,7 @@ namespace Polling {
 
     Epoll::~Epoll()
     {
-        if (epoll_fd > 0)
+        if (epoll_fd >= 0)
         {
             close(epoll_fd);
         }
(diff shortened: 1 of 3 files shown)


kiplingw:
Thanks @mohsenomidi. I think @dennisjenkins75 might be away on holidays right now. But once this is reviewed and merged it looks like a great addition because you've solved multiple problems with a fairly simple patch.

src/common/os.cc
```diff
@@ -147,6 +147,14 @@ namespace Polling {
        epoll_fd = TRY_RET(epoll_create(Const::MaxEvents));
     }
 
+    Epoll::~Epoll()
+    {
+        if (epoll_fd > 0)
```
dennisjenkins75: While it is unlikely that epoll_fd would ever be '0', this is technically possible.  The caller could close STDIN, then invoke a code path that creates an Epoll instance.  You should test for ">= 0".

src/server/listener.cc
```diff
@@ -95,6 +95,12 @@ Listener::~Listener() {
         shutdown();
     if (acceptThread.joinable())
         acceptThread.join();
+   
+    if (listen_fd > 0)
```
dennisjenkins75: Similar to comment for Epoll class.  Test for ">= 0", and set listen_fd to -1, not 0, after you manually close it.

src/common/os.cc
kiplingw: Yes, he's right. File descriptors are suppose to be always allocated from the lowest available ordinal.

mohsenomidi: Yap you are right, changes applied

src/server/listener.cc
mohsenomidi: Yap you are right, changes applied

PR #665, opened 2020-01-15
Added -std=c++14 to Cflags in pkg-config manifest

kiplingw:
libpistache.pc.in: Added -std=c++14 to Cflags in pkg-config manifest...
{debian/changelog,version.txt}: Bumped versioning...


kiplingw:
Quite frequently one or more of the unit tests fail on the PPA build server, but retrying the build gets them a pass. My guess is there are still some hard to find race conditions we'll eventually track down.

kiplingw:
@dennisjenkins75, I realized belatedly but this is actually a really bad idea. Consider the scenario where someone has build flags that require -std=c++17 or greater. Unintentionally, depending on when the `pkg-config --cflags` are inserted into CFLAGS, the compiler may end up overriding the 2017 standard with 2014. I'm sorry. That was my bad. Please revert if you can.

PR #678, opened 2020-01-20
Modify reactor.cc to c++14

### LINKED ISSUE #666: Compilation Problem
I follow the instruction to install pistache (build from source) on my ubuntu16.04.
But always get the follow error. 

image

ivan1993br:
I am using Pistache with [ROS](https://www.ros.org/), in [ubuntu xenial](http://br.releases.ubuntu.com/16.04/), so I need C++14, I want to use the GCC version distributed for my current system. When I started, I searched for other libraries to do the same, but Pistache was the best, considering simplicity, compatibility and features, so nice work to you all!

This is just a patch to allow the Pistache library to still be compiled with c++14. In my opinion, even thought we are in 2020, as this is an API, a higher compatibility should be aimed.

Hoping that you accept my PR,

closes #666 @dennisjenkins75 @kiplingw

### DIFF
--- a/src/common/reactor.cc
+++ b/src/common/reactor.cc
@@ -342,11 +342,11 @@ public:
   std::vector<std::shared_ptr<Handler>>
   handlers(const Reactor::Key &key) const override {
 
-    const auto [idx, marker] = decodeKey(key);
-    if (marker != KeyMarker)
+    const std::pair<uint32_t, uint32_t> idx_marker = decodeKey(key);
+    if (idx_marker.second != KeyMarker)
       throw std::runtime_error("Invalid key");
 
-    Reactor::Key originalKey(idx);
+    Reactor::Key originalKey(idx_marker.first);
 
     std::vector<std::shared_ptr<Handler>> res;
     res.reserve(workers_.size());


kiplingw:
If you are using Ubuntu, I would tell you to use the PPA. However, as you noticed already it doesn't build on Xenial. The development toolchain is too old and I think it is too much work to try and backport Pistache to C++14 while also accommodating new C++17 oriented patches going forward. 

I'd suggest upgrading to Eoan and just using the pre-compiled packages. Whatever hardware you are running on, `armhf` or what have you, it's probably already ported in the build server.

ivan1993br:
> If you are using Ubuntu, I would tell you to use the PPA. However, as you noticed already it doesn't build on Xenial. The development toolchain is too old and I think it is too much work to try and backport Pistache to C++14 while also accommodating new C++17 oriented patches going forward.
> 
> I'd suggest upgrading to Eoan and just using the pre-compiled packages. Whatever hardware you are running on, `armhf` or what have you, it's probably already ported in the build server.

Unfortunately for us upgrading to Eoan is not an option now, because we use the [ROS kinetic](http://wiki.ros.org/kinetic/) framework which is only maintained on Xenial.

For now, the only c++17 feature on the code is in this file, what I would hope is that the new patches would be kept on c++14 for now on, and also set the Pistache c++ version.

Pistache is upgrading on c++ version very quickly, [this commit](https://github.com/oktal/pistache/commit/d72ee8c7d932a3d900d599b92ad7d34b6914d4af) set to c++14 and now to c++17. Do you think that keeping c++17 would be better than to have more compatibility ?

kiplingw:
@ivan1993br, the ROS kinetic framework, and you can correct me if wrong, appears to be abandonware. The website was last updated almost four years ago. Xenial is about that age as well. I'd strongly encourage you to consider alternative solutions.

ivan1993br:
Actually, no, maybe the link to the website was not the best, even thought you can see it was modified on `Wiki: kinetic (last edited 2018-01-08 21:55:42 by TullyFoote)`.

ROS is a robotic framework very famous on the community, maintained by [OpenRobotics](https://www.openrobotics.org/) which has some [famous sponsors](https://www.openrobotics.org/sponsors), mainly because it is used in a lot of fields. Until now, when they say that they will maintain a package for some time, they really do. And the EOL for the kinetic branch is [April, 2021](http://wiki.ros.org/Distributions) alongside with Xenial. You can see on [this repository](https://github.com/ros/ros/tree/kinetic-devel) that their kinetic-devel branch is still be maintained.

Finally, for now it is not possible to change the ROS version, we have a CI setup to help us do that, but today we can't compromise the project doing this change.

kiplingw:
The ROS framework just appears to be mostly snapshots of older releases of various distros. If ROS is providing packages for its own upstream software, that's a whole other issue. But if you're mostly dependent on them shipping old downstream releases, then again, I strongly encourage you to consider upgrading your development toolchain and dependencies. I think it would make more sense for you to maybe spend your time submitting patches to ROS to bring it to 2020 then patching Pistache to take it back to 2016.

ivan1993br:
> The ROS framework just appears to be mostly snapshots of older releases of various distros. If ROS is providing packages for its own upstream software, that's a whole other issue. But if you're mostly dependent on them shipping old downstream releases, then again, I strongly encourage you to consider upgrading your development toolchain and dependencies. I think it would make more sense for you to maybe spend your time submitting patches to ROS to bring it to 2020 then patching Pistache to take it back to 2016.

I understand, but I do not see as bring Pistach back to 2016, but keep it simpler, and most compatible with other codes as possible, change only if there is a necessity. Unfortunately, ROS kinetic framework is defined to be compiled with c++11. The real problem happened because when I started writing the code, Pistache was compatible with C++11, and then I read on the [readme](https://github.com/oktal/pistache/blob/master/README.md): 'Pistache is a modern and elegant HTTP and REST framework for C++. It is entirely written in pure-C++14 and provides a clear and pleasant API.'. So I thought that it would stay on C++14.

I realize that Pistache has not a released version yet,but I think a C++ standard should be set, even if it is C++17, mainly for those who will use this API and insert it on a specific standard code. Change the development toolchain because of a library that you use is not something that you would want.

 I think, for now, if C++17 is kept, we will freeze Pistache on our side here, unfortunately I see no other quick fix.

Finally, thanks for the help @kiplingw !

ivan1993br:
> Given that this one line of code is the only c++17 in pistache atm, I'm willing to merge this PR. I won't want to harm a user over one line of code. However, do note that sooner or later we will unconditionally move into c++17 or c++2x land.

@dennisjenkins75 thanks for the merge.

Do you intend to move the code to c++17 right now or later?(Maybe if you intend to keep C++14 for a little longer, can we have a date?) I am asking that because we use Xenial due to the fact that the project started at the end of 2017, and that time the EOL linux distribution was Xenial, and the project will finish at the end of the year. I am asking this just to know if I should start to think about leaving Pistache for this project.

PR #679, opened 2020-01-20
Provide a way to build .RPMs

win32asm:
Suggested usage scenario:
        cmake <srcdir>
        make rpm

Generated packages:
	libpistache{,-devel,-static}-0.0.1.<githash>-1.el8.<arch>.rpm
	libpistache-0.0.1.<githash>-1.el8.src.rpm

Todo: provide consistent version naming (introduce tags?)
Note: as of 20.01.2020 debuginfo package cannot be created
      see https://bugzilla.redhat.com/show_bug.cgi?id=1206312
      main rpm contains libpistache.so with debuginfo


kiplingw:
Hey @win32asm. I don't think integrating package creation directly into the build system is prudent. As with the Debianization, we keep that metadata separate. I don't know what the Red Hat world's version of the Debian Policy Manual is, but there's likely something similar.

It should also have some way to perform DEP-8 continuous integration on the installed artifacts within a container, or whatever the Red Hat analogue happens to be. This ensures that users are shipped as high quality packages as possible, even if only in the form of unstable daily snapshots.

win32asm:
Ok, will try to integrate my fork on https://build.opensuse.org .

However, I would like to keep the sample spec file with the repository. Would it be ok if I removed the CMakeList.txt changes and resubmitted?

kiplingw:
There's no issue with including the spec in the repository. Usually with distro packaging the general best practise is to avoid having to modify the upstream build environment. The distro packaging metadata can be included in the repository, but you just don't want to have to modify the build environment to make them work. That's how the Debian Policy Manual works and I'm sure there's something analogous for the Red Hat world.

include/pistache/stream.h
```diff
@@ -159,12 +159,12 @@ class DynamicStreamBuf : public StreamBuf<char> {
   }
 
   RawBuffer buffer() const {
-    return RawBuffer((const char *)data_.data(), pptr() - &data_[0]);
+    return RawBuffer((const char*) data_.data(), pptr() - data_.data());
   }
 
   void clear() {
```
win32asm: @kiplingw , @dennisjenkins75 , any idea why we allow the stream to use "unallocated" vector's buffer?

kiplingw: No idea. But thank you for not using the subscript operator.

win32asm:
build and (some) tests succeed at https://build.opensuse.org/package/show/home:win32asm/libpistache

kiplingw:
Thanks @win32asm. You can take a look in debian/rules to see how an "upstream" tarball is generated. You can also see the required build and runtime dependencies in debian/control. You just need to figure out the equivalents on a Red Hat based system, if you haven't already.

The daily deb builds are generated from [here](https://code.launchpad.net/~pistache+team/+recipe/pistache-daily) and injected into the unstable PPA on the README.md [here](https://code.launchpad.net/~pistache+team/+archive/ubuntu/unstable). 

Once you've got stable and unstable daily package repositories setup, submit a PR to add to the README.md.

win32asm:
would it make sense to pick commits 4613b9f60572bc5e0447946e8904208d805c6a2e and fdb244f425ca1189851fa51e6aa8e351c18787d1 to be able to automagically build & test centos 8 packages?

hmrodrigues:
Hello,

So it seems that this PR doesn't work, at least with when rebased with current master branch.
I've created a spec file [here](https://github.com/hmrodrigues/pistache/blob/rpm/libpistache.spec). It's a orphan branch in case you want to have a separate branch like debian. I can try to make a make targe if desired.

This was tested on RockyLinux 8

win32asm:
a couple notes:
- pistache builds with meson on my centos 7 (though I will need to confirm where the packages come from - ideally it would be base and EPEL only), so I will try to target it now too.
- same spec probably should support at least centos-stream, rocky and alma, as well as SUSE and Fedora. We can start with any one
- Probably TravisCI can support RPM builds through respective docker images.
- meson would need a separate build target to prepare sources and the spec file (to include git hash ref or a build date as a part of package version)

I don't know whether I will have time to pursue any of those targets soon.
@hmrodrigues if you could add RPM build with meson to CI process for at least rocky, it would be great.

hmrodrigues:
@win32asm so for Rocky and Alma the same build should work, since they are binary compatible with RHEL.
>     * same spec probably should support at least centos-stream, rocky and alma, as well as SUSE and Fedora. We can start with any one

Tachi107:
@hmrodrigues we now use GitHub Actions, as Travis-ci.org shut down a while ago

PR #687, opened 2020-01-26
Handle exceptions on responce parse in client

hyperxor:
It's possible that `parse()` method can throw an exception. In the current client implementation uncaught exception can happen.

### DIFF
--- a/src/client/client.cc
+++ b/src/client/client.cc
@@ -545,25 +545,29 @@ Fd Connection::fd() const {
 }
 
 void Connection::handleResponsePacket(const char *buffer, size_t totalBytes) {
+  try {
+    parser.feed(buffer, totalBytes);
+    if (parser.parse() == Private::State::Done) {
+      if (requestEntry) {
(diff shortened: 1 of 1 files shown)


kiplingw:
Apt is the Debian package manager. What's amazing is that it should never timeout on Travis because they're suppose to be using a proxy so that it doesn't have to go and re-download all of those packages every time a build is submitted.

hyperxor:
@dennisjenkins75 Hello, unfortunatelly, I cannot restart build because I don't see any buttons that can be related to restart CI.

kiplingw:
Not sure why it is timing out. But on a related note Travis is configured right now to pull packages from a distro that's half a decade old. It's currently set to trusty [here](https://github.com/oktal/pistache/blob/master/.travis.yml#L3). Maybe time to bump that to eoan.

hyperxor:
@dennisjenkins75 **TIL** `git commit --allow-empty`, amazing!

PR #694, opened 2020-02-05
Add override keyword to functions who override

franku:
Added override keyword to reduce compiler warnings (C++11, -Wsuggest-override).

### DIFF
--- a/include/pistache/async.h
+++ b/include/pistache/async.h
@@ -47,7 +47,7 @@ namespace Async {
 
     class BadAnyCast : public std::bad_cast {
     public:
-        virtual const char* what() const noexcept { return "Bad any cast"; }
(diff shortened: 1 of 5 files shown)


include/pistache/async.h
```diff
@@ -47,7 +47,7 @@ namespace Async {
 
     class BadAnyCast : public std::bad_cast {
     public:
-        virtual const char* what() const noexcept { return "Bad any cast"; }
+        virtual const char* what() const noexcept override { return "Bad any cast"; }
```
dennisjenkins75: If you add "override", you should remove "virtual".

I am surprised that cppcheck did not catch this one.

franku: done

Issue #610, opened 2019-09-27
examples/http_client.cc fails to GET content larger than 4096 bytes including header


handrews42:
Consider the following server code:

```
#include <sstream>

#include <pistache/http.h>
#include <pistache/router.h>
#include <pistache/endpoint.h>

class EchoServer {
public:
  EchoServer(Pistache::Address addr)
    : httpEndpoint(std::make_shared<Pistache::Http::Endpoint>(addr)) {}

  void init(size_t thr = 2) {
    auto opts = Pistache::Http::Endpoint::options()
      .threads(thr);
    httpEndpoint->init(opts);
    setup();
  }

  void start() {
    httpEndpoint->setHandler(router.handler());
    httpEndpoint->serve();
  }

private:
  void setup() {
    Pistache::Rest::Routes::Get(router, "/:value/:count", Pistache::Rest::Routes::bind(&EchoServer::doEcho, this));
  }

  void doEcho(const Pistache::Rest::Request& request, Pistache::Http::ResponseWriter response) {
    std::string value = request.param(":value").as<std::string>();
    int count = request.param(":count").as<int>();

    std::stringstream ss;
    for (int i=0; i<count; ++i) { ss << value; }
    response.send(Pistache::Http::Code::Ok, ss.str());
  }

  std::shared_ptr<Pistache::Http::Endpoint> httpEndpoint;
  Pistache::Rest::Router router;
};

int main(int argc, char *argv[]) {
  Pistache::Port port(9080);
  Pistache::Address addr(Pistache::Ipv4::any(), port);

  EchoServer echo(addr);

  echo.init();
  echo.start();
}

```
It is adapted (simplified) from the example/rest_server.cc code in this repo.  When run and hit from a browser, it will repeat the ":value" field ":count" times as expected.  When using the examples/http_client.cc code, however, it works only when the GET response is less than ~4096 characters.  This can be verified by running the http_client.cc program with the following while the server above is running on localhost.

This works as anticipated:
`http_client http://localhost:9080/f/4036`

This fails to retrieve a response:
`http_client http://localhost:9080/f/4037`

Both of these work when accessing the URL from the browser.  I have also verified this using other web servers on my LAN.  If the response is larger than 4096 bytes, the client fails to process it.

I attempted to change  `MaxBuffer` in `include/pistache/config.h` to 16MB, but that resulted in a seg fault.  I didn't try any smaller values.

dennisjenkins75:
[...]

PR #695, opened 2020-02-06
Refactoring in http.h and http.cc

hyperxor:
1) Replaced inheritance to compisition in `ResponseStream` and `ResponseWriter`;
2) Fixed `/* @Refactor: code duplication */` comment: extracted `sendImpl` method in `ResponseWriter` class;
3) Move method definitions for `ResponseStream` and `ResponseWriter` to `http.cc`.

### DIFF
--- a/src/common/http.cc
+++ b/src/common/http.cc
@@ -567,19 +567,24 @@ std::shared_ptr<Tcp::Peer> Request::peer() const {
 }
 #endif
 
+ResponseStream::ResponseStream(ResponseStream &&other)
+    : response_(std::move(other.response_)), peer_(std::move(other.peer_)),
+      buf_(std::move(other.buf_)), transport_(other.transport_),
+      timeout_(std::move(other.timeout_)) {}
+
 ResponseStream::ResponseStream(Message &&other, std::weak_ptr<Tcp::Peer> peer,
                                Tcp::Transport *transport, Timeout timeout,
                                size_t streamSize)
-    : Message(std::move(other)), peer_(std::move(peer)), buf_(streamSize),
+    : response_(std::move(other)), peer_(std::move(peer)), buf_(streamSize),
       transport_(transport), timeout_(std::move(timeout)) {
-  if (!writeStatusLine(version_, code_, buf_))
+  if (!writeStatusLine(response_.version(), response_.code(), buf_))
     throw Error("Response exceeded buffer size");
 
-  if (!writeCookies(cookies_, buf_)) {
+  if (!writeCookies(response_.cookies(), buf_)) {
     throw Error("Response exceeded buffer size");
   }
 
-  if (writeHeaders(headers_, buf_)) {
+  if (writeHeaders(response_.headers(), buf_)) {
     std::ostream os(&buf_);
     /* @Todo @Major:
      * Correctly handle non-keep alive requests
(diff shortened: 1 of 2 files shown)


src/common/http.cc
```diff
@@ -595,6 +600,31 @@ ResponseStream::ResponseStream(Message &&other, std::weak_ptr<Tcp::Peer> peer,
...
+std::streamsize ResponseStream::write(const char *data, std::streamsize sz) {
+  std::ostream os(&buf_);
+  os << std::hex << sz << crlf;
+  os.write(data, sz);
+  os << crlf;
+  return sz;
+}
+
+std::shared_ptr<Tcp::Peer> ResponseStream::peer() const {
+  if (peer_.expired())
```
dennisjenkins75: Please put braces on all single-line blocks inside 'if', 'for'  and 'while' constructs.  I intend to enforce this rule via a linter someday.

src/common/http.cc
```diff
@@ -617,6 +647,109 @@ void ResponseStream::ends() {
...
+  peer_ = std::move(other.peer_);
+  transport_ = other.transport_;
+  buf_ = std::move(other.buf_);
+  timeout_ = std::move(other.timeout_);
+  return *this;
+}
+
+void ResponseWriter::setMime(const Mime::MediaType &mime) {
+  auto ct = response_.headers().tryGet<Header::ContentType>();
+  if (ct)
```
dennisjenkins75: braces

src/common/http.cc
```diff
@@ -617,6 +647,109 @@ void ResponseStream::ends() {
...
+}
+
+Async::Promise<ssize_t> ResponseWriter::sendImpl(Code code, const char *data,
+                                                 const size_t size,
+                                                 const Mime::MediaType &mime) {
+  response_.code_ = code;
+
+  if (mime.isValid()) {
+    auto contentType = headers().tryGet<Header::ContentType>();
+    if (contentType)
```
dennisjenkins75: braces

src/common/http.cc
```diff
@@ -617,6 +647,109 @@ void ResponseStream::ends() {
...
+const Header::Collection &ResponseWriter::headers() const {
+  return response_.headers();
+}
+
+Header::Collection &ResponseWriter::headers() { return response_.headers(); }
+
+Timeout &ResponseWriter::timeout() { return timeout_; }
+
+std::shared_ptr<Tcp::Peer> ResponseWriter::peer() const {
+  if (peer_.expired())
```
dennisjenkins75: braces

src/common/http.cc
hyperxor: @dennisjenkins75 I executed `clang-format -i src/common/http.cc` before commit. Ok, I will fix it.

hyperxor:
@dennisjenkins75 No no, clang-format didn't remove them, they were initially absent, you can look at current `http.h` file (before my change).

hyperxor:
@dennisjenkins75 Hello, yes, this change broke code compatibility. Maybe I can propose adding `body` method to `ResponseWriter` for support backward compatibility? On the other hand, `ResponseWriter` is a class about writing `Response`, client shouldn't have an access to internal data, it's a question of encapsulation. Good API should force client to use it to build response usign high level constructions (just an example `Pistache::Http::Code::Ok`), for instance:
```
response.send(Pistache::Http::Code::Ok, "Hello World\n");
```
Am I right? Maybe you can be stricter and more meticulous to my changes that I propose, especially in refactoring. I don't like to broke API. Now I'm trying to fix the issue #610 and I will be more careful with changes, particular in the question of code compatibility.

hyperxor:
@dennisjenkins75 Your pull request looks good, everything is clear for me. I only made a comment about moving body of `getResponseSize` and `getResponseCode` methods to cpp file.
It's up to you, but in general I prefer to put implementation to cpp when it's possible. 
I will think about logging mechanism after fixing #610 . In my opinion the architecture of this new logging system should be designed in a good way.

hyperxor:
@dennisjenkins75 I’m not sure I agree with you on that. Let me describe my opinition. As I have already said before I'm not so experienced developer and can bo wrong. I will be happy if you explain that I'm wrong.
First of all, in my opinion, headers should describe API and it would be better for client not to see the implementation. In some cases client can rely on this implementation and it's not so good in the case when you decide to change implementation. Good interface should provide clear way to manipulate and change the state of some object. Also, modern compilers are so smart in optimization side and can inline a method body without `inline` specifier, not inline it with `inline`. specifier, for instance, discussion [here](https://stackoverflow.com/a/8450069).

hyperxor:
@dennisjenkins75 I will research more deeply the topic about putting method body to header or cpp from perfomance side. Maybe I'm wrong.

PR #698, opened 2020-02-08
Code simplification in Peer class and small fix in Listener

hyperxor:
1) Code simplification in `Listener` and `Peer`: in my opiniton it would be better create `Peer` only using factory method and not allow user to associate some date (for instance, `fd`) later;
2) Add `fromUnix` overloading to `Address` to improve code in `Listener`;
3) Add `client_fd`socket  closing in the case when memory for SSL cannot be allocated.

### DIFF
--- a/src/common/net.cc
+++ b/src/common/net.cc
@@ -248,6 +248,10 @@ Address Address::fromUnix(struct sockaddr *addr) {
   throw Error("Not an IP socket");
 }
 
+Address Address::fromUnix(struct sockaddr_in *addr) {
+  return Address::fromUnix((struct sockaddr *)&addr);
+}
+
 std::string Address::host() const { return ip_.toString(); }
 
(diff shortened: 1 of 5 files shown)


src/common/net.cc
```diff
@@ -248,6 +248,10 @@ Address Address::fromUnix(struct sockaddr *addr) {
   throw Error("Not an IP socket");
 }
 
+Address Address::fromUnix(struct sockaddr_in *addr) {
+  return Address::fromUnix((struct sockaddr *)&addr);
```
dennisjenkins75: Please use std::static_cast<> (or whichever cast is approprioate), but not the legacy C style casting.  cppcheck will flag this, as will "-Weffc++".

hyperxor:
@dennisjenkins75 It was a wrong cast. Sorry for this mistake, I have fixed it.

PR #699, opened 2020-02-09
Improve Peer SSL API and remove unnecessary code

hyperxor:
Hello,

I this pull request I'm doing two things:
1) Improved `Peer` API in case of SSL;
2) Delete `associateSSL` call because I consider it useless, `SSL_free` will be called in `Peer` destructor.

What do you think about these changes?

### DIFF
--- a/include/pistache/peer.h
+++ b/include/pistache/peer.h
@@ -39,13 +39,13 @@ public:
   ~Peer();
 
   static std::shared_ptr<Peer> Create(Fd fd, const Address &addr);
+  static std::shared_ptr<Peer> CreateSSL(Fd fd, const Address &addr, void *ssl);
 
   const Address &address() const;
   const std::string &hostname();
   Fd fd() const;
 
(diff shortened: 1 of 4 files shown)


hyperxor:
@dennisjenkins75 I have seen these problems with raw pointers, I will definitely think about them and hope that can propose good solution. I'm trying to dive into these codebase, so interesting.

kiplingw:
Just be careful with the deleters. With `unique_ptr` they are part of the type and not an attribute. OpenSSL's API is a bit of a mess and you need to verify what the appropriate destructor function does on a null pointer. For some of its functions behaviour is undefined or they require valid pointers. Others are fine with a nullptr. If the former then you need to be careful to provide a lambda for the deleter that validates the pointer before the API tries to clean it up.

hyperxor:
@dennisjenkins75 Sorry for my mistake, really apologise, I have fixed it. I will think about better solution later and  be careful checking more build condiitons.

PR #705, opened 2020-02-13
Code simplification in ResponseWriter

hyperxor:


hyperxor:
@dennisjenkins75 Done!

kiplingw:
> Odd. The "arm64" "threadname" test failed. I'm not sure why. This test did not fail on the previous PR.

One of the benefits in having unit testing on multiple architectures is you're more likely to catch the false positive successful builds that should not have passed. Minor differences in thread schedulers and a myriad of other nuances on the platform end up inevitably revealing some issue that actually is a problem on all the architectures, but one of them happened to get lucky and catch it.

This problem seems to have also [occurred](https://launchpadlibrarian.net/464875960/buildlog_ubuntu-eoan-i386.pistache_0.0.002-202002132232~ubuntu19.10.1_BUILDING.txt.gz) on i386 recently.

kiplingw:
Yup. That's exactly what it is. Race conditions. That's the great thing about testing on architectures that you have no intention of actually using in practise. I have a POWER 9 machine sitting beside me with a 144 logical cores. I would never use it for personal use, but I've caught so many issues in code with it.

PR #707, opened 2020-02-15
Add version of bind() that can bind to shared_ptrs

christophebedard:
`std::bind()` can handle `std::shared_ptr`s just fine, so at first I expected `Routes::bind()` to handle them as well. However, it does not, since it simply doesn't actually use `std::bind()`.

### DIFF
--- a/include/pistache/route_bind.h
+++ b/include/pistache/route_bind.h
@@ -49,6 +49,19 @@ Route::Handler bind(Result (Cls::*func)(Args...), Obj obj) {
 #undef CALL_MEMBER_FN
 }
 
+template <typename Result, typename Cls, typename... Args, typename Obj>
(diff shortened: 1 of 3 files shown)


christophebedard:
Sure, should I add another file that exclusively tests this?

christophebedard:
> You can, or you can add new test functions to the existing source that contains tests of the same classes that you are modifying.

Alright, I'll look into it.

> My only other concern is the C-style macro. If we ever compile with "-Weffc++", it will complain about those. How do you feel about expanding the macro, since its only used once in each bind method?

Done. I've expanded both.

christophebedard:
I've added the (simple) test. Sorry for the force pushes, I kept messing up my commits :sweat_smile:

christophebedard:
I rebased on master.

PR #709, opened 2020-02-16
ResponseWriter caputres response size (bytes) for use after send().

dennisjenkins75:
1. Add method ResponseWriter::getResponseSize() (returns size_t)
1. Add method ResponseWriter::getResponseCode() (returns Http::Code)
1. Added unit test of the same.
1. Formatted the 3 modified files w/ clang-format

This addresses an issue raised in https://github.com/oktal/pistache/pull/695

### DIFF
--- a/include/pistache/http.h
+++ b/include/pistache/http.h
@@ -402,6 +402,18 @@ public:
 
   std::shared_ptr<Tcp::Peer> peer() const;
 
+  // Expose the response size and code so that user-supplied logger can
+  // access them.  Perhaps we should fully expose the response_ instead?
+  //     const Response& getResponse() const { return respose_; }
+
+  size_t getResponseSize() const {
+    return response_.body().size();
+  }
+
+  Code getResponseCode() const {
+    return response_.code();
+  }
+
   // Unsafe API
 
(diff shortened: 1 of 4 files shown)


include/pistache/http.h
```diff
@@ -402,6 +402,13 @@ class ResponseWriter final {
 
   std::shared_ptr<Tcp::Peer> peer() const;
 
+  // Returns total count of HTTP bytes (headers, cookies, body) written when
+  // sending the response.  Result valid AFTER ResponseWriter.send() is called.
+  ssize_t getResponseSize() const { return sent_bytes_; }
```
hyperxor: I think it would be better to move definitions of `getResponseSize` and `getResponseCode` methods to cpp file.

dennisjenkins75: I'm not 100% sure that moving the function bodies out of the headers is the right thing to do, for all of pistache, but that discussion is beyond the scope if this PR.  These are nearly-trivial getters, so I'll leave them inline for now.  You can move them if you wish, but I have concerns about the performance impact of the mass migration of these methods from .h into .cc files (it prevents the compiler from optimizing them at the call site).

PR #711, opened 2020-02-18
Fix issue #610 and refactoring

hyperxor:
Hello,

The main purpose of this PR is fixing #610 issue. I introduced gettes and setters for request and response sizes in `Http::Handler` and also use them in Parsers.
I decided to remove the following statements:
```
  ArrayStreamBuf<char>::maxSize = options.maxRequestSize_;
  DynamicStreamBuf::maxSize = options.maxResponseSize_;
```
from `Endpoint::init`.  In my opinion `Endpoint` shouldn't know about data structures that are used for parsing implementation and shouldn't participate in setting `maxSize` values because it becomes global. I moved `maxSize` from static to class member variable. Also I rewrote `handleIncoming` method in `Client` to handle situation when the server's response size is more than `Const::MaxBuffer` bytes. Several unit-tests has been added to check changes.


hyperxor:
@dennisjenkins75 Hello, it would be better to check this PR in a more meticulous way.

hyperxor:
@dennisjenkins75 I'm not hurry at all and I mentioned you because you can check more cases.
I'm ready to discuss all things around this PR.

Ok, let's talk about overriding the max request and response size. I implemented API that offers you opportunity to change request and response size in a way when you don't touch basic data structures. In my opinion `maxSize` shouldn't be static because when you set this value in `Endpoint::init` it becomes global, even for `Client` part, from my point of view, it's not so good.

hyperxor:
> You are making some significant API design changes. If there a specific philosophy or design goal that you have or are following?

In general or in this PR?

hyperxor:
@dennisjenkins75 tests are failed - #712 there is a problem with certfificates

hyperxor:
> if there a specific philosophy or design goal that you have or are following?
@dennisjenkins75  SOLID, Separation of Concern and try to write modern and clean C++ code.

hyperxor:
@dennisjenkins75 Excellent! Thank you so much, I have already merged your change in my branch and push it.

hyperxor:
@dennisjenkins75 I have noticed that `http_server_test` is a flaky test:
```
12: [       OK ] http_server_test.client_multiple_requests_disconnection_on_timeout_from_single_threaded_server (6003 ms)

12: [ RUN      ] http_server_test.multiple_client_with_requests_to_multithreaded_server

12/24 Test #12: http_server_test .................***Exception: SegFault 12.14 sec

test 13
```

Example:
https://travis-ci.org/oktal/pistache/jobs/652330479
https://travis-ci.org/oktal/pistache/jobs/652335415

Issue #158, opened 2017-11-05
RUSAGE_THREAD not declared


MonkeySlut:
Hi :)
During 'make' of pistache in docker image I get:
> 'RUSAGE_THREAD' was not declared in this scope
> auto res = getrusage(RUSAGE_THREAD, &now);

I would have used pistache-docker image but it's out of date and has a critical bug for me.
I tried using:

> #define __USE_GNU
> #include <sys/resource.h>

without any luck, anyone can bring some insights?

xoac:
You need to deliver more information about docker image. It's looks like a version of your libs don't have defined `RUSAGE_THREAD`.

MonkeySlut:
Apparently, this was a problem with the docker image, seems strange because it's an ubuntu image...
I instead used alpine docker image, works fine.
Closing issue :)

PR #716, opened 2020-02-23
Exploring flaky tests: more logs and several improvements

hyperxor:
1) Added additional check to `Listener::bind` to handle the case when socket creation is failed
2) Added more logs to debug flaky `http_server_test.multiple_client_with_requests_to_multithreaded_server` test
3) Improved https test and fixed warning

### DIFF
--- a/src/server/listener.cc
+++ b/src/server/listener.cc
@@ -170,10 +170,6 @@ void Listener::bind(const Address &address) {
     throw std::runtime_error(strerror(errno));
   }
 
-  if (fd == -1) {
-    throw SocketError(strerror(errno));
-  }
-
   make_non_blocking(fd);
   poller.addFd(fd, Flags<Polling::NotifyOn>(Polling::NotifyOn::Read),
                Polling::Tag(fd));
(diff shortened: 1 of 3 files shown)


src/server/listener.cc
```diff
@@ -170,6 +170,10 @@ void Listener::bind(const Address &address) {
     throw std::runtime_error(strerror(errno));
   }
 
+  if (fd == -1) {
```
dennisjenkins75: This should not be needed, but I do see a bug above.  On line #158, after "close(fd)", fd should be set back to -1.  Because this might be on the last iteration through the for loop, so the loop would exit with fd>=0, but unbound.  Your code check here will not catch that.  When the for loop exits, either "addr" is not null (and fd is bound, listening and >=0), or addr is null and fd should be ignored.

hyperxor: @dennisjenkins75 You are totally right, my check is useless in this case. On the other hand I don't see any problems on line 158. If it is the last iteration then `addr` will be `nullptr` and the exception will be thrown and it doesn't matter what's the value local variable `fd` has.

PR #722, opened 2020-02-28
Code improvements in Address class and more logs

hyperxor:
1) In these changes I'm trying to improve code in `Address::init` and remove redundant executions;
2) More logs to `http_server_test`.

### DIFF
--- a/src/common/net.cc
+++ b/src/common/net.cc
@@ -20,24 +20,41 @@
 namespace Pistache {
 
 namespace {
-bool IsIPv4HostName(const std::string &host) {
+IP GetIPv4(const std::string &host) {
   in_addr addr;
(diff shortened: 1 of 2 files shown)


miketsts:
This commit uses `static_assert` which belongs to C++ 17. Before, the code has been compiled with C++ 14 without a problem.

Is it a project decision to have C++ 17 as a minimal requirement?

hyperxor:
@dennisjenkins75 @miketsts Hello, I have made a [fix](https://en.cppreference.com/w/cpp/language/static_assert). Fortunately, there is a better solution rather than `#ifdef` use.

hyperxor:
@dennisjenkins75 Oh, never mind, everything is really cool and I like when you give me an advice about my code. In this case it was my mistake, used wrong `static_assert` implementation.

PR #725, opened 2020-03-02
Implemented support for Authorization headers using basic method...

### LINKED ISSUE #438: How to access (basic auth) authorization header?

kiplingw:
{src/common/http_header.cc,include/pistache/http_header.h}: Implemented support for Authorization headers using basic method...
include/pistache/base64.{cc,h}: Added utility classes and helper functions for base 64 encoding and decoding...
tests/headers_test.cc: Added unit test authorization_basic_test...
version.txt: Bumped versioning metadata...

Closes #438 after passing code review.


kiplingw:
> Please format these files with "clang-format" first.

I've never used it, but I've just installed it. Can you please provide me with the complete CLI you'd like me to run?

kiplingw:
Ok, try that. ;)

kiplingw:
@dennisjenkins75, correct me if wrong, but there can't be a DoS attack via a massive header not being able to be decoded since the reactor already limits the maximum client request size?

miketsts:
This commit uses `std::byte` which belongs to C++ 17. Before, the code has been compiled with C++ 14 without errors (warnings about `static_assert`, also from C++ 17 a couple of commits before).

Is it a project decision to have C++ 17 as a minimal requirement?
I compile my project with gcc 5.4 which is default in Ubuntu 16.04 and supports C++ 14.

ivan1993br:
> Why did we drop a c++14-only compiler from travis? That would have caught our use of c++17 (well, if the compiler strictly enforced the c++14 standard).

Hi, yes... now it is C++17. This was the first PR that introduce the c++17 [PR#722](https://github.com/oktal/pistache/pull/722), with the `static_assert ( bool_constexpr )`. If you agree I can create a PR to bring the current modifications to c++14, also I would suggest to set the CMakeLists to c++17, and when you decide to support c++17, modify. What do you think @dennisjenkins75?

Thanks

PR #740, opened 2020-03-22
SSL usage improvements: using RAII for ssl context resource

hyperxor:
1) Added using RAII for working with `SSL_CTX` and cleaning resources automatically.
2) Minor code improvements.
3) It's a good time to think about using `EVP_cleanup();`. Is it required? (see my comment with https://www.openssl.org/news/changelog.txt link)

### DIFF
--- a/src/server/listener.cc
+++ b/src/server/listener.cc
@@ -8,6 +8,7 @@
 #include <pistache/listener.h>
 #include <pistache/os.h>
 #include <pistache/peer.h>
+#include <pistache/ssl_wrappers.h>
 #include <pistache/transport.h>
 
 #include <arpa/inet.h>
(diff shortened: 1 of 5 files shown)


src/server/listener.cc
```diff
@@ -370,23 +365,22 @@ void Listener::dispatchPeer(const std::shared_ptr<Peer> &peer) {
...
+  ssl::SSLCtxPtr ctx{SSL_CTX_new(method)};
+  if (ctx == nullptr) {
     ERR_print_errors_fp(stderr);
     throw std::runtime_error("Cannot setup SSL context");
   }
 
   if (!use_compression) {
     /* Disable compression to prevent BREACH and CRIME vulnerabilities. */
-    if (!SSL_CTX_set_options(ctx, SSL_OP_NO_COMPRESSION)) {
+    if (!SSL_CTX_set_options((SSL_CTX *)ctx.get(), SSL_OP_NO_COMPRESSION)) {
```
dennisjenkins75: Please do not use legacy C style casts; they will show up as warnings in cppcheck.  Since this one gets casted alot, consider writing a "static inline SSL_CTX* get_context(void *foo)..." that hides the cast in one place (and use a C++ style cast instead).

hyperxor: @dennisjenkins75 Thank you for this remark, I totally agree with you.

Issue #748, opened 2020-04-03
Remaining -Wconversion warnings 


miketsts:
Hi,

After my PR introducing -Wconversion compilation flag a couple of weeks ago, there still remain a couple of warnings that I don't know how to resolve:

```[ 65%] Building CXX object src/CMakeFiles/pistache.dir/common/http_defs.cc.o
In file included from /home/user/pistache/src/common/http_defs.cc:11:0:
/home/user/pistache/include/pistache/date.h: In constructor ‘constexpr date::weekday_indexed::weekday_indexed(const date::weekday&, unsigned int)’:
/home/user/pistache/include/pistache/date.h:1641:11: warning: conversion to ‘unsigned char:4’ from ‘unsigned char’ may alter its value [-Wconversion]
     : wd_(static_cast<decltype(wd_)>(static_cast<unsigned>(wd))),
           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/user/pistache/include/pistache/date.h:1642:14: warning: conversion to ‘unsigned char:4’ from ‘unsigned char’ may alter its value [-Wconversion]
       index_(static_cast<decltype(index_)>(index)) {}
              ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
This is conversion warning from "unsigned char" to bitfield 'unsigned char:4'. There is already `static_cast<decltype(index_)>` - apparently this doesn't work. It looks like the following fixes these warnings:
```
@@ -1638,8 +1638,8 @@ inline bool weekday_indexed::ok() const NOEXCEPT {
 CONSTCD11
 inline weekday_indexed::weekday_indexed(const date::weekday &wd,
                                         unsigned index) NOEXCEPT
-    : wd_(static_cast<decltype(wd_)>(static_cast<unsigned>(wd))),
-      index_(static_cast<decltype(index_)>(index)) {}
+    : wd_(static_cast<unsigned>(wd) & 0xF),
+      index_(index & 0xF) {}
```
Any comments?

```
[ 69%] Building CXX object src/CMakeFiles/pistache.dir/common/http.cc.o
In file included from /home/user/pistache/include/pistache/http.h:20:0,
                 from /home/user/pistache/src/common/http.cc:8:
[...]

PR #754, opened 2020-04-12
Fix -wConversion warnings #748

miketsts:
Fix warnings occurring in gcc, described in #748


miketsts:
I'm terribly sorry - it's the downside of working late in the night :-(
I've amended the commit now.

Issue #708, opened 2020-02-15
https_server_test (Timeout) fails


kiplingw:
This appears to be a race condition that periodically pops up on some pbuilder builds for the PPA. Specifically the `https_server_test` unit test fails due to a timeout. This happened recently on the i386 build. The build log is available [here](https://launchpad.net/~pistache+team/+archive/ubuntu/unstable/+build/18709377).

dennisjenkins75:
Strangely, the https_server_test also randomly times out when ran on my Gentoo Linux box for a clang build, but never for gcc build.  I've not attempted to debug it though.

kiplingw:
Yeah. Let's keep this issue open as a placeholder for eventually figuring out what the issue is.

dennisjenkins75:
I take back what I posted above.  I can get the "https_server_test" to timeout on both gcc and clang builds. 

The lockup happens much more frequently on the clang build than the gcc.  However, the gcc build seems to lock up more frequently when there is external cpu contention 9for example, compiling something unrelated on the side).

The clang build seems to lock up about ever 3rd run of the test.

```
$ clang++ --version
clang version 9.0.1 
$ g++ --version
g++ (Gentoo 9.2.0-r3 p4) 9.2.0

$ mkdir build
$ cd build
$ CC=clang CXX=clang++ cmake -G "Unix Makefiles" \
    -DCMAKE_BUILD_TYPE=Debug \
    -DPISTACHE_BUILD_EXAMPLES=true \
    -DPISTACHE_BUILD_TESTS=true \
    -DPISTACHE_ENABLE_NETWORK_TESTS=true \
    -DPISTACHE_USE_SSL=true \
    .. 
$ make
$ cd build/tests
$ while ./run_https_server_test ; do date; done
```

dennisjenkins75:
kip, can you try to repro the above?  I'm a bit tired right now, but soon I'll try to get a stack dump for all threads when it is locked up.

kiplingw:
Hey Dennis. It looks like the trailing `server.shutdown()` in the `basic_tls_request_with_auth_client_cert_not_signed` is hanging. The stack trace leads to L81 of the Listener destructor waiting on `acceptThread.join()`. For whatever reason the `acceptThread` never returns control.

win32asm:
my c5 on the issue: it happens more frequently on SuSE testbed, rather than on Fedora/Cent8
[...]

PR #761, opened 2020-05-01
Fix shutdown data race in listner.cc

chrisvroberts:
A race exists between the runThreaded and shutdown/dtor as follows:

 runThreaded - called
 shutdown - called
 shutdown - check for shutdownFd.isBound returns false (so no notify)
 runThreaded - thread executing 'run' starts
 runThreaded - thread calls shutdownFd.bind
 shutdown - blocks waiting for run thread
 runThreaded - continues to run as no notification received

This deadlock avoided by ensuring the shutdownFd.bind is called before
the thread is created rather than doing it at the start of the thread's
execution.

### DIFF
--- a/src/server/listener.cc
+++ b/src/server/listener.cc
@@ -204,7 +204,8 @@ Port Listener::getPort() const {
 }
 
 void Listener::run() {
-  shutdownFd.bind(poller);
+  if (!shutdownFd.isBound())
+    shutdownFd.bind(poller);
   reactor_.run();
 
   for (;;) {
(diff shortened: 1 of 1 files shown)


chrisvroberts:
Note: The issue this change resolves was seen in a unit test where the server was run up and shutdown very quickly. This is very unlikely to be seen when running a pistache server in the wild. Note also that running under valgrind with helgrind other potential data race warnings are reported but I have not looked into these in any detail. It may be worth adding a helgrind run to the CI.

kiplingw:
Thanks @chrisvroberts. Indeed, the first thing that came to my mind was #708. @dennisjenkins75, after you merge this I'll look further into if we can close that long outstanding issue as well.

Issue #762, opened 2020-05-09
Linking error on Raspberry Pi 4


sineang01:
Hi,

I was trying to use compile pistache on Raspberry Pi 4 Model B and I found out that it fails when it is compiling a test:

`
[ 93%] Linking CXX executable run_request_size_test
/usr/bin/ld: ../src/libpistache.a(client.cc.o): in function 'Pistache::Http::Client::doRequest(Pistache::Http::Request)':
client.cc:(.text+0x6aec): undefined reference to '__atomic_fetch_add_8'
collect2: error: ld returned 1 exit status
make[2]: *** [tests/CMakeFiles/run_threadname_test.dir/build.make:88: tests/run_threadname_test] Error 1
make[1]: *** [CMakeFiles/Makefile2:1688: tests/CMakeFiles/run_threadname_test.dir/all] Error 2
/usr/bin/ld: ../src/libpistache.a(client.cc.o): in function 'Pistache::Http::Client::doRequest(Pistache::Http::Request)':
client.cc:(.text+0x6aec): undefined reference to '__atomic_fetch_add_8'
collect2: error: ld returned 1 exit status
make[2]: *** [tests/CMakeFiles/run_request_size_test.dir/build.make:88: tests/run_request_size_test] Error 1
make[1]: *** [CMakeFiles/Makefile2:2585: tests/CMakeFiles/run_request_size_test.dir/all] Error 2
[ 94%] Linking CXX executable run_headers_test
`

I think the pistache's CMakeLists.txt is not linking the `atomic` library.
I patched the file to include the missing call and it does work fine. If you are interested in the patch I can start a pull request.

Thanks!
Simone

dennisjenkins75:
All PRs are welcome.  I'll need to verify that they pass testing on x86 though, under both clang and gcc.  Our travis config appears broken (in two ways) that I've not been able to fix yet, so I'll have to test your PR manually on my own dev box.

sineang01:
Hi @dennisjenkins75,

thanks for your answer. I looked into the TravisCI issue (as I noticed it too while I was fixing this raspberrypi thing) and I found a possibly fix.
I started a pull request for that: https://github.com/oktal/pistache/pull/763
Once that goes in, if you are happy with that, then I will start another one for raspberry.

//S

dennisjenkins75:
If Travis supports testing a build on a Pi, then great, please add support for it.

PR #763, opened 2020-05-11
Fixing broken TravisCI builds for the clang toolchain

sineang01:
I saw your comment on #762 regarding failing Travis builds and I tried to solve it.
Seems like the issue is that you were loading a `trusty` clang pipeline, which is for Ubuntu 14.04 LTS but at the top of the travis.yml you specify `bionic` which is Ubuntu 18.04 LTS, so I fixed the clang pipeline to use `bionic` too.

This issue seems the same that a guy was reporting here: https://travis-ci.community/t/cannot-apt-get-install-clang-5-0/3250/3
That time (1 year ago) the suggestion was to switch from `trusty` to `xenial` (Ubuntu 16.04 LTS) or, in alternative, to add `dist: trusty` at the top of the file to keep using `trusty` but, since you are already using `bionic`, this suggest didn't seem to fit in your build system.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ matrix:
       - COV_TOOL_ARGS=gcov
     addons:
       apt:
-        sources: ['ubuntu-toolchain-r-test', 'llvm-toolchain-trusty-4.0']
+        sources: ['ubuntu-toolchain-r-test', 'llvm-toolchain-bionic-4.0']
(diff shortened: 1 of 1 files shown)


sineang01:
Here you can see it working: https://travis-ci.org/github/sineang01/pistache/builds/685787442

sineang01:
Here you can see it working: https://travis-ci.org/github/sineang01/pistache/builds/685787442

PR #766, opened 2020-05-12
Remove incorrect information about external dependencies from website

johschmitz:


kiplingw:
Actually even I've been trying to get it updated for over a year now. I've emailed Mathieu but no response.

PR #772, opened 2020-05-26
Allow Pistache::Listener to load and serve certificate chains

mplellison:
Extended HTTPS certificate generation to use an openssl x509_v3_config file.
Swapped SSL_CTX_use_certificate_file() for SSL_CTX_use_certificate_chain_file() in listener.cc.

It's worth considering how to make this SSL API more flexible - possibly by exposing the SSL_CTX, or letting the user pass in a SSL_CTX for the server to use.


mplellison:
The above force-push fixed a broken test - I'd used the wrong certificate file in the new http_client_test.basic_tls_request_with_chained_server_cert test.

Issue #771, opened 2020-05-26
listener.cc logs errors directly to stderr


mplellison:
When the listener object calls OpenSSL library functions during setup or connection handling, errors and validation failures are logged using ERR_print_errors_fp(stderr) separately from any exceptions that might be thrown. This means that these errors are always logged, and are logged without any context.

This could be solved for errors during setup by writing the OpenSSL errors into the exception message (for example), but this won't work as well for SSL validation failures since these don't throw exceptions. That might involve adding another handler somewhere for the user to implement and choose what to log, or implementing some kind of log handler that can be used elsewhere as well.

mplellison:
Having had a closer grep of the code, it looks like listener.cc also logs messages to stderr if a SocketError or ServerError is thrown while calling handleNewConnection() from run(). I've changed the title to reflect this.

A relatively simple solution that would also help going forward would be to optionally compile against a third-party logging library (eg log4cxx), defining Pistache-specific macros for the actual calls to the library.

dennisjenkins75:
Are you going to post a PR with your code changes?

mplellison:
It depends which way we go with this - adding logging is a pretty significant API change and it'd be nice to get it right. I'm hesitant to just go and compile in log4cxx because it's an additional dependency (one that hasn't had a release in several years...), and making it optional would add more things that need support and probably double the build and test time.

What do you think?

dennisjenkins75:
I think that we should avoid adding additional dependencies.  Honestly, I've always found pistache's error handling and logging to be sub-par, but it has not bothered my own use-case enough to fix it.

Off the top of my head, I would propose that pistache should log error logs (not HTTP request logs) through one method, and that method can call back to a user-supplied function so the user can control what to do with the log data.  HTTP logs should be handled separately, but similarly.

Sadly, I don't have time to code this up, and my own use of pistache already handles HTTP logs via an entry/exit macro on each request handler.

mplellison:
I've opened PR #775 with a candidate logging structure - I'm not expecting it to be merged right away though, since it really should have some input from other directions.

Issue #24, opened 2016-10-28
Featurerequest: Add a 'precondition' to Router.


bamkrs:
While using Pistache, I often use the same common code in every Handler for every route.
Something like (Attention: Mostly pseudocode, but you get the idea)

```
response.headers().add(/*Contenttype*/);
ownObject* oo = new ownObject();
if(oo->isAlive())
{
  // Actual Work for Handler
  // Send and Return if successfull
}
oo->foo = 1;
oo->bar = "Service Unavailable";
response.send(Http::Code::Service_Unavailable,oo->dump());
delete oo;
```

Wouldn't it be nice to add a function to the Router that gets called before the actual handler kicks in?

```
Router->addPreworker(/*ptrToFunction*/)
```

Preworkerfunction:

```
void preFunc(Router, request, response){     //or only request, response
  //do stuff
  if(!router->for(request)->callHandler()) {   //and do something like request->callHandler()
    //handler would be bool instead of void and return true when everything went fine
    //and false if a common error occurs

    //do errorreport-generation
    response.send(Http::Code::Service_Unavailable,errorreport);
  }
}
```

Wouldn't be this a nice pattern to reduce code and points of failure?
Maybe I implement this by myself when I find some time.

knadh:
+1

oktal:
Hello,

Yes, that indeed sounds like a great suggestion. Basically, we need *filters*.

We could have two kind of filters: *before-filters* and *after-filters*. `before-filters` will be executed before each requests and can read/write the request/response. after-filters will be executed after each requests and can also read and modify the response.

Now we need to come up with a good API for this, feel free to suggest something if you have any idea :)

knadh:
@oktal How Go does its HTTP middlewares may be a good idea here. They're just simple function wrappers, without the explicit need for having pre/post filter APIs.

oktal:
@knadh could you provide an example on how Go is handling it ?

If you have a clear idea on how to design it as well as the API to provide, feel free to post it here so that we can discuss it

knadh:
In Go, functions that satisfy `http.Handler` (which implements the function `ServeHTTP(ResponseWriter, *Request)`) can act as middleware. A middleware takes a handler and returns a handler (similar to pistache's `(const Rest::Request& request, Http::ResponseWriter response)`). Here's some pseudocode that illustrates this.

```go
func enforcePOST(next http.Handler) http.Handler {
[...]

PR #775, opened 2020-05-29
Logging api - addresses issue #771

mplellison:
Implement a logging API using a collection of PISTACHE_LOG* macros to be used internally.

Users of pre-compiled Pistache can provide an alternative implementation of a LogHandler to direct log messages as required. Users compiling Pistache can include patches to modify the PISTACHE_LOG* macros to use an alternative logging engine if required.

This PR also includes additions to the ssl wrappers to handle OpenSSL BIO objects and some other SSL helpers.

### DIFF
--- a/include/pistache/endpoint.h
+++ b/include/pistache/endpoint.h
@@ -28,6 +28,7 @@ public:
     Options &backlog(int val);
     Options &maxRequestSize(size_t val);
     Options &maxResponseSize(size_t val);
+    Options &logger(PISTACHE_LOGGER_T logger);
 
     [[deprecated("Replaced by maxRequestSize(val)")]] Options &
     maxPayload(size_t val);
@@ -39,6 +40,7 @@ public:
     int backlog_;
     size_t maxRequestSize_;
     size_t maxResponseSize_;
+    PISTACHE_LOGGER_T logger_;
     Options();
   };
   Endpoint();
@@ -151,6 +153,7 @@ private:
   Tcp::Listener listener;
   size_t maxRequestSize_ = Const::DefaultMaxRequestSize;
   size_t maxResponseSize_ = Const::DefaultMaxResponseSize;
+  PISTACHE_LOGGER_T logger_ = PISTACHE_NULL_LOGGER;
 };
 
 template <typename Handler>
--- a/include/pistache/endpoint.h
+++ b/include/pistache/endpoint.h
@@ -28,7 +28,7 @@ public:
     Options &backlog(int val);
     Options &maxRequestSize(size_t val);
     Options &maxResponseSize(size_t val);
-    Options &logger(PISTACHE_LOGGER_T logger);
+    Options &logger(PISTACHE_STRING_LOGGER_T logger);
 
     [[deprecated("Replaced by maxRequestSize(val)")]] Options &
     maxPayload(size_t val);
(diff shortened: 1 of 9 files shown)


mplellison:
While looking at the changes required for SSL handling, I noticed a couple of things around the integration with OpenSSL - the current wrappers are a bit roundabout (unique_ptrs to void) and need casting whenever they're used.

There's also a rather odd separation of the code that's only compiled with SSL - it seems to miss out code that could act on SSL objects (eg the Listener keeps its SSLCtxPtr in the standard build), while the objects themselves have been redefined to guarantee errors if used (unique_ptr to void with a noop deleter) - is this something to do with object structure and the ABI?

mplellison:
We could certainly do something like that - though it might be better to name them based on what the user would get out of them; eg `PISTACHE_STRING_LOGGER_T` - then the type can still be used for any string-based logging (which it is anyway), and different loggers of the same type can be supplied for different logging streams (like a basic version of http stats logging). We've already got a way to identify different log messages as error/info/trace etc, so I don't want to build that into the type - seems like an excellent way to end up with things getting logged straight to stderr again.

Your http req logger suggestion sounds like it could be something like a `PISTACHE_HTTP_STATS_LOGGER_T`, which expects and passes a specific struct to the logger object rather than a string - again, a user might want to supply different instances of these loggers for eg different http endpoints, and we can make our type system flexible enough for that.

To borrow your example, we'd end up with something like

```
Options &string_logger(PISTACHE_STRING_LOGGER_T logger);
Options &http_stat_logger(PISTACHE_HTTP_STAT_LOGGER_T logger);
```

The macros would need a bit of thought to support the different logger types though - maybe just changing them to PISTACHE_LOG_STRING_* would be enough.

mplellison:
Oof... you're not joking about the hack :)

We should probably keep the logging/event API reasonably controlled; I'm not comfortable expecting the user to accept even data that could turn up with an arbitrary type, and I think just having two separate types to accept different kinds of data is straightforward enough (especially if they can be made optional).

What you're looking at seems like it could be handled by extending the request handler concept to execute a chain of operations on/with the request/response pair - something like what's been discussed in issue #24 . Logging request/response data could be handled as a step in the chain, and error handling with a separate handler that gets passed an exception_ptr containing an exception thrown by something in the chain.

mplellison:
Some more thoughts...

The idea of having a logging API is so that the library can make the user aware of events or information that they might want to be aware of but don't require any action, and that they wouldn't be able to access through another API. For example, we shouldn't need to log the contents of an exception if we expect the user to catch the exception, but we should log that a connection failed to pass SSL verification because there's no other API the user can access to get that info. If a connection gets to a user-defined handler, then we can be happy that they can get any information from that that they need.

That's not to say that the other APIs can't be improved, but that's a different problem.

mplellison:
I've pushed a change that uses "string logger" for the various types and macros; that seems like it leaves space for extensions without locking down the types to only be used for certain events.

mplellison:
I think it makes more sense to leave the namespace as "Pistache::Log", since that leaves more room for extending to different concepts of logging while staying related to the existing code and not filling up the Pistache namespace (and the class names already refer to a StringLogger). Similarly, I'd lean slightly towards leaving the files as "log.{h,cc}", but I don't mind so much about that.

Happy to change them if you're sure about it though.

Issue #720, opened 2020-02-27
No crashes in /var/crash after SEGFAULT on CI


hyperxor:
Hello,

I have noticed that there are no crashes in `/var/crash` after Segmentation fault in unit tests. For instance:
1) https://travis-ci.org/oktal/pistache/jobs/655471456
At the end of log I see the following output:
```
...
$ ls -lta /var/crash

total 8

drwxr-xr-x 14 root root 4096 Feb 24 20:58 ..

drwxrwxrwt  2 root root 4096 Feb 18 22:04 .
```
2) https://travis-ci.org/oktal/pistache/jobs/654598559
```
...
$ ls -lta /var/crash

ls: cannot access '/var/crash': No such file or directory
```
Is it possible to get crash logs on CI ?

dennisjenkins75:
Interesting.  I never noticed that at the bottom before.  I wonder if Travis is changing or setting the "ulimit -c" on the container in a way that is different from when that functionality was first introduced.

Consider:
https://github.com/springmeyer/travis-coredump/blob/master/.travis.yml

dennisjenkins75:
Interesting.  I never noticed that at the bottom before.  I wonder if Travis is changing or setting the "ulimit -c" on the container in a way that is different from when that functionality was first introduced.

Consider:
https://github.com/springmeyer/travis-coredump/blob/master/.travis.yml

hyperxor:
@dennisjenkins75 Hello, Travis CI doesn't run anymore?

dennisjenkins75:
Strange.  [1]

Shows that its been running, as recently as an hour ago, but I do not always see the results on the github page for each PR.  It ran for PR #723, but skipped running for PR #735.  Most of the recent travis runs have failed; either the build was legitimately broken, or "apt-get" failed and travis did not retry it.  I'm not a travis expert; not sure why its acting weird.  But I need to pay attention to if the travis run passes or not and I should stop approving PRs unless I get a clear "green" signal from travis.

[1] https://travis-ci.org/github/oktal/pistache/builds?utm_medium=notification&utm_source=github_status

mplellison:
I've just had a quick look at travis (with all my zero knowledge on this CI), since the current build is segfaulting on a couple of tests and there isn't much in the way of backtraces - eg. https://travis-ci.org/github/oktal/pistache/jobs/692742761#L6052

[...]

PR #776, opened 2020-05-30
Travis fix to collect coredumps from build configuration subdirs - #720

mplellison:
Previously, Travis would only attempt to collect coredumps from the last
directory the build script was in, or the home directory. This meant
that most coredumps were missed, and debug information was lost.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -84,6 +84,9 @@ before_script:
   - ulimit -c
   - ulimit -a -S
   - ulimit -a -H
+    # Make sure that crash reports are generated for all executables.
+  - mkdir -p ~/.config/apport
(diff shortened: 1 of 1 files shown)


mplellison:
To summarise:
* The non-AMD64 build environments do not support changing kernel parameters, including core_pattern, so we're stuck with the mixture of coredump locations that are set by default.
* Letting apport intercept the coredump mean that we get coredumps with useful metadata, including which executable it came from. This is essential to get useful debug symbols in a backtrace.
* Since we have multiple executables which could all crash independently, we need to load all crash data on a per-executable basis.

The first commit in this PR covers this. The second modifies the build instructions to get a more consistent and controllable build environment for different compiler versions, instead of relying on the matrix expansion provided by Travis.

PR #779, opened 2020-06-12
Add middleware support

### LINKED ISSUE #24: Featurerequest: Add a 'precondition' to Router.

jiankyu:
This commit partially resolves #24, which is a common requirement on
many RESTful services.

A vector of middleware handlers is added to Router, these handlers are
executed before the URL match and handler invocation, in the order of
registration.

A middleware function has the signature:

bool func(Pistache::Http::Request &request,
          Pistache::Http::ResponseWriter &response);

The first parameter is a reference to Http::Request, not Rest::Request,
at the moment middlewares are called, the request parameters are not yet
filled. This allows some generic operation on the request, such as
authentication or rate limiting.

Parameters are passed as references, to allow the middlewares be able to
modify them, such as injecting custom headers.

If a function returns true, the next function is executed, until all
middleware functions are called and all of them return true, the
registered handler is called.

If one of them returns false, this means there is no need to further
call the subsequent functions and the handler, the handling of this
request completes. Typical case is missing auth info or reaching limit.
Basically, the middleware function that returns false is responsible to
call response.send() to send the response to client.

Like Route::bind, a new method, Route::middleware(), with different
signatures, is added, to convert different callables into a middleware.

There is also requirements on the "post-handling" handlers, such as
logging response, measuring handler latency, and so on, in a centralized
place. This commit doesn't implement it: The response is passed to
handler using std::move, it's not retrievable again after handler
processing.

Signed-off-by: Jiankun Yu <yujiankun@hotmail.com>

### DIFF
--- a/include/pistache/router.h
+++ b/include/pistache/router.h
@@ -65,6 +65,8 @@ struct Route {
 
   typedef std::function<Result(const Request, Http::ResponseWriter)> Handler;
 
+  typedef std::function<bool(Http::Request& req, Http::ResponseWriter& resp)> Middleware;
+
   explicit Route(Route::Handler handler) : handler_(std::move(handler)) {}
 
   template <typename... Args> void invokeHandler(Args &&... args) const {
@@ -218,6 +220,7 @@ public:
   void head(const std::string &resource, Route::Handler handler);
 
   void addCustomHandler(Route::Handler handler);
+  void addMiddleware(Route::Middleware middleware);
 
   void addNotFoundHandler(Route::Handler handler);
   inline bool hasNotFoundHandler() { return notFoundHandler != nullptr; }
@@ -227,12 +230,14 @@ public:
   Route::Status route(const Http::Request &request,
                       Http::ResponseWriter response);
 
-  Router() : routes(), customHandlers(), notFoundHandler() {}
+  Router() : routes(), customHandlers(), middlewares(), notFoundHandler() {}
 
 private:
   std::unordered_map<Http::Method, SegmentTreeNode> routes;
 
   std::vector<Route::Handler> customHandlers;
+
+  std::vector<Route::Middleware> middlewares;
 
   Route::Handler notFoundHandler;
 };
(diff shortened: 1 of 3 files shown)


carloscbl:
As quick workaround i've added this code as posthandler, on top of all my routes handlers

```c++
class posthandler_raii{
    const Pistache::Rest::Request &request;
    const Pistache::Http::ResponseWriter& response;
    public:
    posthandler_raii(const Pistache::Rest::Request &request, const Pistache::Http::ResponseWriter& response):request(request),response(response){}
    ~posthandler_raii(){
        SPDLOG_DEBUG("{} {} -> {} {}", methodString(request.method()), request.resource(), response.getResponseCode(), codeString(response.getResponseCode()) );
    }
};

```

example:
```c++
void DefaultApi::apps_get_handler(const Pistache::Rest::Request &request, Pistache::Http::ResponseWriter response) {
    posthandler_raii ph(request, response);
    try {
      this->apps_get(response);
    } catch (nlohmann::detail::exception &e) {
        //send a 400 error
        response.send(Pistache::Http::Code::Bad_Request, e.what());
        return;
```
And if someone needs minimal example of prehandler

```c++
bool prehandler( Pistache::Http::Request & request, Pistache::Http::ResponseWriter& response) {
    SPDLOG_DEBUG("{} {}", methodString(request.method()) , request.resource());
    return true;
}


void DefaultApi::setupRoutes() {
    using namespace Pistache::Rest;
    router->addMiddleware(Routes::middleware(&prehandler));
    Routes::Get(*router, base + "/apps", Routes::bind(&DefaultApi::apps_get_handler, this));
```

jiankyu:
Yes, this wrapper class is a good approach, it introduces minimal changes to the existing handler methods.

PR #783, opened 2020-06-22
Update transport.cc

### LINKED ISSUE #782: Only throw without catch causes coredump
Hi,
I found that some places only throw but no catch, so when an exception is triggered, the program will coredump, for example, in pistache/src/common/transport.cc Line 75,
if (isPeerFd(tag)) {
auto &peer = getPeer(tag);
handleIncoming(peer);
}
there is no catch, but in pistache/src/common/transport.cc145 Line, threr threw an exception
 if (errno == ECONNRESET) {
          handlePeerDisconnection(peer);
        } else {
          throw std::runtime_error(strerror(errno));
        }

zhanb:
fix #782 bug

### DIFF
--- a/src/common/transport.cc
+++ b/src/common/transport.cc
@@ -139,7 +139,7 @@ void Transport::handleIncoming(const std::shared_ptr<Peer> &peer) {
           handler_->onInput(buffer, totalBytes, peer);
         }
       } else {
-          handlePeerDisconnection(peer);
+        handlePeerDisconnection(peer);
       }
       break;
     } else if (bytes == 0) {
--- a/src/common/transport.cc
+++ b/src/common/transport.cc
@@ -142,7 +142,7 @@ void Transport::handleIncoming(const std::shared_ptr<Peer> &peer) {
         if (errno == ECONNRESET) {
           handlePeerDisconnection(peer);
         } else {
-          throw std::runtime_error(strerror(errno));
+            handlePeerDisconnection(peer);
         }
       }
       break;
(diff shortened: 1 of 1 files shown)


src/common/transport.cc
```diff
@@ -142,7 +142,7 @@ void Transport::handleIncoming(const std::shared_ptr<Peer> &peer) {
         if (errno == ECONNRESET) {
```
dennisjenkins75, line 142: Assuming the peer disconnects for all other errors is probably a good idea.  Throwing this exception is not useful.

However, please clean up the "if()" statement.  Just remove lines 142-146 and simply call "handlePeerDisconnection()" here.

zhanb: I adjusted it, please accept it

src/common/transport.cc
```diff
@@ -139,11 +139,7 @@ void Transport::handleIncoming(const std::shared_ptr<Peer> &peer) {
           handler_->onInput(buffer, totalBytes, peer);
         }
       } else {
-        if (errno == ECONNRESET) {
           handlePeerDisconnection(peer);
```
dennisjenkins75, line 142: (minor nit): Please adjust indent level on "handlePeerDisconnection()", otherwise, I'll have to clean it up the next time I run "clang-tidy" on the code.

zhanb: I adjusted it ,please checkout again

KlaasVortex:
I would say that this should fix my problem. Cannot verify it, since the problem happens rarely. But we are testing it.

PR #785, opened 2020-06-24
Small improvements in Listener constructor

hyperxor:


hyperxor:
@dennisjenkins75 Looks like cppcheck is wrong in this case. it's considered as a good practice to initialize a non-static data member with a default value where they are declared to remove code duplication in constructors (https://www.stroustrup.com/C++11FAQ.html#member-init). For integral types I have set default values, for other types default constructors will be called.

hyperxor:
@dennisjenkins75 I have written a sample:
```
#include <iostream>

struct A
{
    A() = default;

    void dump()
    {
        std::cout << "[A]\n";
    }
};

struct B
{
    B() = default;

    void dump()
    {
        std::cout << "[B]\n";
    }
};

struct C
{
    C() = default;

    void dump()
    {
        a.dump();
        b.dump();
        std::cout << "[C]\n";
    }

private:
    A a;
    B b;
};

int main()
{
    C c;
    c.dump();
}
```

and checked it by the last version of cppcheck:
```
$ ./cppcheck --enable=all --inconclusive /home/igor/playground/ex1
Checking /home/igor/playground/ex1/ex1.cpp ...
/home/ws11/Devel/playground/z1/ex1.cpp:7:10: performance:inconclusive: Technically the member function 'A::dump' can be static (but you may consider moving to unnamed namespace). [functionStatic]
    void dump()
         ^
/home/igor/playground/ex1/ex1.cpp:17:10: performance:inconclusive: Technically the member function 'B::dump' can be static (but you may consider moving to unnamed namespace). [functionStatic]
    void dump()
         ^
```
Everything is ok with constructors.

PR #788, opened 2020-06-26
Ensure no epoll events are processed for closed peer files

chrisvroberts:
An error existed in transport.cc where it incorrectly assumed that
it would not receive epoll events relating to a closed peer
connection. This is not the case because file descriptors can be
shared between processes e.g. on exec. When they are, events could
still be generated. The solution is to explicitly de-register our
interest in the file descriptor before we close it using
epoll_ctl(EPOLL_CTL_DEL). In addition the call to accept in
listener.cc has been updated to set CLOEXEC on the peer file
descriptors as these should not be shared with other processes
anyway.


chrisvroberts:
This may resolve some open issues. In particular oktal/pistache#364 "Unknown fd" and also perhaps oktal/pistache#501?

PR #801, opened 2020-07-18
Small refactoring in asyncWriteImpl

hyperxor:
In this PR I'm proposing refactoring in `asyncWriteImpl` method: extracting two methods `sendRawBuffer` and `sendFile` to improve code readability.

### DIFF
--- a/include/pistache/transport.h
+++ b/include/pistache/transport.h
@@ -194,6 +194,8 @@ private:
 
   // This will attempt to drain the write queue for the fd
   void asyncWriteImpl(Fd fd);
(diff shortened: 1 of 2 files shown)


hyperxor:
@dennisjenkins75 It seems that `run_http_server_test` test is flaky. I'm not sure, but looks like that crash is here `src/common/net.cc:303`:
```
...
303	            std::string(inet_ntoa(*reinterpret_cast<struct in_addr *>(*addr)));
...
```

I will think about it.

hyperxor:
@dennisjenkins75 Hello, can I ask you to look at this PR?

PR #808, opened 2020-08-02
Implement router static checks

andixlm:


andixlm:
@dennisjenkins75 Do you mean clang-format, right? :) I used clang-format-9 and applied it to the lines I changed without any extra args. I can execute it on the whole header file, but I'm a bit confused by the result, diff's here: https://pastebin.com/kR7Djcmj

Command: `clang-format-9 -i include/pistache/router.h`
There's no .clang-format config or smth related, just default clang-format-9.

It looks like the header file is not properly formatted or newer clang-tools have updated default format style. BTW, which version of clang do you use? Would clang-format-9 applied to the header file be acceptable?

PR #811, opened 2020-08-09
Check pull requests source code formatting

andixlm:
LLVM style is set explicitly as I found out that clang-format default style on my Debian is Google style (idk why, can't find any config for this behavior). And LLVM style is used already in Pistache that's why it's set explicitly.

Comparison is performed based on diff, so only modified lines are checked, not the whole file. Once everything is formatted properly, full check is useless, hence next PR should clang-format all the header and source code files.

Check is performed on pull requests only.

Failure example:
image

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,6 +31,7 @@ _apt_pkgs: &_apt_pkgs
   - 'gdb'
   - 'lcov'
   - 'cppcheck'
+  - 'clang-format-9'
 
 _apt_srcs: &_apt_srcs
   - 'ubuntu-toolchain-r-test'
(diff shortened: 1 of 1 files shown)


andixlm:
> This is great, but one caveat (sorry that I might have failed to mention it earlier). "include/pistache/date.h" is not "ours". IMHO, it should not be reformatted. We should move it to "include/pistache/thirdparty" and exclude it from formatting or format checking. Pistache uses some of its functionality, but I'd rather not reformat it in case we want to patch it with an upstream bugfix someday.
> 
> The above is my opinion, and I'm very open to having it debated or changed. Thoughts?

Actually, I agree. If we use smth external, it should be out of this project sources scope. Another question is how to exclude it. Since it's the only header that's taken from another repo, it's meaningless to add that repo as git submodule, obviously. Also, cmake subproject looks like overhead as well. Third party include dir looks pretty while it's the only place to put externals and it's easy to exclude it from check, see the latest commit.

andixlm:
Here it is – https://github.com/oktal/pistache/pull/812

andixlm:
Rebased and squashed

Issue #798, opened 2020-07-14
Memory needed during file upload


harryve:
Hi,

I have created a REST server for uploading files. The code is similar to the code by xoac in issue #105. My problem is that the pistache library allocates a lot of memory during the upload. After the file is written to disk, the memory is freed. 
The allocated amount of memory is about ten times the size of the file. This means that I cannot upload files of 50 MB on my embedded system because I run out of memory.
Is there any way to fix this problem?

dennisjenkins75:
Please show us the code where you initialize the Endpoint, and what options you pass in.

harryve:
Here is my code:
```
class UpdateServer {
public:
    explicit UpdateServer(Address addr)
        : httpEndpoint(std::make_shared<Http::Endpoint>(addr))
    {
    }

    void init(size_t maxRequestSize = 4096)
    {
        auto opts = Http::Endpoint::options().maxRequestSize(maxRequestSize);
        httpEndpoint->init(opts);
        setupRoutes();
    }

    void start()
    {
        httpEndpoint->setHandler(router.handler());
        httpEndpoint->serve();
    }

private:
    void setupRoutes()
    {
        using namespace Rest;

        Routes::Post(router, "/ready", Routes::bind(&Generic::handleReady));
        Routes::Post(router, "/file", Routes::bind(&Generic::handleFile));
    }

    std::shared_ptr<Http::Endpoint> httpEndpoint;
    Rest::Router router;
};

int main(int argc, char *argv[])
{
    Port port(9080);
    Address addr(Ipv4::any(), port);

    cout << "Cores = " << hardware_concurrency() << endl;
    UpdateServer updateServer(addr);
    updateServer.init(50*1024*1024);
    updateServer.start();
}
```

hyperxor:
@harryve Hello, can you show `handleReady` and `handleFile` handlers implementation?

harryve:
Of course (this is actually the first part of my code):
```#include <pistache/http.h>
#include <pistache/router.h>
#include <pistache/endpoint.h>
#include <fstream>
#include <chrono>
#include <iostream>

using namespace Pistache;
using namespace std;

namespace Generic {
void handleReady(const Rest::Request&, Http::ResponseWriter response)
{
    cout << "Ready" << endl;
    response.send(Http::Code::Ok); //, "1");
}

void writeToFile(const string& filename, const string& data)
{
    ofstream out;
    out.open(filename.c_str());
    out << data;
    out.close();
}

[...]

PR #814, opened 2020-08-25
Use a std::shared_ptr<std::string> for the body_ member variable in Pistache::Http::Message

armessia:
Use a std::shared_ptr<std::string> instead of std::string for the body_ member variable in Pistache::Http::Message
    
As a Pistache::Http::Message object is copied several times during the handling of a request this leads to a significant reduction in memory usage when uploading files through a http post.

### DIFF
--- a/include/pistache/http.h
+++ b/include/pistache/http.h
@@ -71,14 +71,14 @@ public:
   friend class Private::BodyStep;
   friend class ResponseWriter;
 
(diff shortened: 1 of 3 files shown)


armessia:
This fixes the following issue: #798

kiplingw:
Did something happen to all of our CI testing? I don't see any tests performed.

PR #828, opened 2020-10-15
CMakeLists.txt: respect BUILD_SHARED_LIBS

ffontaine:
Don't build and install pistache_shared if the standard cmake
BUILD_SHARED_LIBS is set to OFF

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,6 +22,7 @@ if(CMAKE_BUILD_TYPE MATCHES "Debug")
     set(CMAKE_CXX_FLAGS  "${CMAKE_CXX_FLAGS} -g -O0 -fprofile-arcs -ftest-coverage -fstack-protector-all --param=ssp-buffer-size=4")
 endif()
 
+option(BUILD_SHARED_LIBS "build shared library" ON)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -22,6 +22,7 @@ if(CMAKE_BUILD_TYPE MATCHES "Debug")
     set(CMAKE_CXX_FLAGS  "${CMAKE_CXX_FLAGS} -g -O0 -fprofile-arcs -ftest-coverage -fstack-protector-all --param=ssp-buffer-size=4")
 endif()
 
+option(BUILD_SHARED_LIBS "build shared library" ON)
```
dennisjenkins75, line 25: All of the other build-time flags begin with PISTACHE.  Is there a strong reason to add another flag that does not follow this convention?

ffontaine: Yes, there is a strong reason, `BUILD_SHARED_LIBS` is a standard cmake option: https://cmake.org/cmake/help/v3.0/variable/BUILD_SHARED_LIBS.html.

PR #829, opened 2020-10-16
fix chunked transfer decoding

BenjaminBeichler:
The handling of chunked transfer-encoding is broken. 

The usage of complete body_.size() seems to be always wrong. I don't know why this has worked in the first place, maybe only if there is one chunk?

I introduced a separate variable to count the already received bytes of the chunk.

A second bug was at the condition to check, whether the chunk could be completely appended, since the chunk size exclude the trailing /r/n, but we need to have that /r/n, otherwise it is detected as the next start of a chunk (with a missing number of bytes, but this is interpreted as 0). Therefore, we need to check against (size +2).

The second bug is only triggered in very special situations, e.g. a very slow connection, or if the chunk size is bigger or the same as the proccesing buffer size (mhh thats not that rare actually :-) .

I used this commandline to trigger the bug: ``` cat badRequest.http | dd bs=1 | netcat localhost 80 ```

Bad request needs to be a chunked request.

### DIFF
--- a/include/pistache/http.h
+++ b/include/pistache/http.h
@@ -502,6 +502,7 @@ private:
     Message *message;
     size_t bytesRead;
     ssize_t size;
+    ssize_t alreadyAppendedChunkBytes;
   };
 
   State parseContentLength(StreamCursor &cursor,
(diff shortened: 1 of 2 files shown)


kiplingw:
I'll also add to please test this in autopkgtest as described in the CI section of our README.md.

BenjaminBeichler:
Sorry for the late answer ... I try to provide a test, but currently my time budget is very small for this. As I'm using the openAPI generator for my application, I'm a bit overwhelmed from the current Unit-Test descriptions, but maybe I find one which is easy to adjust for triggering this error.

PR #832, opened 2020-11-04
Prevent SSL write failing after temporary error.

polmr:


polmr:
Sure, that is a `save the day` piece of code. Hopefully you can integrate it better with async core.
Note: `Transport::sendFile` -> `ssize_t SSL_sendfile(SSL *out, int in, off_t *offset, size_t count)` could probably use something similar.
Also SSL_read is also susceptible of failing likewise.

src/common/transport.cc
```diff
@@ -290,6 +290,26 @@ ssize_t Transport::sendRawBuffer(Fd fd, const char* buffer, size_t len, int flag
   if (it_->second->ssl() != NULL) {
     auto ssl_ = static_cast<SSL *>(it_->second->ssl());
     bytesWritten = SSL_write(ssl_, buffer, static_cast<int>(len));
+    if (bytesWritten <= 0 && SSL_get_error(ssl_, bytesWritten) == SSL_ERROR_WANT_WRITE)
```
Ne02ptzero, line 293: I think the `SSL_MODE_ENABLE_PARTIAL_WRITE` should be a better option.

```
The write functions will only return with success when the complete contents of
buf of length num has been written. This default behaviour can be changed with
the SSL_MODE_ENABLE_PARTIAL_WRITE option of SSL_CTX_set_mode(3).  When this
flag is set the write functions will also return with success when a partial
write has been successfully completed. In this case the write function
operation is considered completed. The bytes are sent and a new write call with
a new buffer (with the already sent bytes removed) must be started. A partial
write is performed with the size of a message block, which is 16kB.
```

You've got multiple options:
- SSL_write send all the bytes, and returns it, no trouble
- SSL_write did not send all the bytes, and returns it, and the underlying event-loop will retry this write with the updated buffer position (Which is the same behavior as `send(2)` 3 lines below)

(This should also mitigate the issue in `SSL_sendfile`)

Either way, I do not consider a sleep in an event-loop a valid solution.

Though your patch raises a valid point:
- What does happen when `SSL_write` (or `send`, for that matter) returns `< 0`?

kiplingw:
@Ne02ptzero, I've tried cleaning up his patch slightly by fixing the type casting warnings I was getting with `bytesWritten` being a `long int` where it was needed as an `int`. I also took @dennisjenkins75's recommendation to not bail on a failed assert.

I ran into what I suspect is this bug last night and this is a critical blocker. Does this look ok now, notwithstanding his sleeping in the write loop?

Ne02ptzero:
> @Ne02ptzero, I've tried cleaning up his patch slightly by fixing the type casting warnings I was getting with `bytesWritten` being a `long int` where it was needed as an `int`. I also took @dennisjenkins75's recommendation to not bail on a failed assert.
> 
> I ran into what I suspect is this bug last night and this is a critical blocker. Does this look ok now, notwithstanding his sleeping in the write loop?

See https://github.com/pistacheio/pistache/pull/995

kiplingw:
Closing to move progress over to @Ne02ptzero's #995.

PR #834, opened 2020-11-16
Expose disconnection

iroddis:
Http::Handler made the onDisconnection(..) callback private. This PR removes that mask so that derived classes can get notification of disconnections.

This is handy for cases where requests might take a while to complete, and if a client disconnects it's possible to get notified to stop processing.

- Added basic functionality
- Added test case in http_server_tests.cc
- Verified all parts compile
- Verified all tests pass (gcc 10.2.1)

### DIFF
--- a/include/pistache/http.h
+++ b/include/pistache/http.h
@@ -581,7 +581,6 @@ public:
 
 private:
   void onConnection(const std::shared_ptr<Tcp::Peer> &peer) override;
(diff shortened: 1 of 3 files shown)


iroddis:
Actually this PR was already merged in via #838. I had originally submitted them as two PRs, since they were somewhat independent. Since that one was merged, I'll just close out this PR. Sorry for the confusion.

PR #836, opened 2020-11-17
Adding a Peer ID to identify connections

iroddis:
Building on PR #834 , this PR adds a per-connection identifier that makes it easier to keep track of which client is connecting, and reference their request.

Use case: Long-running queries can be assigned an ID, rather than using the peer FD (reused), which helps with debugging and provides an easier accounting method when cleaning up a long-running query's client disconnect before a full response has been generated.

### DIFF
--- a/include/pistache/http.h
+++ b/include/pistache/http.h
@@ -581,7 +581,6 @@ public:
 
 private:
   void onConnection(const std::shared_ptr<Tcp::Peer> &peer) override;
-  void onDisconnection(const std::shared_ptr<Tcp::Peer> &peer) override;
   void onInput(const char *buffer, size_t len,
                const std::shared_ptr<Tcp::Peer> &peer) override;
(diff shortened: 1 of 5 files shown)


kiplingw:
@dennisjenkins75, I'm fine with signing off on this if you are.

kiplingw (approved):

kiplingw:
Broke in the PPA for all architectures. Here's the [log](https://launchpadlibrarian.net/507608521/buildlog_ubuntu-groovy-amd64.pistache_0.0.002-202011180218~ubuntu20.10.1_BUILDING.txt.gz) for amd64.

kiplingw:
That's what I was thinking too.

iroddis:
Sorry about that. This was one of those "this small change to tidy things won't affect anything" commits, that obviously did. I've applied the fix, did a complete, clean build, and ran all the tests (all pass).

Issue #842, opened 2020-11-24
Unit test client_multiple_requests_disconnects_handled is failed


hyperxor:
Hello,

I have run unit tests on the latest souce and got the following error:
 
```
12: [ RUN      ] http_server_test.client_multiple_requests_disconnects_handled
12: Trying to run server...
12: INITING
12: Server address: localhost:38919
12: [server] Sent: 127.0.0.1
12: [server] Sent: 127.0.0.1
12: [server] Sent: 127.0.0.1
12: Response code is OK
12: Response code is OK
12: Response code is OK
12: resolves: 3, rejects: 0, timeout: 1, wait: 6
12: /home/hyperxor/code/pistache/tests/http_server_test.cc:475: Failure
12: Value of: CLIENT_REQUEST_SIZE
12:   Actual: 3
12: Expected: counter
12: Which is: 0
12: [  FAILED  ] http_server_test.client_multiple_requests_disconnects_handled (1003 ms)
```

It seems it happened after this #838 pull request.

hyperxor:
Hello, @iroddis, can you describe the expected scenario of the `client_multiple_requests_disconnects_handled` unit test, maybe I can fix it.

dennisjenkins75:
I cannot reproduce the failure on my dev box (From right after PR #840 is merged).  I built with g++ and clang++, both in "Release" mode and "Debug" mode, and ran all unit tests (all pass), plus my own app runs fine.

All 4 builds are very noisy with compiler warnings (I'd love for those to be cleaned up some day), but all 4 "worked".

hyperxor@, what is your build platform?

```
$ uname -a
Linux ostara 5.4.72-gentoo-x86_64 #1 SMP PREEMPT Mon Nov 9 22:45:29 PST 2020 x86_64 Intel(R) Core(TM) i5 CPU 760 @ 2.80GHz GenuineIntel GNU/Linux

$ g++ --version
g++ (Gentoo 10.2.0-r3 p4) 10.2.0

$ clang++ --version
clang version 11.0.0
```

hyperxor:
@dennisjenkins75 Ok, thank you, I will see.

```
$ uname -a
Linux hyperxor-ws 5.4.0-54-generic #60-Ubuntu SMP Fri Nov 6 10:37:59 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux

$ g++ --version
g++ (Ubuntu 9.3.0-17ubuntu1~20.04) 9.3.0
```

[...]

PR #845, opened 2020-11-26
Fix issue #842

hyperxor:
In this PR I'm fixing #842 

1) Move server shutdown after delay to fix test;
2) Reading `handler->activeConnections.size()` and comparing with `0UL` is wrong is the test because changes for `activeConnections` happen in the **copy** of handler object (`ClientCountingHandler`).

### DIFF
--- a/tests/http_server_test.cc
+++ b/tests/http_server_test.cc
@@ -419,7 +419,7 @@ TEST(http_server_test, response_size_captured) {
 struct ClientCountingHandler : public Http::Handler {
   HTTP_PROTOTYPE(ClientCountingHandler)
 
-  ClientCountingHandler(std::atomic<size_t> & counter) : counter_(counter) { std::cout << "INITING" << std::endl;}
+  ClientCountingHandler(std::atomic<size_t> & counter) : counter_(counter) { std::cout << "[server] Ininting.." << std::endl; }
 
   void onRequest(const Http::Request &request,
                  Http::ResponseWriter writer) override {
@@ -440,8 +440,7 @@ struct ClientCountingHandler : public Http::Handler {
     activeConnections.erase(peer->getID());
   }
 
-  size_t getClientsServed() const { return counter_; }
-
+private:
   std::unordered_set<size_t> activeConnections;
 
   std::atomic<size_t> & counter_;
@@ -469,9 +468,8 @@ TEST(
   const size_t CLIENT_REQUEST_SIZE = 3;
   clientLogicFunc(CLIENT_REQUEST_SIZE, server_address, 1, 6);
 
+  std::this_thread::sleep_for(std::chrono::seconds(2));
   server.shutdown();
-  std::this_thread::sleep_for(std::chrono::seconds(1));
 
   ASSERT_EQ(counter, CLIENT_REQUEST_SIZE);
-  ASSERT_EQ(handler->activeConnections.size(), 0UL);
 }
--- a/tests/http_server_test.cc
+++ b/tests/http_server_test.cc
@@ -419,7 +419,7 @@ TEST(http_server_test, response_size_captured) {
 struct ClientCountingHandler : public Http::Handler {
   HTTP_PROTOTYPE(ClientCountingHandler)
 
-  ClientCountingHandler(std::atomic<size_t> & counter) : counter_(counter) { std::cout << "[server] Ininting.." << std::endl; }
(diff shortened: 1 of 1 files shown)


tests/http_server_test.cc
```diff
@@ -469,9 +468,8 @@ TEST(
   const size_t CLIENT_REQUEST_SIZE = 3;
   clientLogicFunc(CLIENT_REQUEST_SIZE, server_address, 1, 6);
 
+  std::this_thread::sleep_for(std::chrono::seconds(2));
```
dennisjenkins75, line 471: Instead of sleeping a fixed amount of time, can you wait (with timeout) on a semaphore?  sleeping like this will slow down the unit tests (slightly, but it can add up) and make them brittle.  If ran on a machine that is oversubscribed for CPU resources (eg, load avg > # logical CPU cores), then this test can fail due to CPU starvation.

kiplingw:
@oktal

tests/http_server_test.cc
oktal: I'll try to dig more deeply into the logic behind this test when I get some time but I do agree with @dennisjenkins75 

When tests are running asychronously, instead of waiting an arbitrary amount of time (which will slow down unit-tests like Dennis said), I prefer using a `std::mutex` with a `std::condition_variable` to wait on for a period of time (timeout) before making the test fail.

For example, maybe could you just lock the `activeConnections` and add

```
template<typename Duration>
bool waitForActiveConnections(size_t count, Duration timeout);
```

function and then assert on that:

```
ASSERT_TRUE(handler->waitActiveConnections(CLIENT_REQUEST_SIZE, std::chrono::seconds(2)));
```

hyperxor:
@dennisjenkins75 @oktal Thank you for suggestions, I will think about them. By the way, it's impossible to implement the following code:
```
handler->waitActiveConnections
```
As I have already said before everything happens  in the **copy** of `handler` object (`ClientCountingHandler`) on the server side.

oktal:
I see what you mean.

So, from my understanding, this test is supposed to assert that:
* the server correctly served CLIENT_REQUEST_SIZE requests ;
* no more peer is connected from the server-side

Can you confirm ?

hyperxor:
@oktal  I'm not an author of this test and I asked the question here #842 about the expected scenario of the this test.

    

- the server correctly served CLIENT_REQUEST_SIZE requests ;

- no more peer is connected from the server-side

I would add:
- `CLIENT_REQUEST_SIZE` clients have been disconnected.

hyperxor:
@oktal @dennisjenkins75 I have pushed slightly modificated code based on your suggestions. I will think a little bit how to do it better.

hyperxor:
@dennisjenkins75 Ok, I don't plan push any changes to this PR.

kiplingw:
I'm fine with it.

kiplingw:
Unfortunately it still appears to be happening. I saw [this](https://launchpadlibrarian.net/530759192/buildlog_ubuntu-hirsute-amd64.pistache_0.0.002-202103292104~ubuntu21.04.1_BUILDING.txt.gz) today in the unstable build server for Hirsute on amd64:

```
...
[ RUN      ] http_server_test.client_multiple_requests_disconnects_handled
Trying to run server...
[server] Ininting...
Server address: localhost:36565
[server] Sent: 127.0.0.1
Response code is OK
[server] Sent: 127.0.0.1
[server] Sent: 127.0.0.1
Response code is OK
Response code is OK
resolves: 3, rejects: 0, timeout: 1, wait: 6
[server] Disconnect from peer ID 19 connecting from 127.0.0.1
[server] Disconnect from peer ID 20 connecting from 127.0.0.1
./tests/http_server_test.cc:503: Failure
Expected equality of these values:
  result
    Which is: false
  true
[  FAILED  ] http_server_test.client_multiple_requests_disconnects_handled (2202 ms)
[----------] 8 tests from http_server_test (23693 ms total)

[----------] Global test environment tear-down
[==========] 8 tests from 1 test suite ran. (23795 ms total)
[  PASSED  ] 7 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] http_server_test.client_multiple_requests_disconnects_handled

 1 FAILED TEST

      Start 13: http_client_test
13/26 MemCheck #13: http_client_test .................   Passed   18.77 sec
      Start 14: listener_test
14/26 MemCheck #14: listener_test ....................   Passed   15.66 sec
      Start 15: request_size_test
15/26 MemCheck #15: request_size_test ................   Passed    9.17 sec
      Start 16: streaming_test
...
```

PR #847, opened 2020-11-29
Delete unused code and clang-format

hyperxor:
The field `Fd fd;` of `Event` class is not used in code. The `fd` value is obtained from `tag` field.


kiplingw:
Looks fine to me. @dennisjenkins75  ?

PR #855, opened 2020-12-30
Make fPIC optional

uilianries:
For static library propose, fPIC is not mandatory, however, the current behavior is forcing it.

This PR brings fPIC as optional, but keeps it activated by default.


kiplingw:
Thanks @uilianries. Please fix the build issues Travis CI identified, and then follow the CI instructions in `README.md` to pass it through `autopkgtest`.

kiplingw (changes requested):
See other comments in main thread.

kiplingw:
Passes autopkgtest on amd64.

Issue #835, opened 2020-11-17
Build failure on Clang - reference to 'string_view' is ambiguous


Tachi107:
I'm unable to build Pistache on Clang 11 when using libc++ 11 (through -stdlib=libc++ command line parameter), as I get the compiler error `reference to 'string_view' is ambiguous` in multiple files (include/pistache/string_view.h and include/pistache/router.h).
Compiled on Debian Testing.

Edit:
Forgot to mention that this seems related to #826

dennisjenkins75:
string_view was added to the C++ standard with c++17.  At the time our private implementation was added to pistache, I think that we were still compiling with C++11 or c++14, but the contributor wanted to use string_view.

We need to determine at compile time if we should omit our private implementation, or just move entirely to c++17 and remove it.  However, IDR which distros we "support" but still only supply c++14 compilers.

Kip, Octal, your thoughts?

kiplingw:
I think it's time to move entirely to C++17 in my view - no pun intended. @oktal ? One less thing to have to maintain when relying on the system's STL implementation.

hyperxor:
@kiplingw @Tachi107 Hello, I think this issue can be closed, am I right?

Tachi107:
Nope, #897 hasn't been merged yet.
Unrelated but also related, sorry for my recent inactivity, but I'm approaching the end of high school and I'm concentrating most of my time into that. I've also broke a finger, and typing isn't exactly pleasant haha.
I hope that by the time I'll have more time to spend on this project oktal will have reviewed my PR, so that we'll be one step closer to 1.0 :)

hyperxor:
@Tachi107 Hello, I'm not sure, but I think that it would be better to split #897 on several PRs. There are lot of changes that are not related to the name your PR.
@kiplingw

Tachi107:
I'll do it as soon as I'll have time

kiplingw:
Thanks @Tachi107.

PR #859, opened 2021-01-13
Upgrade to C++17 and use std::optional

### LINKED ISSUE #858: [FEATURE REQUEST] used std::optional if available
`Pistache::Optional` has a slight, but noticeably, different interface to c++17's `std::optional`. Similar to how `include/pistache/string_view.h` uses c++17's `std::string_view` if available, it would be nice if `Pistache::Optional` could be an alias for `std::optional` using the same `__has_include` trick to only conditionally build the custom implementation.

Here are the API differences between the pistache and the standard,
| `pistache::Optional` | `std::optional`|
|---------------------|---------------|
| `getOrElse` | [`value_or`](https://en.cppreference.com/w/cpp/utility/optional/value_or) |
| `isEmpty` | [`has_value`](https://en.cppreference.com/w/cpp/utility/optional/has_value) |
| `get` | [`value`](https://en.cppreference.com/w/cpp/utility/optional/value)/[`operator*`](https://en.cppreference.com/w/cpp/utility/optional/operator*) |
| `Pistache::types::None` | [`std::nullopt_t`](https://en.cppreference.com/w/cpp/utility/optional/nullopt_t) |
| `Pistache::types::Some` | [`std::optional::optional`](https://en.cppreference.com/w/cpp/utility/optional/optional) has `T` constructors |
| - | `std::is_trivial<std::optional<T>>::value == std::is_trivial<T>::value`  |
| - | modifiers: [`swap`](https://en.cppreference.com/w/cpp/utility/optional/swap),  [`reset`](https://en.cppreference.com/w/cpp/utility/optional/reset),  [`clear`](https://en.cppreference.com/w/cpp/utility/optional/clear) |

## Describe the solution you'd like
Seeing how different the implementation is, it may be worth importing a c++11 compatible `std::string_view` implementation.
For example, ~~[`string_view-standalone`](https://github.com/bitwizeshift/string_view-standalone) (licensed MIT,)~~* [`tl::optional`](https://github.com/TartanLlama/optional) could replace the current optional implementation, if the `__has_include(<optional>)` check fails.

This is an API breaking change, as downstream users will need to update their usage. In the code base that I work in, the only usage of `Pistache::Optional` that would need to be updated are, `getOrElse` -> `value_or`, `isEmpty` -> `has_value`, and `.get()` -> `.value()` (or `operator->`,) as we never construct or pass `Pistache::Optional`'s outside of the initial extraction of data from a request.

* edit: had a brain fart and posted a `std::string_view` alternative, not a `std::optional` alternative.

jacobbogdanov:
Fixes #858.

As per the recommendation by @kiplingw in https://github.com/pistacheio/pistache/issues/858#issuecomment-758279752 this PR upgrades from C++14 to C++17 to take advantage of `std::optional`.


kiplingw:
Thanks @jacobbogdanov. I'm going to hold this for the time being since this will have consequences for many that use the library. @oktal, @dennisjenkins75 and I will discuss this internally at our next team call 14 February, 2021. Personally I am in favour of the migration, but it should be discussed.

kiplingw:
@jacobbogdanov, a few things while we discuss internally. If we do decide to migrate to C++17, then you'll also need to ensure:

- `README.md` is updated;
- Update `debian/control` package description;
- Update `index.html`
- Think about whether `-std=c++17` should propagate down into the `pkg-config` manifest, which is the standard way people use the `libpistache-dev` package. I don't know what the best practise is in this regard, and you may want to liaise with with someone with freedesktop.org.

kiplingw:
@jacobbogdanov, we discussed internally and we're fine with your PR in principal - notwithstanding the above issues in need of addressing. What we are going to do is hold your PR for the time being. When @oktal is finished a fix for the only remaining critical issue (that we are aware of), we will merge your PR and prepare our first stable 1.0.0 release.

kuzkry (commented):
Hey! Nice work. Shouldn't you also change README.md?
```
- It is entirely written in pure-C++14 and provides a clear and pleasant API.
+ It is entirely written in pure-C++17 and provides a clear and pleasant API.
```

kiplingw:
Yup. I mentioned that above.

kuzkry:
Oops, I didn't notice :)

Tachi107:
Upgrading to C++17 would also fix #835

kiplingw:
@jacobbogdanov, can you try rebasing?

Tachi107:
Some other changes related to C++17

- [include/pistache/common.h#L63](https://github.com/pistacheio/pistache/blob/6416a53948e2f06783705056ae0b5904b8cc852d/include/pistache/common.h#L63)
- [`[[maybe_unused]]`](https://en.cppreference.com/w/cpp/language/attributes/maybe_unused) (if a parameter is not used, why not `/* comment out  */` its name?) 
- [include/pistache/http.h#L358](https://github.com/pistacheio/pistache/blob/6416a53948e2f06783705056ae0b5904b8cc852d/include/pistache/http.h#L358)
- [Nested namespaces](https://clang.llvm.org/extra/clang-tidy/checks/modernize-concat-nested-namespaces.html) ?
- [`[[nodiscard]]`](https://en.cppreference.com/w/cpp/language/attributes/nodiscard) ?
- [`std::scoped_lock`](https://en.cppreference.com/w/cpp/thread/scoped_lock) instead of [`std::lock_guard`](https://en.cppreference.com/w/cpp/thread/lock_guard)
- [`std::string_view`](https://en.cppreference.com/w/cpp/string/basic_string_view)

kiplingw:
@Tachi107, feel free to make any of those changes in his branch. Just be careful and be sure to test.

kiplingw:
@jacobbogdanov, there were three issues that we wanted to address before a 1.0 release. The first was DoS issue in #864 which was just fixed this morning. The second was implementing `format` and `formatcheck` hooks. The third is yours. The first two are merged now (correct me if wrong @oktal), so that just leaves your issue. If you could please rebase, we're almost ready for our 1.0 release.

Tachi107:
If we upgrade to C++17 I think that we should also replace the custom `std::string_view` implementation, that causes issues with libc++. I'll try to prepare a PR in the following hours. 

Talking about the 1.0 release, I have a question (it isn't related to this PR so sorry for being off topic); what about renaming the pkg-config file to `pistache.pc` instead of `libpistache.pc`? Since the project is named "Pistache" it would make more sense, in my opinion. This would also make working with Meson subprojects easier; if the pkg-config file is named the same way as the git repository, one could simply write `dependency('pistache')` to use the installed version if available or building the project from source otherwise; currently it is required to search for "libpistache" and manually specify the fallback if the dependency is not found, so `depenency('libpistache', fallback: 'pistache')`. The Meson thing is not a big deal at all, but I find `pistache.pc` to be the more intuitive option.

kiplingw:
You could rename it, but be mindful of three things: All existing user build environments that use the library will break, pending their renaming of the name passed to pkg-config. Second thing is it may be necessary if the name isn't whatever the current best practise recommended by the freedesktop.org standard. Third, if rename is necessary, don't forget to update the readme.

Tachi107:
freedesktop.org recommends both `libname.pc` and `name.pc`, but thinking about it the change is not worth the issues it would cause. Also, `libpistache` matches the Debian package name, so it makes sense to have them named the same way.

jacobbogdanov:
all conflicts should be resolved, and everything seems to be compiling (on my machine at least.) I haven't been able to get the tests running locally and was hoping to see if there were any issues using CI... but I don't see a link to travis. Is there something I need to do to trigger that?

Tachi107:
Travis integration is currently broken (#893), but you can see your build [here](https://travis-ci.org/github/pistacheio/pistache/jobs/766917657). It is failing because you didn't remove `optional.h` from [include/pistache/meson.build#L27](https://github.com/pistacheio/pistache/blob/86ef2a43e1b8b05ecafe45634390963989a09b57/include/pistache/meson.build#L27)

Tachi107:
Sorry, I just realised that you have to also remove `optional_test` from [tests/meson.build#L27](https://github.com/jacobbogdanov/pistache/blob/optional/tests/meson.build#L27)

kiplingw:
Running through autopkgtest now. @jacobbogdanov, let me know which steps in the readme broke for you in preparing your own local environment.

tests/headers_test.cc
```diff
@@ -821,10 +821,10 @@ TEST(headers_test, raw_headers_are_case_insensitive)
         step.apply(cursor);
 
         // or the header you try and get, it should work:
-        ASSERT_FALSE(request.headers().tryGetRaw("Custom-Header").isEmpty());
-        ASSERT_FALSE(request.headers().tryGetRaw("CUSTOM-HEADER").isEmpty());
-        ASSERT_FALSE(request.headers().tryGetRaw("custom-header").isEmpty());
-        ASSERT_FALSE(request.headers().tryGetRaw("CuStOm-HeAdEr").isEmpty());
+        ASSERT_FALSE(request.headers().tryGetRaw("Custom-Header").has_value());
```
Tachi107, line 824: Changing from `isEmpy()` to `has_value()` inverts the boolean returned, so you should also change `ASSERT_FALSE` to `ASSERT_TRUE` ;)

kiplingw:
Well @jacobbogdanov, the good news is that your PR is probably fine! But I can't be certain because it managed to "solve" another problem in reproducing #842. That issue has been quite elusive to replicate, but since I ran your PR in autopkgtest, the thread scheduler, or at least on my local machine, appears to have been altered just enough to create whatever the underlying race condition that @hyperxor has been trying to track down.

tests/mime_test.cc
```diff
@@ -21,9 +21,9 @@ TEST(mime_test, basic_test)
     ASSERT_TRUE(m2.toString() == "application/xhtml+xml");
 
     auto m3 = MIME(Text, Plain);
-    ASSERT_TRUE(m3.q().isEmpty());
```
Tachi107, line 24: Same thing here

kiplingw:
Good eye, @Tachi107, but that's not the test that's failing. It's the `http_server_test.client_multiple_requests_disconnects_handled` test.

Tachi107:
Yes, but also `mime_test` and `headers_test` are failing, see [here](https://travis-ci.org/github/pistacheio/pistache/jobs/766935342#L3095)

kiplingw:
Indeed, but if you look carefully you'll see Travis CI didn't catch the `client_multiple_requests_disconnects_handled` race condition again. That's the more important one because the other failures are simple inversions of the asserts.

Tachi107:
Oh yes, but I doubt that's related to this pr

kiplingw:
His (@jacobbogdanov) PR is not the cause of the race condition, but his PR was helpful in changing the thread scheduler's timing in just the right way to make #842 reproducible.

kiplingw (changes requested):
@jacobbogdanov, can you please ensure your changes that required inversions @Tachi107 discovered of any unit test asserts are addressed? Once you've got that done I think this is ready for merging.

Tachi107:
> His PR is not the cause of the race condition, but his PR was helpful in changing the thread scheduler's timing in just the right way to make #842 reproducible

Oh I didn't understand that it is now 100% reproducible, nice!

Tachi107:
Tests are now passing (apart from `http_client_test.test_client_timeout`, but that's unrelated). 

The PR should be ready to be merged.

PR #864, opened 2021-02-08
Disconnect idle peers after some configurable timeout

### LINKED ISSUE #237: Denial of Service: silent connections will never get a timeout
Let's say you have started your Pistache server and it is listening on port `10000`.
You now run `nc localhost 10000` from the terminal.
Pistache picks this up as a new HTTP request, but since you are just opening the connection but never sending anything to the server, the connection remains hanging forever. Even if a timeout handler is specified it will never occur. This allows us to very easily exhaust all of the server's available TCP descriptors.

oktal:
closes #237


kiplingw:
Thanks @dennisjenkins75. I actually already tested via autopkgtest and it [fails](https://pastebin.com/RFHv4m0c). He may just need to rebase.

PR #863, opened 2021-02-06
Docusaurus migration

Tachi107:
Following @oktal's request in https://github.com/pistacheio/pistache/issues/862#issuecomment-773547718, here's the pistache.io documentation site migrated to docusaurus.

Documentation files are contained in pistache.io/docs, and are in Markdown format. The files that I added contain almost the same things as in the "old" site. 

Sidebar elements can be changed by editing pistache.io/sidebars.js. 

The homepage contains a few [undraw](https://undraw.co) images and, again, almost the same things as the original site. Differently from the documentation files, this one is written in (i believe) JSX, that is JavaScript plus HTML templating for React, if I remember correctly, but it could be changed to a Markdown file if needed.

I also wrote a GitHub Actions file that automatically deploys the site using GitHub Pages. Works for me in [PwRAu/scraf-backend/.github/workflows/raml-gh-pages.yml](https://github.com/PwRAu/scraf-backend/blob/main/.github/workflows/raml-gh-pages.yml), and I think should also work for Pistache. If it doesn't, there's [a better way of doing this](https://v2.docusaurus.io/docs/deployment), but it requires access to the repo's settings. 

Another thing that I would like to enable but requires access to the repo's settings is [HTTPS](https://docs.github.com/en/github/working-with-github-pages/securing-your-github-pages-site-with-https).

In the docs I also mention that Pistache requires a C++17 compiler. I know that this is not (yet) true, but this should make the switch (#859) easier.

Needless to say, this would fix #862.

Some screenshots:
image

image

image

PR #865, opened 2021-02-08
Fix #863 issues

Tachi107:
This PR should fix the GitHub Action (installing yarn via npm and fixing the directory that the action would push to the `gh-pages` branch), as well as one dead link pointing to a nonexistent blog section and a note about `std::string_view` that really shouldn't be published with the documentation.

I tested everything on a clone of the repo, and everything seems to work ([pistache.pappacoda.it](https://pistache.pappacoda.it) is on and fine)

### DIFF
--- a/.github/workflows/pistache.io.yaml
+++ b/.github/workflows/pistache.io.yaml
@@ -7,7 +7,7 @@ on:
     branches: [ master ]
 
 jobs:
-  build-and-deploy:
+  pistache.io-deploy:
     runs-on: ubuntu-20.04
 
     steps:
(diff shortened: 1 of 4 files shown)


Tachi107:
Alright, the action wasn't recognized by GitHub, probably because I set `pistache.io-deploy` ([.github/workflows/pistache.io.yaml](https://github.com/pistacheio/pistache/blob/master/.github/workflows/pistache.io.yaml), line 10)... Could you please change it to `pistacheio-deploy`?
It will work this time, I hope.

kiplingw:
Done.

PR #873, opened 2021-03-26
Fix #187

### LINKED ISSUE #187: please remove all relative namespace references in all documentation and examples
This has deeply impacted the quality of the documentation and the examples, making alot of it unusable.  Please replace all `use namespace <anything>` with subsequent code blocks that use absolute classpaths.

Tachi107:
This little PR fixes #187, while also replacing `std::endl` with `'\n'` in the examples, since most of the time you don't want to flush the output stream when outputting a newline.

Also, this PR should really only contain one commit; I'm not a Git expert, so I don't know how to exclude the three useless commits, but if I remember correctly when merging a PR there's the option to merge it as a single commit, so please do that :)


kiplingw:
Hey @Tachi107. You are right that generally there is no need to flush the output buffer after every line. However, because these are examples that are most likely used by people who are new to the library, flushing the buffer can lead to less confusion when they know immediately that something happened. The emphasis should be on pedagogy and comprehension over performance in the examples.

Tachi107:
That's true, buffering might not be expected by default. 

At the same time, since examples are examples they should reflect what you should actually do. Also, it is likely that examples are most useful for inexperienced people, who might think that `std::endl` is what you use in C++ to simply send a newline character, and that could cause issues expecially when working with files. A lot of people think that, and I was one of them.

See [this note](https://en.cppreference.com/w/cpp/io/manip/endl) from cppreference and also [this note](https://github.com/gcc-mirror/gcc/blob/14b554c462d5b6450fa24afb7ba55435ebd4b46f/libstdc%2B%2B-v3/include/std/ostream#L674) in the libstdc++ source

Obviously this is not an issue at all, and if you prefer to keep endl it's completely fine

kiplingw:
> See [this note](https://en.cppreference.com/w/cpp/io/manip/endl) from cppreference and also [this note](https://github.com/gcc-mirror/gcc/blob/14b554c462d5b6450fa24afb7ba55435ebd4b46f/libstdc%2B%2B-v3/include/std/ostream#L674) in the libstdc++ source

From cppreference:

_This manipulator may be used to produce a line of output immediately, e.g. when displaying output from a long-running process, logging activity of multiple threads or logging activity of a program that may crash unexpectedly._

That's a close description to the category I would put the examples in. Actually logging in general for most daemons really.

> Obviously this is not an issue at all, and if you prefer to keep endl it's completely fine

I think for the time being `std::endl` makes more sense, given the context.

Tachi107:
Why did you close the PR? I could've reverted the endl change, but issue 187 still needs to be fixed. Should I open a new PR with just that fix?

kiplingw:
No worries @Tachi107. If you want to fix it in this PR, feel free to do so.

Tachi107:
`std::endl` restored (except when using `std::cerr`), ready to merge

kiplingw (approved):

PR #876, opened 2021-03-30
Swagger: Adding Query parameter to route description.

bthrall:
We handled this by using enums to extend the Parameter class, and a function to allow easy adding of query parameters.

We're not 100% sure if the query function should be on the subpath as well, though that should be easily enough removed if necessary.

### DIFF
--- a/include/pistache/description.h
+++ b/include/pistache/description.h
@@ -172,8 +172,18 @@ template <typename T> std::unique_ptr<DataType> makeDataType() {
   return std::unique_ptr<DataType>(new DataTypeT<T>());
 }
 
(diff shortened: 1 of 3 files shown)


kiplingw:
Thanks @bthrall. @oktal, if this looks fine to you I'll merge. I'm not really familiar with the integrated Swagger functionality and haven't used it.

bthrall:
> I'm concerned that this PR directly modifies rapidjson instead of patching it upstream and then properly integrating that fix here. Are you doing a hard-fork of rapidjson?

It modifies the rapidjson serializer used by Pistache to generate the Swagger JSON.

PR #878, opened 2021-04-01
Add basic authentication support for Swagger JSON

bthrall:
Multiple authentication schemas are supported, for both query and header locations.

Security is only applied to the root at this time.

### DIFF
--- a/include/pistache/description.h
+++ b/include/pistache/description.h
@@ -143,6 +143,45 @@ private:
   Info *info_;
 };
 
+#define AUTHENTICATION_TYPES                                      \
+  AUTH_TYPE(ApiKey, "apiKey")
+
+enum class AuthenticationType {
+#define AUTH_TYPE(name, _) name,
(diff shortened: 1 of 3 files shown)


Tachi107:
> IMHO, we should not be making changes to rapidjson

[`include/pistache/thirdparty/serializer/rapidjson.h`](https://github.com/pistacheio/pistache/blob/697cf1f7a65333e99d4bfef9ff29dff49d8c84da/include/pistache/thirdparty/serializer/rapidjson.h) is not really part of RapidJSON, it is just a serializer used for Swagger/OpenAPI that uses RapidJSON to handle JSON (or something along these lines, I've never used nor looked into it). I don't know why it is in the `thirdparty` directory since it is part of Pistache.

kiplingw:
Hey @bthrall. Just checking in, but do you still intend to work on this PR? If so, you may want to start by rebasing.

PR #879, opened 2021-04-02
Fix CMake pkg-config example

### LINKED ISSUE #877: [BUG] Can't install via cmake
## Current Behavior
I am trying to install pistache and use it within my project. This already worked about three months ago and I successfully installed pistache on one of my pcs. However this does not work anymore:

```cmake
sudo apt-get install libssl-dev
mkdir third_party
cd third_party || exit
git clone https://github.com/oktal/pistache.git
cd pistache
git submodule update --init
mkdir build
mkdir prefix
cd build
cmake -G "Unix Makefiles" \
      -DCMAKE_BUILD_TYPE=Release \
      -DPISTACHE_BUILD_EXAMPLES=true \
      -DPISTACHE_BUILD_TESTS=false \
      -DPISTACHE_BUILD_DOCS=false \
      -DPISTACHE_USE_SSL=true \
      -DPISTACHE_INSTALL=true \
      -DCMAKE_INSTALL_PREFIX=$PWD/../prefix \
      ../
make -j 4
sudo make install
````

First problem:

The install fails because no install target exists in the makefile.

```bash
[ 91%] Built target pistache
Scanning dependencies of target pistache_static
Scanning dependencies of target pistache_shared
[ 95%] Linking CXX static library libpistache.a
[100%] Linking CXX shared library libpistache-0.0.002-git20210107.so
[100%] Built target pistache_static
[100%] Built target pistache_shared
make: *** No rule to make target 'install'.  Stop.
```

Second problem:

Even for the old build that worked I had to set the pistache_dir manually for cmake because it could not be detected by cmake otherwise. I tried the package install via the ppa but I have the same problem.

```cmake
set(Pistache_DIR "${CMAKE_SOURCE_DIR}/third_party/pistache/prefix/lib/cmake/pistache") 
find_package(Pistache REQUIRED)
target_link_libraries(rawStream ${OpenCV_LIBS} pistache_shared ${LIBS})
````

While the new build with meson works it does not generate the /pistache/prefix/lib/cmake/pistache folder for cmake anymore?? So I can't use it in my project..

## Expected behavior
This should install the pistache library.

## Context (Environment)

- Pistache version: newest git clone
- Operating System: Ubuntu 20.04 and Debian 10 (both with the same problem)
- Architecture: amd64
- Installed from: Built from source
- Used `pkg-config`: no

Tachi107:
Fixes #877

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,7 +151,7 @@ cmake_minimum_required(VERSION 3.6)
 project("MyPistacheProject")
 
 find_package(PkgConfig)
-pkg_check_modules(Pistache REQUIRED IMPORTED_TARGET pistache)
+pkg_check_modules(Pistache REQUIRED IMPORTED_TARGET libpistache)
 
 add_executable(${PROJECT_NAME} main.cpp)
 target_link_libraries(${PROJECT_NAME} PkgConfig::Pistache)
--- a/README.md
+++ b/README.md
@@ -147,14 +147,14 @@ dependencies: dependency('pistache', fallback ['pistache', 'pistache_static_dep'
 To use with a CMake build environment, use the [FindPkgConfig](https://cmake.org/cmake/help/latest/module/FindPkgConfig.html) module. Here is an example:
 
 ```cmake
-cmake_minimum_required(3.4 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.6)
 project("MyPistacheProject")
 
-# Find the library.
-find_package(Pistache 0.0.2 REQUIRED)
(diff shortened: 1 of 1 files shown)


kiplingw (approved):

kiplingw:
Thanks @Tachi107. I take it the Meson `pkg-config` instructions are correct too?

Tachi107:
Yes, Meson uses pkg-config by default

kiplingw:
Ok, you just made me like Meson a little bit better.

PR #889, opened 2021-04-09
Move pkg-config private libraries to Libs.private

Tachi107:
Previously private libraries needed by Pistache were declared as "public" libraries, even if this isn't needed. Now private libraries are properly set in `Libs.private`. 

This could improve linking speed for users, but this is more about correctness that performance. 

The `.pc` file is properly generated, but apart from that I didn't test anything, mainly because this is a really small change and shouldn't cause any issues.

I've also bumped the git version.


kiplingw:
Good eye Andrea.

PR #890, opened 2021-04-09
Meson improvements

Tachi107:
This PR focuses on fixing how build targets were defined and on generating pkg-config files using Meson's [`pkgconfig`](https://mesonbuild.com/Pkgconfig-module.html) module. 

Previously both the static and the shared libraries were always defined, causing Meson to compile Pistache twice. Now the library build target is defined using a single [`library()`](https://mesonbuild.com/Reference-manual.html#library) call, that behaves like a shared or static library (or even both) depending on the value of [`default_library`](https://mesonbuild.com/Builtin-options.html).

The generation of the pkg-config file is also changed. Instead of creating a `.pc.in` file and perform substitutions on it, Meson now uses its own `pkgconfig` generator, that automatically sets the appropriate values of `Libs`, `Cflags`, etc, making maintenance easier. Here's a comparison of the hand-written and generated `.pc` file:

<table>
<tr>
<th>Manual .pc</th>
<th>Generated .pc</th>
</tr>
<tr>
<td>

    prefix=/usr
    exec_prefix=/usr/bin
    libdir=/usr/lib/x86_64-linux-gnu
    includedir=/usr/include
    Name: libpistache
    URL: http://pistache.io/
    Description: An elegant C++ REST framework.
    Version: 0.0.002-git20210107
    Requires:
    Libs: -L'${libdir}/' -lpistache
    Libs.private: -pthread -lssl -lcrypto
    Cflags: -I'${includedir}/'

</td>
<td>

    prefix=/usr
    libdir=${prefix}/lib/x86_64-linux-gnu
    includedir=${prefix}/include
    Name: Pistache
    Description: An elegant C++ REST framework.
    URL: http://pistache.io
    Version: 0.0.002-git20210409
    Requires.private: RapidJSON, openssl
    Libs: -L${libdir} -lpistache
    Libs.private: -pthread
    Cflags: -I${includedir}

</td>
</tr>
</table>

Instead of using `Libs.private` to set OpenSSL flags it used the `Requires` keyword, that is the [recommended](https://people.freedesktop.org/~dbn/pkg-config-guide.html#writing) method. It also set RapidJSON as an additional dependency, a thing that I forgot to do in bf53fb7946bfcbe6a9615a58dc4b1c866d4b0fde. Practical example of how this would help with maintainability ;)

Everything works fine under `autopkgtest`


kiplingw:
Well done @Tachi107.

Tachi107:
@kiplingw I just noticed that I forgot to remove a call to CheckAtomic in CMakeLists.txt, I pushed the commit but you had already merged the branch, you've been super fast haha. Could you please remove those two lines so that I don't have to open another PR? You can find the commit here: ea913d91447eaa5a047b558d6f66c432a1c9a3d5

kiplingw:
What I've done is reverted your commit. If you can prepare a new PR, or some how continue with this one, that would be helpful. I have unfortunately a hundred thousand things on my plate right now.

Tachi107:
Alright, I'll do that tomorrow morning

PR #900, opened 2021-04-12
Make Travis less verbose

Tachi107:
I find it very difficult to navigate Travis CI logs, since all the tests output gets printed. With this change `meson test` and `cmake --target test` will print test logs only when they fail, reducing eccessive noise.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -250,7 +250,7 @@ before_script:
   # Print debug system information
   - cat /proc/sys/kernel/core_pattern
   - cat /etc/default/apport || true
-  - systemctl list-units --type=service --state=running
+  - systemctl list-units --type=service --state=running || true
 
   # Meson debug build
   - meson setup meson_build_debug
--- a/.travis.yml
+++ b/.travis.yml
@@ -250,8 +250,7 @@ before_script:
   # Print debug system information
   - cat /proc/sys/kernel/core_pattern
   - cat /etc/default/apport || true
-  - service --status-all || true
-  - initctl list || true
+  - systemctl list-units --type=service --state=running
 
   # Meson debug build
   - meson setup meson_build_debug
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -250,8 +250,7 @@ before_script:
   # Print debug system information
   - cat /proc/sys/kernel/core_pattern
   - cat /etc/default/apport || true
-  - service --status-all || true
-  - initctl list || true
+  - systemctl list-units --type=service --state=running
```
kiplingw, line 253: Not sure if this was intentional, but you removed the `true` which will bail if `systemctl(1)` bails.

CMakeLists.txt
```diff
@@ -13,7 +13,7 @@ project (pistache
 
 include(GNUInstallDirs)
 
-add_compile_options(-Wall -Wconversion -pedantic -Wextra -Wno-missing-field-initializers)
+add_compile_options(-Wall -Wextra -Wpedantic -Wconversion -Wno-sign-conversion -Wno-missing-field-initializers)
```
kiplingw, line 16: Disabling sign-conversion could be dangerous. Do you have a good reason for this, other than to reduce build log verbosity?

kiplingw (changes requested):

CMakeLists.txt
Tachi107: Yes. Have you ever tried compiling with Clang? Pistache does a lot of sign conversions (like `long - size_t`), and Clang throws hundreds of warnings, making it impossible to spot the important ones. 

Maybe it would be better to disable that warning only on Clang.

.travis.yml
Tachi107: Thanks for spotting this, I'll change it now.

CMakeLists.txt
Tachi107: Reading the [GCC docs](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html) it turns out that _Warnings about conversions between signed and unsigned integers are disabled by default in C++ unless -Wsign-conversion is explicitly enabled_, so this changes just makes sure that both GCC and Clang follow the same behavior. 

As for Meson, I have applied the same option in the [C++17 PR](https://github.com/pistacheio/pistache/pull/897/files?w=1#diff-30d8f6be6320feeacf686be94f48c70869b52630e01ea625f0f15adc0d57c3e4R20).

kiplingw: I think it would probably be better if we don't suppress potentially useful warnings and instead and take care of the type conversions with `static_cast`, more appropriate types, or whatever the case may be.

kiplingw: With the Autotools, and there may be some way of doing this with Meson, but the user can always override the compiler flags during configure time by setting CXXFLAGS which would be more appropriate for your situation where you are drowning in potentially innocuous warnings.

Tachi107: I don't think that placing hundreds of explicit casts would be better than allowing implicit casts. Also, if GCC permits it there are a lot of good reasons (as they are god-level programmers). 

Unfortunately I don't have the time to investigate if any of the sign conversion warnings are actually useful, but since most of them look like "`implicit conversion changes signedness: 'ssize_t' (aka 'long') to 'size_t'`" I doubt that

PR #903, opened 2021-04-15
Remove custom format target in Meson

Tachi107:
I've just found out that Meson automatically sets up a clang-format target when clang-format is found on the system, so there's no need to manually add that target.

It can be ran with `ninja clang-format`.

Details about this Meson feature can be found [here](https://mesonbuild.com/Code-formatting.html#clangformat)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -228,7 +228,7 @@ before_install:
   # Check source code formatting
   - if [ $TRAVIS_PULL_REQUEST != "false" ]; then changed_src=$(git diff --name-only master...HEAD | egrep "\.(h|cc)$" | grep -v "include/pistache/thirdparty" || [ $? == 1 ]); fi
   - if [ ! -z "$changed_src" ]; then git-clang-format-10 --quiet --binary $(which clang-format-10) --diff master HEAD -- $changed_src > ./clang-format-diff; fi
-  - if [ -s ./clang-format-diff ]; then cat ./clang-format-diff && echo "Format source code according to .clang-format rules. Make sure to run make format" && false; fi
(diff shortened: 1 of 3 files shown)


kiplingw:
Is it automatically run when `check` target is used? @oktal this is probably one you should review, given what you've been working on recently.

Tachi107:
I don't think so (nobody has ever defined a `check` target in Pistache, there's only `test`), but the manual `format` target didn't either. The only thing that this PR changes is that instead of manually defining a format target we let Meson define it for us, but they behave the same

kiplingw:
There should be `check` target, even if it's an alias for `test`. The reason being is the former is an [industry standard](https://www.gnu.org/prep/standards/html_node/Standard-Targets.html). As for the `format` target, this is @oktal's area and I'll leave him to review when he has time. Thanks @Tachi107.

Tachi107:
I don't think that aliasing `check` to `test` would be possible. Both Meson and CMake create the `test` target implicitly, and there's no way (that I know of) to reference it to create an alias

kiplingw:
You cannot create a phony target? I find that really hard to believe. That's such a common task with Automake and makefiles in general. Some smaller projects even have more phony targets than they do real ones that create artifacts.

Tachi107:
You can, but `test` is a special target

Tachi107:
Anyway, I think that we could enforce formatting using a pre-commit git hook. We just need to move the format script into `.git/hooks/pre-commit` ([source](https://github.com/andrewseidl/githook-clang-format)). 

This way, we don't even have to set up a format target, since it would be run automatically on every commit

Tachi107:
What's holding this back from being merged?

kiplingw:
Did you add a phony `check` target? It should depend on the `formatcheck` hook.

Tachi107:
Nope, but this PR is unrelated

PR #906, opened 2021-04-16
Add pre-commit format hook

Tachi107:
When a user runs `meson setup` (the configure step that precedes the compilation step) Meson tells git to set the git hooks directory to `.hooks`, so that when a user runs `git commit` git automatically runs `clang-format` and `git add` on all the modified C++ files.

`tools/format.sh` can be now removed, together with the format target


Tachi107:
This should run only when you run `git commit`, so it doesn't affect unstaged changes.
I'm not sure though, I should try

Tachi107:
Sorry, maybe I wasn't clear. This hook does not (in theory) add any file to the commit. It just formats the committed files (it lists the files to be committed, selects just C++ files, formats them, and re-adds them to the commit), in a completely transparent manner.

Since formatting is an aesthetic and trivial change I think that people should be bothered by errors and extra steps to follow. But that's just my opinion, and I would like to know what others think, and, more importantly, why

kiplingw:
@Tachi107, the closest thing to a standard on how this is typically done is the user runs a `check` target. In the case of Pistache's build environment it uses `test`, but we should have an alias for the former as well since this is what most new users will first attempt. Your proposed commit hook should simply invoke this.

Tachi107:
So, based on your feedback and my opinions, this hook could do three different things; let's discuss about the reasons, pros and cons about all of them:

#### Automatically format committed changes
- Why: code formatting shouldn't bother contributors, and this would allow transparent formatting on commit
- Pros: easy to implement, minimal distractions
- Cons: the code is changed during the commit (but we are talking about whitespace changes)

#### Check for formatting on commit, abort if code is not formatted
- Why: `git commit` should not modify any files
- Pros: easy to implement, does not modify any file
- Cons: makes contributors perform extra steps that could be automated

#### Run a `check` target on commit
- Why: recommended by the GNU coding standards
- Pros: runs both unit tests and formatting checks on commit
- Cons: harder (impossible?) to implement, drastically slows down `git commit`, annoying if you've already ran unit tests

Here's what I think are the main benefits and cons of every approach, but please correct me if I understood something wrong. To me, it seems that the first and second approaches are the most viable, while the third one is way too disruptive.

I strongly believe that formatting should bother contributors at all

kiplingw:
The first option is a bad idea because nothing should be changing just any code except people. The second option allows the user to make the necessary changes, but should not run all unit tests, only formatting (if I understand you correctly). The third is a problem because it would not allow users to save partial work in progress for a PR while one or more unit tests still remains broken (even if they're not the ones to have made the changes that broke them).

The formatting check should happen during `check` target which is the standard behaviour. That's the whole reason for having that target. Adding a phony `check` target is trivial with Automake. If you can't add one to Meson that's a serious problem. Typical build environments have to have many phony targets, like `update-po` and `installcheck` in order to be useful.

I recommend finding a way to add a phony `check` target, but not having it as a dependency during commit. Only @oktal's `formatcheck` should be run during a commit.

Tachi107:
As I already said, I think it is impossible to alias `check` to `test` in Meson and CMake. What should a `check` target do? Why is it needed?

I still don't understand what's so bad about automatically formatting code on commit (it is not so different from "format on save", a feature that many editors have, especially when working with languages like Dart), but if _everyone_ agrees that only a format check should be run during a commit that's fine.

kiplingw:
If you can't add a phony target, then we unfortunately need to scrap Meson. I do find it hard to believe that this can't be done though. The check target is the standard one that is so ubiquitous, it's not limited to the GCS. Even CMake can add phony targets.

As for a commit hook that modifies the code, it isn't a good idea. It is impossible to predict the infinite number of creative ways this may introduce subtle problems later. Further, it's the programmer's responsibility to format their commit correctly. The formatcheck target simply should validate that it is done correctly and give them the opportunity to correct it.

kiplingw:
Possibly helpful [thread](https://github.com/mesonbuild/meson/issues/2517) for you. Looks like others have lodged the same complaint against Meson. There may be workarounds.

Tachi107:
You can create phony targets (https://github.com/pistacheio/pistache/pull/903#issuecomment-820929750)... We have a `format` target, a `docs` target, and soon a `formatcheck` target. `test` is a special target, and can't be referenced in both Meson and CMake (https://github.com/pistacheio/pistache/pull/903#issuecomment-820692799).

What should a check target do in Pistache? Why is it needed?

kiplingw:
The check target does exactly what the previously cited GCS [says](https://www.gnu.org/prep/standards/html_node/Standard-Targets.html) it does:

__Perform self-tests (if any). The user must build the program before running the tests, but need not install the program; you should write the self-tests so that they work when the program is built but not installed.__

This is a standard target that is present in projects that don't use the GCS and ones that do. Why Meson's developers decided to depart from that convention that's been adopted for decades is beyond me. Pretty much every project I've ever dealt with from ffmpeg, libav, and everything else has it.

In the case of Pistache, all in tree unit tests should be built and run. @oktal's `formatcheck` phony target should be a dependency so that it can exit with a non-zero shell exit code. If people want to continue using `test`, that's fine. But there should be an alias for the industry standard `check` target that users and downstream distros generally expect. 

I'm not sure why this is so difficult to implement. If Meson has a built-in `test` target, then you don't need to modify it. Leave it as is, but add a `check` alias to it. And if it is too difficult to even implement that, then this is what John had [warned](https://en.wikipedia.org/wiki/GNU_Autotools#Responses_to_Criticism) you about a few weeks ago.

Tachi107:
Please stop talking only about Meson, I spent a lot of time writing Meson support and it is not a nice thing to do. Moreover this is not an "issue" that only Meson has; as I already said multiple times, both CMake and Meson have been designed this way. Autotools has `check`, the rest of the world (Meson, CMake, build2...) has `test`

I'm putting "issue" in quotes because this isn't really an issue, and here's why. 
We have to make sure that every commit/PR is tested on multiple architectures → there's Travis for that
We have to make sure that code is formatted when committed to the repo → there's the `git commit` hook for that

Having `formatcheck` as a dependency of `test` would be superfluous, and, in my opinion, bad. Let's see why.

Let's say I'm writing a super big fix to the HTTP client and my coding style is different from the one that the project uses. I write 245 lines of code, I run the unit test (`test` target), and they fail. After a few minutes I realize that the unit tests actually succeeded, but I didn't formatted the code. Fine, let's run `ninja clang-format`, and re-run the unit tests. But then I decide to change a few lines of code, and I re-run `test`. Oh no! I forgot to format the code. I have to abort the tests, re-format, re-run...

This is to say that the format check shouldn't be run during _code development_, because it can be annoying and doesn't serve any purpose. Instead, it should be run in _code deployment_, that is when I'm submitting the code to the main project, when running `git commit`

Tachi107:
After experimenting a bit with the automatic formatting hook I realised that it could cause some inconsistency issues. Thanks for pointing it out, @dennisjenkins75 and @kiplingw :)

kiplingw:
> Please stop talking only about Meson, I spent a lot of time writing Meson support and it is not a nice thing to do. 

I'm sorry @Tachi107, but if you are going to take criticism of a development tool personally, you are going to have a hard time in your new career. More importantly, when you originally proposed transitioning to Meson, we told you that you had to ensure that it provided the baseline functionality of CMake. I also told you to study the GNU Autotools to understand why certain decisions had been made over the span of 27 years, regardless of whether we actually use it or not.

> Moreover this is not an "issue" that only Meson has; as I already said multiple times, both CMake and Meson have been designed this way. 

Yes, and I also told you that CMake wasn't my first choice either. Both CMake and Meson were born years later out of frustration with the GNU Autotools - but by people who, as John explained to you, haven't ever worked on a project that was complex enough to require them to deviate from the default behaviour. They also began the two projects at a time when there were very few resources that actually explained how the GNU Autotools were suppose to be used, which is a fair criticism (then, not now).

> Autotools has `check`, the rest of the world (Meson, CMake, build2...) has `test`

That is not a good argument, especially when the standard target had been around for at least a decade before CMake.

> I'm putting "issue" in quotes because this isn't really an issue, and here's why.
> We have to make sure that every commit/PR is tested on multiple architectures → there's Travis for that

Yes.

> We have to make sure that code is formatted when committed to the repo → there's the `git commit` hook for that
> 
> Having `formatcheck` as a dependency of `test` would be superfluous, and, in my opinion, bad. Let's see why.
> 
> Let's say I'm writing a super big fix to the HTTP client and my coding style is different from the one that the project uses. I write 245 lines of code, I run the unit test (`test` target), and they fail. After a few minutes I realize that the unit tests actually succeeded, but I didn't formatted the code. 

If it took you a few minutes after the tests had completed to determine the failure was a result of code formatting, then there is a problem with the output generated. If it took you a few minutes because you were waiting for the tests to complete, the `formatcheck` target could be run first because it doesn't require any compilation. Further, there's nothing to stop the user from running the target on its own. It doesn't just bail with a failed shell exit code, but is also suppose to produce output of whatever the `format` target determined was wrong with the formatting.

> Fine, let's run `ninja clang-format`, and re-run the unit tests. But then I decide to change a few lines of code, and I re-run `test`. Oh no! I forgot to format the code. I have to abort the tests, re-format, re-run...

That's not any tool's fault. That is the programmer's responsibility to format the code "correctly", but to be given help from the tool to remind them when they haven't done it correctly.

> This is to say that the format check shouldn't be run during _code development_, because it can be annoying and doesn't serve any purpose. 

The purpose is to verify that the code was formatted in a consistent way. Personally I don't use anything like this in any other project I've worked on intimately, but it was a choice Dennis made and the best way to implement it is what he and I have already suggested. The tools should be subordinate to the programmer and provide constructive feedback that code hasn't been formatted properly, but it should not modify files on their own. Neither Lintian nor any other similar tool does that by default.

> Instead, it should be run in _code deployment_, that is when I'm submitting the code to the main project, when running `git commit`

No, that's not a good idea. If you wrote several thousand lines of code and the formatting rules don't like the placement of your commas, line breaks, indentation, or something, it is much better that you find out early rather than at the very end.

> Edit: `check` vs `test` is just a difference in naming, not an issue. Users do not expect a `check` target to run _tests_ unless they have previously used the Autotools, downstream distros don't either (dh knows exactly how to work with Meson)

I'm not going to argue with you on this further Andrea.

> Edit 2: thinking about it, it would be possible to create a `check` target that does what you want. It would have to be a shell script that runs `meson test -C builddir && meson compile -C builddir formatcheck`, but I still think that it wouldn't be ideal, for the reasons described above

You should not need to create a separate shell script simply to do what people have had to do daily in their build environments for decades, create phony targets to update message translation catalogue templates, start system services, perform diagnostics, install only PDFs, perform a distcheck, create only install directories, and so forth.

While well intentioned, there is a lot of practical experience that you are lacking that you will absorb over time. But sometimes decisions made in mature development tools don't make initial sense until you realize later that had they done it differently, it would have made one particular problem easier to solve, but a whole other class of problems would have been created. This is why I recommended reading John's book so you'd understand many of them.

Tachi107:
> I'm sorry @Tachi107, but if you are going to take criticism of a development tool personally, you are going to have a hard time in your new career. 

Sorry, you're right. There are a lot of things (completely unrelated to Pistache) that are putting me under pressure, and when I'm not in a good mood I don't see things the way they are. 

> If you wrote several thousand lines of code and the formatting rules don't like the placement of your commas, line breaks, indentation, or something, it is much better that you find out early rather than at the very end.

Can `clang-format` actually change the behavior of the code? Until now I supposed that it couldn't, but if it can than you're right, and formatting should be part of code development. 

> You should not need to create a separate shell script simply to do what people have had to do daily in their build environments for decades

Sorry, maybe I wasn't clear enough. By shell script I don't mean that the user should run an actual shell script, but that the `check` target should run `meson test && meson compile formatcheck`, without the need of an actual script.

kiplingw:
> Sorry, you're right. There are a lot of things (completely unrelated to Pistache) that are putting me under pressure, and when I'm not in a good mood I don't see things the way they are.

It's alright. If I didn't lift weights and run I would go insane. 

> Can `clang-format` actually change the behavior of the code? Until now I supposed that it couldn't, but if it can than you're right, and formatting should be part of code development.

I have no idea. I just know that you were proposing that something could automatically modify the code to conform to the project's notion of syntactical aesthetics, which Dennis and I don't think is a good idea (the automatic part).

> > You should not need to create a separate shell script simply to do what people have had to do daily in their build environments for decades
> 
> Sorry, maybe I wasn't clear enough. By shell script I don't mean that the user should run an actual shell script, but that the `check` target should run `meson test && meson compile formatcheck`, without the need of an actual script.

I would reverse the order of that conjunction so it short circuits on a failure of formatting, which is less expensive to perform and the most likely thing to fail in any new contribution to the project.

kiplingw:
Hey @Tachi107. It looks like all architectures failed to build on [s390x](https://launchpad.net/~pistache+team/+archive/ubuntu/unstable/+build/21679377/+files/buildlog_ubuntu-focal-s390x.pistache_0.0.002-202106100348~ubuntu20.04.1_BUILDING.txt.gz), [ppc64el](https://launchpad.net/~pistache+team/+archive/ubuntu/unstable/+build/21679376/+files/buildlog_ubuntu-focal-ppc64el.pistache_0.0.002-202106100348~ubuntu20.04.1_BUILDING.txt.gz), [armhf](https://launchpad.net/~pistache+team/+archive/ubuntu/unstable/+build/21679375/+files/buildlog_ubuntu-focal-armhf.pistache_0.0.002-202106100348~ubuntu20.04.1_BUILDING.txt.gz), [amd64](https://launchpad.net/~pistache+team/+archive/ubuntu/unstable/+build/21679373/+files/buildlog_ubuntu-focal-amd64.pistache_0.0.002-202106100348~ubuntu20.04.1_BUILDING.txt.gz), and [arm64](https://launchpad.net/~pistache+team/+archive/ubuntu/unstable/+build/21679379/+files/buildlog_ubuntu-groovy-arm64.pistache_0.0.002-202106100349~ubuntu20.10.1_BUILDING.txt.gz). I'm guessing you forgot to add git to `Build-Depends`. 

A gentle reminder to always test your PRs in `autopkgtest` to mitigate future similar problems.

PR #910, opened 2021-04-21
Added rfc7231 date format support

starjasmine:
Hi everyone,

Thank you for building such a lovely project!

When I was using the HTTP  client part, ``Pistache::exceptionPrinter()` kept issuing a message like..

> An exception occured: Invalid Date format

When the REST server's response arrives.

And I found the RFC 7231 format parser will be helpful. 

So, I added the code as src/common/http_defs.cc looks now.


kiplingw:
Thanks @starjasmine, but it looks like you've added your build tree to your commits. Can you please revert so that your PR only contains what you intended to change?

On a more relevant note, from what I can see of what you meant to change, it looks fine. However you do need to add a unit test and verify in autopkgtest per the readme.

starjasmine:
@kiplingw @dennisjenkins75 OK. I'll do them. Thanks.

PR #926, opened 2021-05-19
Add fuzzer for OSS-Fuzz integration

DavidKorczynski:
This adds a fuzzer and build set up for Pistache to work with OSS-Fuzz. 

Cross-referencing: https://github.com/google/oss-fuzz/pull/5793

@kiplingw


DavidKorczynski:
@Tachi107 - perhaps resume the conversation here? How about proceeding like this:
1) I remove the cmake stuff from this PR
2) We then integration the fuzzer in without integrating into the Meson build 
3) We then then get this one https://github.com/google/oss-fuzz/pull/5793 merged in

Tachi107:
I'm not an expert, so do what you think is best :)
In the point 2, what do you mean by without integrating into the meson build?

DavidKorczynski:
We should be able to integrate the fuzzing into the meson build so I can run a command along the lines of:

```
meson build --with-fuzzers
```
Currently, I first build hte library and then simply build the fuzzer with `clang++`:
https://github.com/google/oss-fuzz/pull/5793/files#diff-975f43592569f8744f683172297c41ecba0ec9b0700c056ce24565e52ee6cd2d

Tachi107:
Got it. I've taken a look at how projects like [dav1d](https://github.com/videolan/dav1d) set up fuzzing with Meson, and it is not hard; I'll do it as soon as I'll have time (around the end of June, but maybe sooner). For the time being directly using Clang if fine to me

DavidKorczynski:
sounds good. Do you want to merge the PR in as it is now (i.e. we could in fact keep the cmake set up as well) or should I clean up the cmake stuff?

Tachi107:
I don't have write permissions, so we'll have to wait for Kip or Dennis. 
Removing the fuzzing targets from CMake without re-adding them to Meson isn't really useful, and keeping them will not harm anyone. I personally don't like having stuff in two places, as CMake should be used only for compatibility, but that's not a problem, that can be cleaned up in the future, so do what you prefer :)

DavidKorczynski:
Sounds good. Then, I would consider this one good to go. Will wait for Kip or Dennis to review!

PR #932, opened 2021-06-05
Use std::string_view

Tachi107:
Splitting #897.
[`include/pistache/view.h`](https://github.com/pistacheio/pistache/blob/27b3c93b8a50fecc5028cd4b4a8b60dcb0e1f46f/include/pistache/view.h) is now unused, and could be removed, even though I would keep it until C++20 (`std::span`)


Tachi107:
I think this is fine, but this method is the only place where I had to do a little more than just replacing `StringView` with `std::string_view`

https://github.com/Tachi107/pistache/blob/27b3c93b8a50fecc5028cd4b4a8b60dcb0e1f46f/src/client/client.cc#L38-L54

PR #936, opened 2021-06-10
Make date.h and cpp-httplib proper dependencies

Tachi107:
They were previously embedded in the source tree, and are now proper subprojects.
I've also moved `rapidjson.h` out of include/pistache/thirdparty since it is part of Pistache - this is a breaking change.

CMake is probably broken, since it can't use Meson's wraps to download dependencies (must use `FetchContent`).

`autopkgtest` is broken as well, since it is not allowed to download dependencies at configure time and neither [`cpp-httplib`](https://github.com/yhirose/cpp-httplib) nor [`Hinnant-date`](https://github.com/HowardHinnant/date) are available in Debian repos.


Tachi107:
Nope, rapidjson.h is not the same rapidjson you're thinking of. It is a serializer that uses rapidjson, but it is indeed written by oktal (take a look at the code).
It has been moved to the thirdparty directory by a contributor some time ago, probably by mistake.

Tachi107:
Mh, I think that this approach would be impossible to adopt when building deb packages. Embedding external dependencies in the generated tarball (using `meson dist --include-subprojects`) is impossible, since the archive would be different from the original source tree, and (as far as I know) `debuild` does not allow it.
@kiplingw, do you have any idea?

kiplingw:
My thoughts are you should avoid including third party libraries in your local source tree. It's a maintainence nightmare, a security risk, and an unnecessary additional complexity. You should always rely on the system package manager whenever possible. If the library isn't Debianized, you should ask yourself why? It's possible no Debian developer has had a chance to do so yet (and maybe it's in their queue). But another reason is it could be far from ready for production use.

What I suggest doing is Debianizing the libraries you want to use (simple libraries aren't hard to do) and then we can add that to the PPA. You can configure your local `autopkgtest` environment to use additional apt repositories, so it will pull any dependencies you want from the PPA that are specified in the `Build-Depends` stanza.

I'd also add that you can allow network access in a DEP-8 test. The standard allows for that. What you are looking for is the `needs-internet` feature. But the designers probably weren't contemplating downloading source code. What they were probably thinking of was where network connectivity was required to do some protocol testing, or something like that.

In short, always avoid the complexity of building from source unless you actually need to (and 99% of the time people don't).

Tachi107:
I agree, but unfortunately those tho libs aren't in Debian, and they are already used by Pistache. cpp-httplib is used for the tests, while hinnant-date is used by the library.

They are currently embedded in the source tree like any other header file; hinnant-date is in include/pistache/third-party/, while cpp-httplib is in tests/. Differently from GoogleTest and RapidJSON, that are clearly separated from the main source being in the subprojects directory, those other two are "hidden", and this has caused them to not be updated in a long time. While this isn't a big deal for cpp-httplib since it is just used for tests, for hinnant-date this in a bit concerning since the library is part of Pistache and shipped to the users, and an outdated version could be buggy or dangerous.

Adding them to the Pistache PPA seems like a good idea for building the deb package, but we can't expect all users to use a third party PPA - maybe they don't even use a Debian-based distro - so we have to keep them available as a fallback in case the headers are not found in the system.

kiplingw:
> re: rapidjson.h. ok
> 
> re: Debian: Not everyone uses Debian, and Debian derivatives are not the only distribution out there. Please avoid doing anything to Pistache that would prevent some non-Debian user from doing a "git clone; git submodules-init; $make-equiv" from building a working Pistache.

Dennis, we both know that's not what I had said, nor implied, and if you actually took the time to show up at our meetings you'd know that.

Tachi107:
I've moved `hinnant-date` and `cpp-httplib` in the subprojects folder, updated them and removed the wrap files. Deb packaging is now working again.

Some header-only libraries have been moved around, and that caused a huge number of changes.

I don't like this solution that much, as it is ugly and it is a workaround for something that we shouldn't really worry about (packaging should be done in a different place), but this is certainly better that mixing first and third party stuff.

Tests and autopkgtest pass

kiplingw:
Thanks guys. @Tachi107, it looks like it is still failing due to git(1) not being in the `Build-Depends` stanza. Also remember that if you want to do anything to access a remote resource over the network, DEP-8 allows you to do that with the `needs-internet` feature discussed above.

Tachi107:
autopkgtest is failing because this branch was created before e47d23e59d3cd2af31a630c2e5ce2350193823bb was merged. After merging this PR in master all tests will pass (that's what I did locally).

Regarding `needs-internet`, this approach does not need it :)

kiplingw:
Got it. Can you rebase and I'll try again?

Tachi107:
How do I do it?

kiplingw:
I think what you need to do is pull from the main branch into your own. [Here](https://www.atlassian.com/git/tutorials/rewriting-history/git-rebase) is a tutorial.

Tachi107:
I guess I did something wrong...

kiplingw:
Thank you sir. All tests are a pass.

PR #943, opened 2021-06-26
Make Pistache REUSE compliant

Tachi107:
Reuse is a specification that aims at making licensing of open source software easier. You basically have to include an SPDX header in all files, and when this isn't possible you create a file named filename.license containing the header. While I find this a bit too much this is also the only way of reliably determining the license and the copyright holder of a certain file.

The use of SPDX tags has been adopted by projects like the Linux Kernel (https://github.com/torvalds/linux/commit/a61127c2130236168321cc76c5a58e15c00ad154), systemd, etc, and I find it way better that including a long copyright notice on top of every file.

I didn't add any email address, but if you'd like to it is completely possible to add them in the `SPDX-FileCopyrightText` tag, between angle brackets.

The specification is available on [reuse.software](https://reuse.software)


kiplingw:
One other minor thing regarding the Debian packaging, the tarball that is being created contains `debian/` which technically it's not suppose to.  That's in the Debian changes set that's applied. That's why lintian is raising a [no-debian-changes](https://lintian.debian.org/tags/no-debian-changes) warning. We should probably clean this up before this becomes an official Debian package.

Tachi107:
As suggested by the warning, the proper solution would be to remove the `debian/` directory from the upstream source and create a new repository where packaging actually happens. This would also help with downstream packaging, since they don't like when upstream directly debianize the source. 

This is unrelated to the PR though, see you on Matrix/IRC :)

kiplingw:
No, that's a solution, but not necessarily the best one. When you generate a tarball, as I've mentioned before, you're not supposed to take a scorched earth approach and just archive the entire directory. This is for a multitude of reasons as I pointed out in Calcote.

PR #946, opened 2021-06-28
Add autopkgtest GitHub Action

Tachi107:
This CI workflow builds and tests every commit and PR using autopkgtest. It builds a Debian Testing image using `autopkgtest-build-qemu` (`autopkgtest-buildvm-ubuntu-cloud` was too slow), builds the source package and test is with `autopkgtest`.

I'm trying to set this up for multiple architectures (see https://github.com/Tachi107/pistache/tree/actions-autopkgtest), but for whatever reason Ubuntu's `autopkgtest-build-qemu` is unable to build images for foreign architectures, as it fails with `"E: Unable to locate package linux-image-arm64"`, while it should be downloading it from https://packages.debian.org/testing/linux-image-arm64. This works fine on my Debian machine.

If you know how to help please let me know.

### DIFF
--- a/.github/workflows/autopkgtest.yaml
+++ b/.github/workflows/autopkgtest.yaml
@@ -1,0 +1,32 @@
+name: autopkgtest
+
+on:
+  [push, pull_request]
+
+defaults:
+  run:
+    shell: sh
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Install dependencies
+      run: |
+        sudo apt -qq update
+        sudo apt -qq install autopkgtest genisoimage qemu-system vmdb2 meson cppcheck debhelper-compat dh-exec libc6-dev libcurl4-openssl-dev libgtest-dev libssl-dev libstdc++6 lintian pkg-config rapidjson-dev valgrind devscripts --no-install-recommends
+
+    - name: Build image
+      run: |
+        sudo autopkgtest-build-qemu --size 12G testing $HOME/autopkgtest-testing.img
+
+    - name: Build Debian package
+      run: |
+        ./debian/rules get-orig-source
+        debuild -S -sa -uc -us
+
+    - name: Run autopkgtest
+      run: autopkgtest ../pistache_*.dsc -- qemu --timeout-reboot=120 $HOME/autopkgtest-testing.img
--- a/.github/workflows/autopkgtest.yaml
+++ b/.github/workflows/autopkgtest.yaml
@@ -1,3 +1,7 @@
+# SPDX-FileCopyrightText: 2021 Andrea Pappacoda <andrea@pappacoda.it>
+#
+# SPDX-License-Identifier: Apache-2.0
+
 name: autopkgtest
 
 on:


.github/workflows/autopkgtest.yaml
```diff
@@ -0,0 +1,36 @@
...
+  test:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Install dependencies
+      run: |
+        sudo apt -qq update
+        sudo apt -qq install autopkgtest genisoimage qemu-system vmdb2 meson cppcheck debhelper-compat dh-exec libc6-dev libcurl4-openssl-dev libgtest-dev libssl-dev libstdc++6 lintian pkg-config rapidjson-dev valgrind devscripts --no-install-recommends
```
kiplingw, line 24: @Tachi107, it looks like some of these build dependencies are for building the library itself. To ensure you don't have to maintain the list in multiple places, try something like this:

```
$ PISTACHE_BUILD_DEPS=$(perl -ne 'next if /^#/; $p=(s/^Build-Depends:\s*/ / or (/^ / and $p)); s/,|\n|\([^)]+\)//mg; print if $p' < debian/control)
$ sudo apt -qq install $PISTACHE_BUILD_DEPS
```

kiplingw:
@pabs3, can you have a look when you get a chance? Much appreciated. @Tachi107, Paul is an official Debian developer and knows everything about the DPM and related.

.github/workflows/autopkgtest.yaml
pabs3: Don't parse debian/control yourself, let apt do that:

```
$ sudo apt -qq build-dep .
```

pabs3:
It seems like a reasonable start, but two things:

When installing build dependencies for a source package, the right way is to ask apt to do it like this:

```
$ sudo apt -qq build-dep .
```

The autopkgtest shouldn't build the source package again, it should use the pre-built source package from the package build workflow.

Tachi107:
1. How does `build-dep .` work? It looks at the `debian/` directory?
2. Isn't this the case? I build the source package with debuild in the "Build Debian package" step and use it in the "Run autopkgtest" step

kiplingw:
@pabs3 is the Archimedes of Debian. Any time I can't figure something out, he nearly always does.

pabs3:
Yes, `apt build-dep .` looks at the `debian/control` file.

Sorry, I had assumed there would another separate workflow for building the packages that the autopkgtest workflow could use the output from, but I see there isn't one of those. I also wasn't aware autopkgtest could build binary packages by itself. Sigh, so many layering violations in Debian stuff.

Tachi107:
Little update: `autopkgtest-build-qemu` is currently unable to build non-x86 images, see this bug report: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=990415

kiplingw:
Good find @Tachi107. I've been having problems with it lately as well in trying to generate images for non-x86 architectures. His suggestion to use the tools from source might actually not be that difficult to implement.

pabs3:
Another option might be to use qemu in system call emulation mode (aka
user mode) rather than full system emulation, which will enable you to
have a non-x86 chroot on x86. No idea if autopkgtest can use that tho.

https://wiki.debian.org/QemuUserEmulation
https://qemu.readthedocs.io/en/latest/user/main.html

-- 
bye,
pabs

https://bonedaddy.net/pabs3/

PR #948, opened 2021-07-01
Add support for multiple archs in autopkgtest CI

Tachi107:
This took **a lot** of time, as `autopkgtest-build-qemu` has been able to build non-x86 for less than a release (version 5.17, unreleased), and it isn't meant to be (easily) used on Ubuntu (and unfortunately GH Actions only support Ubuntu).

This is extremely fragile; adding a CPU core breaks something, adding a MB of RAM breaks something, installing recommended packages breaks something... you get it. But it works. Beautiful, isn't it?

This currently tests amd64, arm64. armhf, i386 and ppc64el. s390x seems unsupported by `autopkgtest-build-qemu`.

If you're curious to see how many trials it took, have a look at this test branch: [Tachi107/pistache/tree/actions-autopkgtest-multiarch-test](https://github.com/Tachi107/pistache/tree/actions-autopkgtest-multiarch-test)

### DIFF
--- a/.github/workflows/autopkgtest.yaml
+++ b/.github/workflows/autopkgtest.yaml
@@ -5,8 +5,13 @@
 name: autopkgtest
 
 on:
-  [push, pull_request]
+  push:
+    paths-ignore: pistache.io/**
+  pull_request:
+    paths-ignore: pistache.io/**
 
+# The default Debian shell (dash) is faster than bash at running scripts, and
+# using bash when it is not needed doesn't make sense.
 defaults:
   run:
     shell: sh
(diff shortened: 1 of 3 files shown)


kiplingw:
Well done @Tachi107. That looks like a lot of experimentation and head banging on your desk!

Tachi107:
Yep, I'll have to report a lot of issues upstream

kiplingw:
I was also thinking last night that something that might give you some more experience, help solve this problem, and probably others, was if we added the autopkgtest related packages to the PPA.

kiplingw:
I just pushed their git head into our unstable PPA.

PR #955, opened 2021-07-07
Cleanups

Tachi107:
Nested namespaces in serializer/rapidjson.h and client.cc, reformat all files, add .clang-format-ignore ignoring subprojects, use angled brackets when including gtest.h.

Huge diff because of whitespace changes. If you want to take a look at the diff in GitHub remember to ignore whitespace changes appending [`?w=1`](https://github.com/pistacheio/pistache/pull/955/files?w=1) to the URL.

### DIFF
--- a/.clang-format-ignore
+++ b/.clang-format-ignore
@@ -1,0 +1,5 @@
+# SPDX-FileCopyrightText: 2021 Andrea Pappacoda <andrea@pappacoda.it>
+#
+# SPDX-License-Identifier: CC0-1.0
+
+subprojects/**
--- a/include/pistache/client.h
+++ b/include/pistache/client.h
@@ -132,7 +132,7 @@ namespace Pistache::Http
     public:
         ConnectionPool() = default;
 
-        void init(size_t maxConnsPerHost, size_t maxResponseSize);
+        void init(size_t maxConnectionsPerHost, size_t maxResponseSize);
 
         std::shared_ptr<Connection> pickConnection(const std::string& domain);
         static void releaseConnection(const std::shared_ptr<Connection>& connection);
--- a/include/pistache/serializer/rapidjson.h
+++ b/include/pistache/serializer/rapidjson.h
@@ -18,238 +18,232 @@
 #include <pistache/http_defs.h>
 #include <pistache/mime.h>
 
-namespace Pistache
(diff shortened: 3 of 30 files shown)


kiplingw:
@Tachi107, is there some kind of paid upgrade we can obtain (my company can pay for it) so our CI hosting isn't so shitty? The timeouts are happening with every PR now and even the ones that pass take far too long. If you want to look into it and get back to me, I will get it if the cost is reasonable.

kiplingw:
Sure, but isn't that how it's already setup? I always try to check the results before merging any PRs.

Tachi107:
> The timeouts are happening with every PR now and even the ones that pass take far too long.

Those aren't timeouts, as GitHub Actions run for 6 hours before timing out. They randomly fail because autopkgtest is flaky, but I'm looking for a solution. They take so long because the different architectures are emulated with Qemu, and especially arm64 images are quite slow.

I'm trying to find a way to ignore those random CI failures, and I think I found it. Tomorrow I'll submit a PR.

> If you want to look into it and get back to me, I will get it if the cost is reasonable.

I don't really like the idea of letting you pay for CI runners; it is a long term investment and it wouldn't be right to have just you to pay it. Of course I cannot stop you, but you shouldn't feel obligated in any way

kiplingw:
It's alright. I am heavily dependent on Pistache for a commercial product. While I'm not the only one, I don't feel bad contributing financially where reasonable. I also wouldn't complain if we had other sponsorship.

I suspect that after we are accepted downstream we will probably be inundated with increased interest and PRs where having better infrastructure wouldn't be a bad thing.

Tachi107:
Well, I hope so :)
For the time being though I think that waiting a two hours is fine, since the number of PRs is not that high. If this changes I agree with you and we'll have to find a better solution

kiplingw:
Even the two hours, as annoying as it is, is tolerable. But the timeouts could pose an eventual danger when they prevent us from catching a true negative test result that never had a chance to run.

Tachi107:
> But the timeouts could pose an eventual danger when they prevent us from catching a true negative test result that never had a chance to run.

Preparing a fix here: https://github.com/Tachi107/pistache-orig/commit/88733d007668f9ac4a52ef91e6dae9b2c01684cf

PR #957, opened 2021-07-07
Move the Http client to `Pistache::Experimental`

### LINKED ISSUE #952: Move the Http client to the `Pistache::experimental` namespace
Does not work well, and we keep getting issue reports. Until someone fixes it, it should be moved to that namespace. (agreed during the last meeting)

Tachi107:
Closes #952. To be merged after #955


Tachi107:
The diff is big because the `experimental-client` branch is based on `2021-07-07-cleanups`. The actual new commit is 10e205e89a9ae18e2153bde82699537026c6223e

PR #965, opened 2021-08-05
include/pistache/typeid.h: include stddef.h

ffontaine:
Include `stddef.h` to avoid the following build failure with gcc 11:

```
In file included from /tmp/instance-3/output-1/build/pistache-f2f5a50fbfb5b8ef6cf1d3d2a9d442a8270e375d/src/../include/pistache/async.h:10,
                 from /tmp/instance-3/output-1/build/pistache-f2f5a50fbfb5b8ef6cf1d3d2a9d442a8270e375d/src/../include/pistache/client.h:9,
                 from /tmp/instance-3/output-1/build/pistache-f2f5a50fbfb5b8ef6cf1d3d2a9d442a8270e375d/src/client/client.cc:7:
/tmp/instance-3/output-1/build/pistache-f2f5a50fbfb5b8ef6cf1d3d2a9d442a8270e375d/src/../include/pistache/typeid.h:26:12: error: expected type-specifier before 'size_t'
   26 |   operator size_t() const { return reinterpret_cast<size_t>(id_); }
      |            ^~~~~~
```

Fixes:
 - http://autobuild.buildroot.org/results/2443559df8c2357476e4cbdbebb08280cbb80a3b

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/include/pistache/typeid.h
+++ b/include/pistache/typeid.h
@@ -7,6 +7,7 @@
 
 #pragma once
 
+#include <cstddef>
 #include <functional>
 
 namespace Pistache


include/pistache/typeid.h
```diff
@@ -8,6 +8,7 @@
 #pragma once
 
 #include <functional>
+#include <stddef.h>
```
kiplingw, line 11: Can you change this to cstddef? Or is that part of the problem?

kiplingw (changes requested):

include/pistache/typeid.h
Tachi107: If I remember correctly cstddef contains only std::size_t (by the standard), while the code uses size_t. In practice size_t is almost always included as well, but since Pistache is compiled and used on a variety of platforms with different stdlibs it should be better to follow the standard closely

kiplingw: I think another option would be to err on the size of the standard and use cstddef, and then correct `size_t` to be prefixed with the `std` name space in any code that breaks later.

PR #968, opened 2021-08-10
also check for linking with libatomic

jurij-robba:
On some operating systems (like Raspberry Pi OS) with libatomic installed meson configuration succeeds when it shouldn't. Actually building fails on the linking phase because of missing methods in 32 bit version of libatomic.

Improved meson checks added to avoid confusion (and false hope).

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -49,11 +49,11 @@ int main() {
 	return ++x;
 }
 '''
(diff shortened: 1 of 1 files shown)


Tachi107:
Looks fine to me 👍

kiplingw:
@Tachi107, did all our CI testing disappear?

Tachi107:
Thanks @jurij-robba :)

PR #970, opened 2021-08-27
[SERVER] ssl: Add the possibility for the user to set a password callback

Ne02ptzero:
In order to use passphrase-protected keys.
This patch does not change the default behavior:

    server.useSSL("./cert/server.crt", "./cert/server.key");
    $> ./a.out
    Enter PEM pass phrase:
    [...]

    static int password_callback(char *buf, int size, int rwflag, void *u)
    {
        static const char *password = "foobar"; // _Please_ don't do that, that's an example
        strncpy(buf, password, size);
        return strlen(password);
    }

    server.useSSL("./cert/server.crt", "./cert/server.key", false, &password_callback);
    $> ./a.out
    Listening on 0.0.0.0:9080

Key has been generated with:

    $> openssl genrsa -aes128 -passout pass:foobar -out server.key 3072

If you intend to use this feature, please look at the recent
developments on memfd_secret[1] in order to actually store a passphrase
with relative security in memory.

[1] https://lwn.net/Articles/865256/

### DIFF
--- a/include/pistache/endpoint.h
+++ b/include/pistache/endpoint.h
@@ -109,7 +109,8 @@ namespace Pistache::Http
    *
    * \param[in] cert Server certificate path
    * \param[in] key Server key path
-   * \param[in] use_compression Wether or not use compression on the encryption
+   * \param[in] use_compression Whether or not use compression on the encryption
+   * \param[in] cb_password OpenSSL callback for a potential key password. See SSL_CTX_set_default_passwd_cb
    *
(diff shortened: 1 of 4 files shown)


Tachi107:
Looks good to me :+1: 
If someone could merge this it would be great

kiplingw:
@Ne02ptzero, if you or @Tachi107 could add a unit test, we will merge because this is a useful feature.

PR #981, opened 2021-09-15
ci: add conventional commits check

Tachi107:
[Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) is a specification that aims at improving commit messages and history.

It consists of a simple canonical format for commit messages: `type(optional scope): description`

Having a small list of types helps understand when a change is breaking, when it adds a feature, where it patches something, or when it is unrelated to the main library (build system changes, CI, tests...).

This is really useful when determining when to release and what type of release should be made (major, minor, or patch release).

Getting used to it takes a while, but its benefits are real


kiplingw:
@dennisjenkins75?

PR #990, opened 2021-09-28
build(debian): use d/watch to download latest HEAD

Tachi107:
This allows the use of uscan to get the upstream source, a much more elegant approach.

See https://manpages.debian.org/uscan.1.html#direct_access_to_the_git_repository_(HEAD)

I've also fixed a few Lintian warnings, like including the license with the docs, using the wrong section for the package, not including the ITP bug in the first changelog entry and wrong capitalization of Vcs-Browser. The only major violation left is the use of the "libpistache0" package name even if the SONAME is 0.0. I would suggest removing the minor SOVERSION, as ABI is either compatible or incompatible, there's no "minor" incompatibility. More of this in a separate PR.

This should not be merged yet, as I have to update the autopkgtest CI job.

### DIFF
--- a/changelog
+++ b/changelog
@@ -1,5 +1,5 @@
-pistache (0.0.002-pistache1) impish; urgency=medium
+pistache (0.0.002-1) UNRELEASED; urgency=low
 
-  * Latest upstream.
+  * Initial release. Closes: #929593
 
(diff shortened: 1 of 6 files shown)


Tachi107:
This should be ready. I'm not sure if Launchpad will break with this, but I believe it won't.

kiplingw:
Hey @Tachi107. Looks good to me.

PR #993, opened 2021-09-29
ci: add abipkgdiff check

Tachi107:
`abipkgdiff` is a tool capable of comparing two library packages and tell if one of them introduced ABI-breaking changes.

Plain `abidiff` is also available, but since we already build Debian packages I thought it would be more convenient to use pkg version.

### DIFF
--- a/.github/workflows/abipkgdiff.yaml
+++ b/.github/workflows/abipkgdiff.yaml
@@ -1,0 +1,148 @@
+# SPDX-FileCopyrightText: 2021 Andrea Pappacoda <andrea@pappacoda.it>
+#
+# SPDX-License-Identifier: Apache-2.0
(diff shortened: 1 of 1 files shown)


kiplingw:
@pabs3, what do you think?

Tachi107:
I have to thank pabs for this, since he is the one who mentioned abipkgdiff on the #debian-mentors IRC; without him I wouldn't even know that this tool exists :)

kiplingw:
> I have to thank pabs for this, since he is the one who mentioned abipkgdiff on the #debian-mentors IRC; without him I wouldn't even know that this tool exists :)

Genius!

Tachi107:
Ready to be merged :)

Issue #823, opened 2020-09-29
Client - Exception "Invalid Argument" in Reactor module


kocica:
Greetings all,

we are experiencing quite strange problem with our pistache client.
We are receiving "Invalid argument" exception from standard library invoked by a call in reactor.cc, as shown on attached screenshots below.

![Screenshot 2020-09-29 191527](https://user-images.githubusercontent.com/22328693/94594565-e4f57a00-0289-11eb-9e90-9c939973bdbd.png)

![Screenshot 2020-09-29 191550](https://user-images.githubusercontent.com/22328693/94594576-e7f06a80-0289-11eb-879c-27b8709241ae.png)

**This is the scenario:**
Our connection management thread does create and add a new connection in handleConnectionQueue (i.e. connection is in a state EINPROGRESS), then, thread is "frozen" for two about two minutes and then it tries to send a request, which results in `conn->handleError("Could not send request");` method being invoked and exception being thrown as shown on the screenshot above.

I've tried to print stack trace and debug with GDB but can't find a root of the problem whatsoever.
My only idea is to catch the "invalid argument" exception and remove the erroneous connection.

ljluestc:
```

#include <pistache/http.h>
#include <pistache/client.h>
#include <iostream>
#include <exception>
#include <thread>
#include <chrono>

using namespace Pistache;

class ConnectionManager {
public:
    ConnectionManager() : client(Http::Client::create()), shutdownFlag(false) {}

    ~ConnectionManager() {
        shutdown();
    }

    void addConnection(const std::string& host) {
        std::lock_guard<std::mutex> lock(mutex);
        try {
            auto opts = Http::Client::options()
                            .threads(1)
                            .maxConnectionsPerHost(1);
            client->init(opts);

            // Add connection to the queue
            connections.emplace_back(host);
        } catch (const std::exception& e) {
            std::cerr << "Error adding connection: " << e.what() << std::endl;
        }
    }

    void handleConnectionQueue() {
        while (!shutdownFlag) {
            {
                std::lock_guard<std::mutex> lock(mutex);
                for (auto it = connections.begin(); it != connections.end();) {
                    try {
                        // Attempt to send a request
                        auto response = client->get(*it).send();
                        response.then(
                            [](Http::Response rsp) {
                                std::cout << "Response received: " << rsp.code() << std::endl;
                            },
                            [](std::exception_ptr exc) {
                                try {
                                    std::rethrow_exception(exc);
                                } catch (const std::exception& e) {
[...]

PR #995, opened 2021-09-30
ssl: Enable partial write on SSL context

Ne02ptzero:
This patch enables the partial write mode on the SSL server context,
giving us "normal" behavior on non-blocking sockets.

This patch is meant to be a sleep-less alternative to https://github.com/pistacheio/pistache/pull/832,
with the advantage of also fixing a potential issue on file streaming.

Should not break any behavior.

### DIFF
--- a/src/server/listener.cc
+++ b/src/server/listener.cc
@@ -138,6 +138,7 @@ namespace Pistache::Tcp
                 throw std::runtime_error(err);
             }
 
+            SSL_CTX_set_mode(GetSSLContext(ctx), SSL_MODE_ENABLE_PARTIAL_WRITE);
             return ctx;
         }
 


kiplingw:
Bravo @Ne02ptzero. Tis is a far cleaner alternative.

Tachi107:
How is this different from #832?

kiplingw:
@polmr made a good first try. He was sleeping during the write loop. He originally was also using an `assert()` where a `throw` probably would have been more appropriate.

Feel free to add anything @Ne02ptzero.

Ne02ptzero:
The original issue is that OpenSSL tries to write the whole buffer to a socket on `SSL_write`, and since the socket is non-blocking, sometimes the entire buffer is not sent, but rather a small part. This is the expected behavior on `send/write`. It is up to the caller to move a cursor and retry to write some time after.

The original author fixed this issue by while-ing on the write, and sleeping a bit between each call. This is a totally valid solution in itself, but taking the thread hostage with a sleep and while-everything-is-not-written, which can be catastrophic on an event-loop based server.

This patch only enables OpenSSL to *not* write everything, and returns the number of bytes it has written. Since the behavior is the same with `send(2)`, the code to handle this and retry the write when the socket is ready, is already there; it seems to me that it was a better solution.

Tachi107:
Oh ok, thanks for the explanation, makes perfect sense :)

kiplingw:
Unfortunately the fix still doesn't work for me. What's interesting is that if I attempt to the download request from `curl` on the same machine as the server, it works no problem. If I do it from a separate machine on separate networks, it hangs at 98030 bytes received. The server is not bound to localhost, but to a real network interface.

Any suggestions @Ne02ptzero?

kiplingw:
I actually already got a live core dump via `gcore` and examined it. What I found was nothing. None of the threads appeared to be doing anything except polling their fds waiting for the next activity.

kiplingw:
I'll try recompiling the library to reproduce the issue locally, instead of breaking our production server. What's weird is the `send()` returns, the socket isn't closed, and as far as the server is concerned, it's done doing whatever it was that it was supposed to. It's almost like @Ne02ptzero's `SSL_MODE_ENABLE_PARTIAL_WRITE` flag is ignored. Actually, it could well be. Maybe it needs to be set with every new connection.

Ne02ptzero:
Eh.

Sounds like `pistache` is doing everything right, or something very wrong; either we:
- Have a serious issue in the loop that causes a hang or an too-soon socket shutdown
- Or a TCP sync issue with a missing QUICK_ACK or something.

A `tcpdump` of the sending network interface might help, in order to see if the issue is network related.

On a side note, does the fix proposed in #823 works for you?

kiplingw:
I don't see a proposed fix in #823. Or do you mean #832?

What I need to try and do is reproduce this locally on my workstation instead of only on the production server. Then I can not feel guilty about disrupting my teams' work and installing a myriad of development tools on it. Will get back to you guys on what I find.

kiplingw:
I've managed to reproduce it locally. Trying to step into Pistache, but for some reason gdb can't seem to find debugging symbols even though the library is unstripped. Will continue investigating. It's a good thing though that we caught this bug now before we put out a stable 1.0 release.

kiplingw:
@Ne02ptzero, I've found the problem. After the `SSL_write` call in `Transport::sendRawBuffer` in `transport.cc`, I added some logging on its return. This is what I saw in response to my client's request to the server:

SSL_write 16384 of 202032 bytes
SSL_write 16384 of 185648 bytes
SSL_write 16384 of 169264 bytes
SSL_write 16384 of 152880 bytes
SSL_write 16384 of 136496 bytes
SSL_write 16384 of 120112 bytes
SSL_write -1 of 103728 bytes
SSL_get_error: 3
SSL_write -1 of 103728 bytes
SSL_get_error: 1

So something is not handling the `SSL_ERROR_WANT_WRITE` error, which is what the 3 means from `SSL_get_error` is.

kiplingw:
I think I figured it out. `SSL_write` needs to be passed the same data at the same location when the previous call asked to be retried (`EAGAIN`). If you look _really_ carefully at the above, the retry is using a newly allocated buffer at a new address, even if it's the same contents. @Ne02ptzero?

kiplingw:
New fix proposed in #996.

Issue #807, opened 2020-07-28
Sending data on SSL stops at 98210 bytes


softwaremarkdahl:
I have a rest ssl server, Pistache compiled from master branch. It's fully working, but when the response body is bigger than 98210 bytes, the data are not fully sent and the connection is terminated after timeout. This is the verbose of curl:

*   Trying ####...
* TCP_NODELAY set
* Connected to #### port 443 (#0)
* ALPN, offering h2
* ALPN, offering http/1.1
* successfully set certificate verify locations:
*   CAfile: /etc/ssl/certs/ca-certificates.crt
  CApath: /etc/ssl/certs
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
* TLSv1.3 (IN), TLS handshake, Server hello (2):
* TLSv1.2 (IN), TLS handshake, Certificate (11):
* TLSv1.2 (IN), TLS handshake, Server key exchange (12):
* TLSv1.2 (IN), TLS handshake, Server finished (14):
* TLSv1.2 (OUT), TLS handshake, Client key exchange (16):
* TLSv1.2 (OUT), TLS change cipher, Client hello (1):
* TLSv1.2 (OUT), TLS handshake, Finished (20):
* TLSv1.2 (IN), TLS handshake, Finished (20):
* SSL connection using TLSv1.2 / ECDHE-RSA-AES256-GCM-SHA384
* ALPN, server accepted to use http/1.1
* Server certificate:
*  ####
> GET /api/development/#### HTTP/1.1
> Host: ####
> User-Agent: curl/7.58.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< Server: nginx/1.14.0 (Ubuntu)
< Date: Tue, 28 Jul 2020 15:07:40 GMT
< Content-Length: 196693
< Connection: keep-alive
< Access-Control-Allow-Origin: *
< 
{
   "error" : null,
   "result" : [
      {
         "close" : 1.17516,
         "high" : 1.17516,
         "low" : 1.17516,
         "open" : 1.17516,
         "timestamp" : 1595883720000
      },
 #### many other lines ####
      {
         "close" : 1.1731100000000001,
         "high" : 1.1733199999999999,
         "low" : 1.1731100000000001,
         "open" : 1.1733199999999999,
         "timestamp" : 1595916900000
      },
      {
* TLSv1.2 (IN), TLS alert, Client hello (1):
* transfer closed with 98483 bytes remaining to read
* stopped the pause stream!
[...]

PR #996, opened 2021-10-02
fix(ssl): enable ssl mode for moving buffers

Ne02ptzero:
Due to internals of pistache, a write after a retry may be with a
different address than the original call. By default, OpenSSL doesn't
like that:

    SSL_MODE_ACCEPT_MOVING_WRITE_BUFFER

        Make it possible to retry SSL_write() with changed buffer location
        (the buffer contents must stay the same). This is not the default to
        avoid the misconception that non-blocking SSL_write() behaves like
        non-blocking write().

This should fix any weird behavior on large file / responses (#807)

### DIFF
--- a/src/server/listener.cc
+++ b/src/server/listener.cc
@@ -139,6 +139,7 @@ namespace Pistache::Tcp
             }
 
             SSL_CTX_set_mode(GetSSLContext(ctx), SSL_MODE_ENABLE_PARTIAL_WRITE);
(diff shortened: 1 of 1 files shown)


Ne02ptzero:
I cannot reproduce the weird behavior with this patch (Stop sending after X bytes), waiting on Kip to confirm that is has indeed solved the issue.

kiplingw:
@Ne02ptzero, it works! One minor thing though, I think we need to call `ERR_clear_error()` before each call to `SSL_write` to clear the error queue. Either I can do this or you can. Let me know.

Issue #985, opened 2021-09-22
Build errors using meson


nbommersbach:
Hello,
I had problems building pistache from source as meson could not find openssl. I solved this by tweaking around with the openssl visibility settings and finally could make it visible to pkg-config. I was then able to successfully generate the build folder using:

`meson setup build \
    --buildtype=release \
    -DPISTACHE_USE_SSL=true \
    -DPISTACHE_BUILD_EXAMPLES=true \
    -DPISTACHE_BUILD_TESTS=true \
    -DPISTACHE_BUILD_DOCS=false \
    --prefix=$PWD/prefix`

Running the next step from that README (`meson compile -C build`) left me with countless errors I cannot identify and solve:

`ninja: Entering directory `build'
[1/138] Compiling C++ object src/libpistache.0.0.dylib.p/common_stream.cc.o
FAILED: src/libpistache.0.0.dylib.p/common_stream.cc.o 
c++ -Isrc/libpistache.0.0.dylib.p -Isrc -I../src -Iinclude -I../include -I../subprojects/rapidjson-1.1.0/include -I../subprojects/hinnant-date/include -I/usr/local/Cellar/openssl@1.1/1.1.1l/include -fcolor-diagnostics -DNDEBUG -Wall -Winvalid-pch -Wnon-virtual-dtor -Wextra -Wpedantic -std=c++17 -O3 -Wconversion -Wno-sign-conversion -Wno-missing-field-initializers -DPISTACHE_USE_SSL -MD -MQ src/libpistache.0.0.dylib.p/common_stream.cc.o -MF src/libpistache.0.0.dylib.p/common_stream.cc.o.d -o src/libpistache.0.0.dylib.p/common_stream.cc.o -c ../src/common/stream.cc
In file included from ../src/common/stream.cc:12:
In file included from ../include/pistache/stream.h:15:
../include/pistache/os.h:55:9: error: unknown type name 'cpu_set_t'
        cpu_set_t toPosix() const;
        ^
1 error generated.
[2/138] Compiling C++ object src/libpistache.0.0.dylib.p/common_net.cc.o
FAILED: src/libpistache.0.0.dylib.p/common_net.cc.o 
[...]

Issue #960, opened 2021-07-15
Documentation doesn't say this is incompatible with Mac OS


Stokestack:
Nowhere do I see a notice that this project won't build on a Mac (compilation fails with "error: unknown type name 'cpu_set_t'", [a reported problem](https://github.com/pistacheio/pistache/issues/760)).

Please add this notice prominently, so developers don't waste time attempting to build this product.

Thanks!

Tachi107:
I agree, in fact this has already been pointed out in https://github.com/pistacheio/pistache/issues/825#issuecomment-871320363.

Will do as soon as I'll get back home :)

Stokestack:
I appreciate that!

Issue #183, opened 2018-01-24
Compilation error on Mac OS: TimerFD


4x10m:
Hi there,

Just to tell you that i stop to try to make pistache working on Mac OS.
After a lot of problems, i have the following error that i can't solve:

	evan@MacBook-Pro-de-****
	  ~/Documents/GitHub/pistache/build   master ●  cmake -G "Unix Makefiles" -DCMAKE_BUILD_TYPE=Release ..
	make
	sudo make install
	-- RapidJSON found. Headers: /usr/local/Cellar/rapidjson/1.1.0/include
	-- Could NOT find GTest (missing: GTEST_LIBRARY GTEST_INCLUDE_DIR GTEST_MAIN_LIBRARY)
	-- Configuring done
	-- Generating done
	-- Build files have been written to: /Users/evan/Documents/GitHub/pistache/build
	[  2%] Building CXX object src/CMakeFiles/pistache.dir/common/description.cc.o
	In file included from /Users/evan/Documents/GitHub/pistache/src/common/description.cc:10:
	In file included from /Users/evan/Documents/GitHub/pistache/include/pistache/description.h:19:
	In file included from /Users/evan/Documents/GitHub/pistache/include/pistache/router.h:12:
	/Users/evan/Documents/GitHub/pistache/include/pistache/http.h:14:10: fatal error: 'sys/timerfd.h' file not
	      found
	#include <sys/timerfd.h>
	         ^~~~~~~~~~~~~~~
	1 error generated.
	make[2]: *** [src/CMakeFiles/pistache.dir/common/description.cc.o] Error 1
	make[1]: *** [src/CMakeFiles/pistache.dir/all] Error 2
	make: *** [all] Error 2

I can give you more log if you want.
Regards

xoac:
Pistache don't work on Mac OS see #1 And from some time there is no update to this repo, even from PR. If u resolved evnet pool problem u can fork and link to your repo this will make easier to help.

4x10m:
OK sorry for duplicating.

I don't know if I solved evnet pool problem. How do i know that ? Do you mean that you believe I have solve this problem ?

By the way, i will fork.

brew list

	autoconf	ffmpeg		grep		libpng		opencv		sphinx-doc
	automake	flex		help2man	libtiff		openexr		sqlite
	bash		gawk		ilmbase		libtool		openssl		tbb
	binutils	gcc		isl		libunistring	openssl@1.1	wget
[...]

Issue #41, opened 2017-01-13
Compile error on macOS Sierra 10.12.2


briviere:
I'm geting the following compile error:

served/pistache/include/os.h:42:5: error: unknown type name
'cpu_set_t'
cpu_set_t toPosix() const;

thanks

brian

oktal:
Hello,

Unfortunately, Pistache does not work on Mac OS yet, see https://github.com/oktal/pistache/issues/1

PR #1009, opened 2021-10-21
allow compiling on darwin

mavrukin:
allow building on 'darwin' and not just on 'linux'

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -17,7 +17,7 @@ project(
 	meson_version: '>=0.50.0'
 )
 
-if host_machine.system() != 'linux'
+if host_machine.system() not in ['linux', 'darwin']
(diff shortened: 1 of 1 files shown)


kiplingw:
Thanks @mavrukin. But my guess is you will probably need to make a lot more changes than this to get the library to build on OS X. Have you tried successfully?

Tachi107:
This is not possible, see #1, #41, #183, #760, #960, #985

mavrukin:
Sadly, doesn't look like it will work

barracuda156:
@Tachi107 This is perhaps a strong statement. It should be possible, just needs to be fixed.

Tachi107:
> @Tachi107 This is perhaps a strong statement. It should be possible, just needs to be fixed.

Yeah of course, nothing is impossible, but _someone_ has to do the work. Patches are always welcome if you want to help :)

PR #1010, opened 2021-10-23
build: use Semantic Versioning

### LINKED ISSUE #953: Tweak versioning scheme
When we'll finally have a 1.0 release the `VERSION_GIT_DATE` would no longer be useful.

Another change I would like to talk about is removing SONAME versions. Having to think about both API and ABI compatibility is not straightforward for everyone (including me), and not strictly necessary. In my opinion it would be easier if the major version number would be used to express both ABI and API compatibility. This is done by projects like OpenSSL, {fmt}, Qt, LLVM...

Edit: this would also resemble [Semantic Versioning](https://semver.org)

Tachi107:
See [semver.org](https://semver.org) for details, but it basically formalizes our decision to not break API nor ABI in minor releases.

A thing I'd like to talk about before merging is when and how we should use the commit date. We currently use it in the pkg-config file and in version.h, but not as part of the shared library. Also, we add the `-git` prefix to the commit version in the pkg-config file, but nowhere else. This is inconsistent. Another issue pointed out in #679 is that having an arbitrary string in the version number can cause issues.

I would propose to always treat the commit date as part of the version number, and getting rid of `-git` in any version string, as it doesn't really give any valuable information to our users (the fact that we use Git as our VCS is an "implementation detail"). In short, we would have a version number that looks like `major.minor.patch.date`.

Closes #953.

Must be merged **after** #1012


Tachi107:
@kiplingw, I've also update the README. I believe this is ready to be merged

README.md
```diff
@@ -38,29 +38,23 @@ Pistache was originally created by Mathieu Stefani (`@octal`). He continues to c
...
 
 If your contribution has modified the interface, you may need to update the major or minor interface versions. Otherwise user applications and build environments will eventually break. This is because they will attempt to link against an incorrect version of the library -- or worse, link correctly but with undefined runtime behaviour.
 
-The major version should be incremented every time a non-backward compatible change occured in the ABI. The minor version should be incremented every time a backward compatible change occurs. This can be done by modifying `version.txt` accordingly.
+The major version should be incremented when you make incompatible API or ABI changes. The minor version should be incremented when you add functionality in a backwards compatible manner. The patch version should be incremented when you make backwards compatible bug fixes. This can be done by modifying `version.txt` accordingly. Also remeber to always update the commit date in the aformentioned file.
```
kiplingw, line 57: Typo for _remeber_.

README.md
```diff
@@ -38,29 +38,23 @@ Pistache was originally created by Mathieu Stefani (`@octal`). He continues to c
...
-### Release Versioning
+### Versioning
 
-Please update `version.txt` accordingly with each unstable or stable release.
-
-### Interface Versioning
-
-The version of the library's public interface (ABI) is not the same as the release version. The interface version is primarily associated with the _external_ interface of the library. Different platforms handle this differently, such as AIX, GNU/Linux, and Solaris.
-
-GNU Libtool abstracts each platform's idiosyncrasies away because it is more portable than using `ar(1)` or `ranlib(1)` directly. However, it is a pain to integrate with CMake so we made do without it by setting the SONAME directly.
```
kiplingw, line 49: Meson unfortunately also doesn't use libtool. To get the equivalent functionality, the user has to do [more work](https://mesonbuild.com/FAQ.html#how-do-i-do-the-equivalent-of-libtools-exportsymbol-and-exportregex). Please restore this paragraph in adapted form.

kiplingw (changes requested):

README.md
Tachi107: Since I don't yet know how Libtool works I don't know what I could write there. Does this make sense?

> GNU Libtool abstracts each platform's idiosyncrasies away because it is more portable than using `ar(1)` or `ranlib(1)` directly. However, it is [not supported in Meson](https://mesonbuild.com/FAQ.html#how-do-i-do-the-equivalent-of-libtools-exportsymbol-and-exportregex) so we made do without it by setting the SONAME directly.

kiplingw: Yup. That's fine.

kiplingw (approved):

kiplingw:
Ok. Will merge after CI completes.

Tachi107:
CI will still fail, as it is failing since #1008, and the issue is probably related to the way the Debian package is built. I'll fix it in a different PR

PR #1012, opened 2021-10-30
d/watch: remove manual `pretty` option

Tachi107:
This was needed to replace the default `~git$date.$hash` suffix with `+git$date.$hash`, because in Debian `+` is greater than `~`, and it was needed to make the PPA happy. This unfortunately requires the d/watch file to be updated every time the version number is bumped.

Dropping the custom version in d/watch and relying on the default is just better, and will also fix CI.

This must be merged **before** #1010, and if this gets merged #1010 must be merged as well

### DIFF
--- a/watch
+++ b/watch
@@ -1,4 +1,4 @@
 version=4
-opts="mode=git, pgpmode=none, pretty=0.0.002+git%cd.%h" \
+opts="mode=git, pgpmode=none" \
 https://github.com/pistacheio/pistache.git \
 HEAD debian uupdate


Tachi107:
Note that this could break the PPA

Tachi107:
Yep, this would break the PPA, in the sense that `0.0.3`**~**`git20211030` is not greater than `0.0.3`**+**`git20211030`, so with these changes the PPA would refuse to upload new package versions because they would be seen as older than what is already present in the archive. To fix this, it would be sufficient to update `version.txt` to `0.0.4.20211030`.

Tachi107:
After experimenting a bit locally:
- Changes to d/watch do not affect the PPA, as it extracts the version from d/changelog and builds the upstream tarball on its own
- d/watch is only used by the autopkgtest and abipkgdiff CIs
- We cannot remove the `pretty` option otherwise the tarball generated by `uscan` will always be at version `0.0~git[...]`
- To make versioning automatic in d/watch we have to use [`pretty=describe`](https://manpages.debian.org/testing/devscripts/uscan.1.en.html#pretty=), but we can't until we tag a version in the repo
- We'll always have to manually bump the version in d/changelog every time we'll tag a new release, because this is how the PPA works
- This PR did not solve anything

kiplingw:
Shall I revert?

Tachi107:
It's not necessary, as it was also broken before. I have to find another solution

Issue #1007, opened 2021-10-20
[BUG] Server crash if ResponseStream client connection is lost


tn-jed:
## Current Behavior
If a client disconnects while server is in the process of responding with a ResponseStream, server will exit / crash with a `received signal SIGPIPE, Broken pipe.`

## Expected behavior
Server should not crash. Perhaps exception should be thrown (so that process doing the streaming can stop what its doing) and connection/context for that peer should be cleaned up?

(Apologies in advance if I've misunderstood the expected usage here and this isn't actually a bug...) 

## Steps to reproduce the bug
The following sets up an artificially long-running response to more easily observe the problem. Connect with a client (`curl` in my case) and get the response started, then kill the client so that the connection is interrupted/closed. 
```
void long_response(const Rest::Request& request, Http::ResponseWriter response){
      auto stream = response.stream(Http::Code::Ok);
      for(int i=0; i<10; i++){
        stream << "Data...\n";
        stream << Http::flush;
        sleep(2);
      }
      stream << Http::ends;
  }
```
Server will crash and exit quietly. `gdb` will show `Thread 2 "pistache-test" received signal SIGPIPE, Broken pipe.`

## Context (Environment)

- Pistache version: 0.0.002-20211002 ff9db0d9439a4411b24541d97a937968f384a4d3
- Operating System:  Ubuntu 20.04.3 LTS
- Architecture: amd64
- Installed from: Built from source
- Used `pkg-config`: no

Screenshots, logs, stacktraces or other relevant information here.
```
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
[New Thread 0x7ffff7923700 (LWP 14658)]
[New Thread 0x7ffff7122700 (LWP 14659)]

Thread 2 "pistache-test" received signal SIGPIPE, Broken pipe.
[Switching to Thread 0x7ffff7923700 (LWP 14658)]
__libc_send (flags=<optimized out>, len=13, buf=0x7ffff7920480, fd=20) at ../sysdeps/unix/sysv/linux/send.c:28
28      ../sysdeps/unix/sysv/linux/send.c: No such file or directory.
(gdb) bt
#0  __libc_send (flags=<optimized out>, len=13, buf=0x7ffff7920480, fd=20) at ../sysdeps/unix/sysv/linux/send.c:28
[...]

PR #1022, opened 2022-01-06
Fix server crash on client disconnection in transport.cc i.e. issue#1007.

amang8662:
Fixed issue #1007 of server crash on client disconnection.
https://github.com/pistacheio/pistache/issues/1007

### DIFF
--- a/src/common/transport.cc
+++ b/src/common/transport.cc
@@ -355,8 +355,9 @@ namespace Pistache::Tcp
         }
         else
         {
-#endif /* PISTACHE_USE_SSL */
-            bytesWritten = ::send(fd, buffer, len, flags);
+#endif /* PISTACHE_USE_SSL */ 
+           //Change to fix Issue#1007. https://github.com/pistacheio/pistache/issues/1007
+            bytesWritten = ::send(fd, buffer, len, flags|MSG_NOSIGNAL);
 #ifdef PISTACHE_USE_SSL
         }
 #endif /* PISTACHE_USE_SSL */


Tachi107:
Hi, thanks for your contribution :)
How/why does this fix #1007?

amang8662:
Hi @Tachi107 , the issue was that send() API used for data transfer generated SIGPIPE signal in case client got disconnected. To resolve it, I have passed a flag MSG_NOSIGNAL to it. Hence, now the API will return failure along with errno EPIPE which is being handled in Transport::asyncWriteImpl function line 304.

Tachi107:
Looks like a good solution to me. Are you aware of any drawbacks? @kiplingw, @dennisjenkins75, if this looks good to you too it would be nice to merge this

tn-jed:
Thanks @amang8662!

A few thoughts...
* I believe it'll be important to also detect that the pipe is broken (check for `errno == EPIPE` I think?) and then stop trying to send and do any cleanup. Otherwise, server will keep attempting and failing on the `send()` - could take a while if it was to be a big transfer.
* Took a quick look and this _seems_ to be the only usage of `::send()` - would be important that this fix is applied to all other instances

kiplingw:
@amang8662, thanks for your PR. Could you add a unit test that demonstrates the old code breaking and verifies that your proposed fix works?

amang8662:
@tn-jed, regarding the points you shared:
1. I checked whether we can stop further calls to send() by indicating user but haven't found any solution as of now. The changes I made ensures that system won't get crashed on disconnection and doesn't seem to impact any other workflows.
2. Changes at other API can be planned in future. Though, I don't think that affects current issue.

@kiplingw the replication steps described by @tn-jed  in #1007 can be considered as a testcase. We can execute and verify the same testcase with and without the fix.

kiplingw:
Thanks, but it would still be preferable with a unit test in the actual tree.

tn-jed:
@kiplingw see pr #1023 for a test case

Tachi107:
It seems that this commit is not fixing the test failure introduced in #1023

kiplingw:
Indeed @tn-jed. Looks like it still needs a bit of work (or possibly the unit test).

jakob-stark:
@Tachi107 on my local machine, this fix makes the test #1023 pass.

tn-jed:
@jakob-stark you're 100% right. I think this check is already done correctly and it's just the `EPIPE` that needs to be taken care of.

Tachi107:
Hi, could you please rebase? The CI had an issue that made it test code in the master branch as opposed to these changes, but c3e740a11d9787ba2bf0d454386de424f2d7febc fixed it

Tachi107:
This PR got closed as #1040 was merged instead

PR #1023, opened 2022-01-07
Add test demonstrating server crash in issue #1007

tn-jed:
Repro for issue #1007

Cause server to crash with SIGINT by forcing a client disconnect in the middle of a ResponseStream


kiplingw (approved):

kiplingw:
@Tachi107, any idea why CI testing is breaking when attempting to clone the repo?

PR #1024, opened 2022-01-07
watch: use manual pretty option

Tachi107:
I tried dropping it in cec95bc171675c908ae1140ee1277e44a42109df, but unfortunately until we tag a release we'll have to manually update the watch file.

I'm not sure this will completely fix all CI issues, but merging this would help me debugging it anyway

### DIFF
--- a/watch
+++ b/watch
@@ -1,4 +1,4 @@
 version=4
-opts="mode=git, pgpmode=none" \
+opts="mode=git, pgpmode=none, pretty=0.0.3+git%cd.%h" \
 https://github.com/pistacheio/pistache.git \
 HEAD debian uupdate


kiplingw:
Thanks @Tachi107. Re-running all CI tests in #1023 now.

kiplingw:
It looks like CI's are raising false negatives of their being no problems. Check out [armhf](https://github.com/pistacheio/pistache/runs/4745149336?check_suite_focus=true) failing, and yet reporting success.

```
autopkgtest [00:32:00]: starting date: 2022-01-08
autopkgtest [00:32:00]: version 5.17
autopkgtest [00:32:00]: host fv-az200-185; command line: /usr/bin/autopkgtest ../pistache_0.0.3+git20211227.d1d2340-0ubuntu1.dsc -- autopkgtest-virt-qemu --dpkg-architecture=armhf --timeout-reboot=120 --ram-size=3072 /home/runner/autopkgtest-testing.img
qemu-system-arm: terminating on signal 15 from pid 64959 (/usr/bin/python3)
<VirtSubproc>: failure: The VM does not start a root shell on ttyS1 or hvc1 already. The only other supported login mechanism is through --user and -*** the guest ttyS0
autopkgtest [00:33:02]: ERROR: testbed failure: cannot send to testbed: [Errno 32] Broken pipe
Warning: autopkgtest failed, but it's not your fault
```

Tachi107:
Yes, I know. But as the message says those failures are not related to the PR, but to autopkgtest itself. When emulating other architectures it can randomly fail :/

kiplingw:
But surely there must be a way to check for that? It defeats the whole purpose of CI if failures are reported as successes.

Tachi107:
As far as I know you can't mark a GitHub job as something like "skipped" or "ignored", so marking a job as ok and issuing a warning is the best we can do :/
Not all jobs fail (amd64 always works), so if you have an all green run that generally means that the commit is ready to be merged, as it has passed at least one autopkgtest run successfully

kiplingw:
Got it. I take it then there's no way to grep through the log for "Warning: autopkgtest failed, but it's not your fault" and set a failure bit somehow.

PR #1040, opened 2022-02-23
Possible solution for BUG#1007

IHni3:
I'm also currently facing the issue described in #1007. I added the MSG_NOSIGNAL Flag as proposed by @amang8662. But there was still the problem, how to get the info if the socket is still alive. So I checked on the documentation (https://linux.die.net/man/3/getsockopt). With the function `getsockopt()` you are able to receive the last error of the socket. In case the pipe breaks, the error is returned through this function. So we can check if the stream is still alive by testing the return value. 

I added the function `isOpen()` which returns true if there isn't an error on the socket and false otherwise.

This is my first PR in any open-source project. So I hope for leniency.

### DIFF
--- a/src/common/transport.cc
+++ b/src/common/transport.cc
@@ -356,7 +356,8 @@ namespace Pistache::Tcp
         else
         {
 #endif /* PISTACHE_USE_SSL */
-            bytesWritten = ::send(fd, buffer, len, flags);
+            //MSG_NOSIGNAL is used to prevent SIGPIPE on client connection termination
+            bytesWritten = ::send(fd, buffer, len, flags | MSG_NOSIGNAL);
 #ifdef PISTACHE_USE_SSL
         }
 #endif /* PISTACHE_USE_SSL */


kiplingw:
Thanks @IHni3 for your attention to detail. Can you also add a unit test?

Tachi107:
It seems that unit tests are failing because the `streaming_test` is still failing with signal `SIGPIPE` (the `rest_swagger_server_test` fail is a known issue that we still have to solve, unrelated to the PR)

Tachi107:
The actual test that is failing should be the one added here: https://github.com/pistacheio/pistache/pull/1023/files

IHni3:
You're right, the test is still failing with SIGPIPE. Locally, I don't face this issue. I will check on it!

IHni3:
https://linux.die.net/man/2/send

EPIPE
The local end has been shut down on a connection oriented socket. In this case the process will also receive a SIGPIPE unless MSG_NOSIGNAL is set.

IHni3:
https://stackoverflow.com/questions/108183/how-to-prevent-sigpipes-or-handle-them-properly

Also interesting, the post of [dvorak](https://stackoverflow.com/users/19235/dvorak). 
He proposes to ignore SIGPIPE with:

```
signal(SIGPIPE, SIG_IGN);
```

But i don't know where the correct place in the project is to add this line.

Tachi107:
Maybe ignoring it is a solution, but why is it triggered in the first place?

IHni3:
I think it is caused by a client side termination of the connection, and the SIGPIPE is default behavior for a Linux system.

"send SIGPIPE on errors on stream oriented sockets when the other end breaks the connection" (https://linux.die.net/man/2/send)

IHni3:
I'm dealing with the signals in my application like this;
```cpp
#ifdef __linux__
static void sigHandler(int sig) {
    switch (sig) {
    case SIGPIPE:
        std::cout << "SIGPIPE received" << std::endl;
        break;
    case SIGINT:
    case SIGQUIT:
    case SIGTERM:
    case SIGHUP:
    default:
        httpEndpoint->shutdown();
        exit(0);
    }
}

static void setUpUnixSignals(std::vector<int> quitSignals) {
    sigset_t blocking_mask;
    sigemptyset(&blocking_mask);
    for (auto sig : quitSignals)
        sigaddset(&blocking_mask, sig);

    struct sigaction sa;
    sa.sa_handler = sigHandler;
    sa.sa_mask = blocking_mask;
    sa.sa_flags = 0;

    for (auto sig : quitSignals)
        sigaction(sig, &sa, nullptr);
}
#endif

int main(int argc, char** argv) {
#ifdef __linux__
    std::vector<int> sigs{SIGQUIT, SIGINT, SIGTERM, SIGHUP, SIGPIPE};
    setUpUnixSignals(sigs);
#endif
...
...
...
```

jakob-stark:
@Tachi107 Excuse me if I step in here, but in my opinion ignoring the signal is not an option for a library. `SIGPIPE` is an appilcation wide signal, that also can occur in situations completely unrelated to socket IO. Ignoring it, or installing a signal handler is very intrusive to the library user. Adding the `MSG_NOSIGNAL` flag to the `send` call is the right choice in my opinion.

Tachi107:
@jakob-stark sorry for my late response, but in general I agree with you. In fact, I'd like to understand why the `SIGPIPE` signal is triggered in the first place- ignoring it is certainly not a clean solution, and as you and @dennisjenkins75 say it will introduce new issues.

jakob-stark:
@Tachi107 In normal operation (after a tcp connection has been established with the client) the server can send data over the tcp connection by calling the `send()` c library function with the socket file descriptor and the data it wants to send. In pistache this happens in [this line](https://github.com/pistacheio/pistache/blob/3ec9d7c4f8b828fdd391550fff81b01e72dd6269/src/common/transport.cc#L359).

Now consider the case where a client forcefully closes the connection. In TCP this is normally done by sending a RST (reset) flag. If the server tries to send data through a socket belonging to such a forcefully closed connection, the `send()` call fails with either a `EPIPE` or a `ECONNRESET` error. That means `send()` returns `-1` and `errno` is set to the corresponding error code. Which of the two errors occurs, depends on potential outstanding data in the kernel output buffer, that has not yet been sent to the wire when the TCP reset happened. So the two errors help to distinguish if there was data lost by the forced connection close. However this is most likely not relevant for pistache.

The posix standard requires that if `EPIPE` occurs, a signal (called `SIGPIPE`) is deliverd to the process. I do not know the exact reason for this behavior, but it is probably an early desing consideration in UNIX (see e.g. [here](https://stackoverflow.com/questions/8369506/why-does-sigpipe-exist)). Anyway the signal does not mean something harmful. It is just an api feature and in fact I believe that most application should ignore it and handle the `EPIPE` and `ECONNRESET` error codes correctly if `send()` returns failure (`-1`).

Ignoring a signal is done by installing a signal handler that simply does nothing. I strongly dislike this for the reasons I mentioned in my previous post. A much better way would be to tell the `send()` function, that we don't want it to create a signal if an `EPIPE` error happens and indeed the api features this possibility. The [linux programming manual](https://man7.org/linux/man-pages/man3/send.3p.html) states:

>MSG_NOSIGNAL (since Linux 2.2)
              Don't generate a SIGPIPE signal if the peer on a stream-
              oriented socket has closed the connection.  The EPIPE
              error is still returned.  This provides similar behavior
              to using [sigaction(2)](https://man7.org/linux/man-pages/man2/sigaction.2.html) to ignore SIGPIPE, but, whereas
              MSG_NOSIGNAL is a per-call feature, ignoring SIGPIPE sets
              a process attribute that affects all threads in the
              process.

By setting the `MSG_NOSIGNAL` flag, we will still receive the `EPIPE` error if a client closes the connection forcefully. But instead of beeing killed by the signal (or having to install an application wide signal handler), we can just abort the current transmission, close the socket, and return to other work.

 This PR (#1040) adds some functionality besides adding the `MSG_NOSIGNAL`. It adds `isOpen()` and `isClosed()` functions to the `Request` class but I can't think of a scenario where this information might be useful. Actually I think pistache already handles the `EPIPE` and `ECONNRESET` errors almost correctly. Therefore I would vote to merge #1022, which just adds the `MSG_NOSIGNAL` flag.

kiplingw:
I'm fine merging as well, but as @dennisjenkins75 already said, the unit tests need to pass. The 64-bit arm build, as an example, is choking on the `streaming_test` test due to a `SIGPIPE`. @IHni3, if you can make the necessary corrections we will get this into mainline.

@Tachi107: Is our unstable PPA broken or is it still building nightlies? The reason I ask is I haven't seen any in a long time, but then we also haven't had a lot of substantive PRs that deal directly with the code in a while.

Tachi107:
@jakob-stark thanks for your detailed explanation :)

I would also like to merge the PR (maybe the one that introduces less changes and only fixes the issue), but the CI tests have to pass. They are currently running in QEMU, but I don't belive that it could cause issues.

@kiplingw the PPA works, but builds keep failing because of the failed tests

kiplingw:
Got it. Thanks for clarifying. Ok we just need to wait then for him to fix the unit test failures and then we should be good.

IHni3:
Sorry for my late Response.

I added the functions isOpen isClosed to use them for Server send Events (SSE). This is the only case i can think of they are necessary.

IHni3:
Currently i dont have time to look into it. Maybe next week.

tn-jed:
I agree with the great comment by @jakob-stark above. Agreed that we just need the `MSG_NOSIGNAL`. #1022 should be sufficient (prefer that we leave out the extra functionality brought in by #1040)

kiplingw:
@IHni3, can you rebase so we can see how it's performing in CI? Thank you.

IHni3:
Yes, I will rebase it. Also, I will move the extra functionality to a different PR.

Tachi107:
Hi, could you please rebase again? The CI had an issue that made it test code in the master branch as opposed to these changes, and that's probably why your PR didn't fix the tests...

Tachi107:
(@kiplingw we should also modify debian/rules so that flaky tests are run again. it should be sufficient to drop `--no-suite=flaky` from override_dh_auto_test)

kiplingw:
> (@kiplingw we should also modify debian/rules so that flaky tests are run again. it should be sufficient to drop `--no-suite=flaky` from override_dh_auto_test)

I agree. I'm fine with that. I'll make the changes now.

kiplingw:
Done.

include/pistache/http.h
```diff
@@ -739,4 +739,4 @@ namespace Pistache
         }
 
     } // namespace Http
-} // namespace Pistache
+} // namespace Pistache
```
muttleyxd, line 742: Is this needed?

src/common/http.cc
```diff
@@ -1181,4 +1181,4 @@ namespace Pistache::Http
         return std::static_pointer_cast<RequestParser>(peer->getData(ParserData));
     }
 
-} // namespace Pistache::Http
+} // namespace Pistache::Http
```
muttleyxd, line 1184: this one too - I don't think we need to remove blank lines from end of a file

muttleyxd (changes requested):

kiplingw:
I'm letting CI re-run now. Provided all tests pass on all architectures, I will merge unless anyone has any outstanding issues they'd still like to see addressed in this PR.

muttleyxd:
> I'm letting CI re-run now. Provided all tests pass on all architectures, I will merge unless anyone has any outstanding issues they'd still like to see addressed in this PR.

These blank lines aren't anything important for me. I wonder if we care about commitlint, cause commit messages won't pass in current form.

kiplingw:
I'm not a stickler for enforcing a commit log format, within reason. As long as it's readable and gives a high level summary of what was changed, it's fine.

Tachi107:
> I'm letting CI re-run now. Provided all tests pass on all architectures, I will merge unless anyone has any outstanding issues they'd still like to see addressed in this PR.

@kiplingw not all the tests will pass, as the swagger test introduced in #1033 is still unfixed

@IHni3 if you could clean up your commit history and restore the newlines deleted in `include/pistache/http.h` and `src/common/http.cc` it would be really nice. While you're at it, you could even make commitlint pass (with a message like "fix: prevent SIGPIPE on client connection termination")

IHni3:
No Problem i will do it later today.

IHni3:
I hope everything is fine now. 👍👍

muttleyxd (approved):

Tachi107 (approved):

Tachi107:
The test is now fixed also in CI; only `rest_swagger_server_test` is still failing, but it's unrelated to this PR:

    Summary of Failures:

    17/25 pistache:flaky / rest_swagger_server_test FAIL             0.22s   killed by signal 6 SIGABRT

    Ok:                 24  
    Expected Fail:      0   
    Fail:               1   
    Unexpected Pass:    0   
    Skipped:            0   
    Timeout:            0

I'm merging this (my first merge :D)

PR #1050, opened 2022-04-16
Possible fix for #1007 (continued)

muttleyxd:
Since #1040 seems stuck (as author didn't have time to look into the tests), here's my attempt at it - I forced curl to try to connect to IPv4 address in streaming test (that's why it was broken for me on Debian VM, it works fine on Arch host both ways), other tests pass, so let's see what happens on CI now.

### DIFF
--- a/include/pistache/http.h
+++ b/include/pistache/http.h
@@ -169,7 +169,10 @@ namespace Pistache
         } // namespace Uri
 
         // Remove when RequestBuilder will be out of namespace Experimental
-        namespace Experimental {class RequestBuilder; }
+        namespace Experimental
+        {
+            class RequestBuilder;
+        }
 
         // 5. Request
         class Request : public Message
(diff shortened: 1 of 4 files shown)


kiplingw:
Nice work @muttleyxd. It looks like you'll need to bump the soname minor version, per [this](https://github.com/pistacheio/pistache#versioning) because your two additional methods don't break the ABI, but only extend it.

dennisjenkins75 (changes requested):
pistacheio is bullt ontop of epoll().  epoll is way more efficient that select(), and epoll() can tell you when your socket was closed.  I'd rather see a solution for pistache using epoll to detect when the socket is closed as part of its reactor loop.

I think that you need to look for examples using `EPOLLHUP`.

muttleyxd:
Closing this - hope #1040 gets merged soon

PR #1052, opened 2022-04-19
build(test): make use of Meson test suites

Tachi107:
Flaky tests can now be skipped with `meson test --no-suite flaky`, and the same applies to network tests.


meson_options.txt
```diff
@@ -3,8 +3,6 @@
 # SPDX-License-Identifier: Apache-2.0
 
 option('PISTACHE_BUILD_TESTS', type: 'boolean', value: false, description: 'build tests alongside the project')
-option('PISTACHE_ENABLE_FLAKY_TESTS', type: 'boolean', value: true, description: 'if tests are built, run ones that are known to be flaky')
-option('PISTACHE_ENABLE_NETWORK_TESTS', type: 'boolean', value: true, description: 'if tests are built, run ones needing network access')
```
kiplingw, line 7: Hey @Tachi107. Is there any reason why you are stripping out `PISTACHE_ENABLE_NETWORK_TESTS`?  This build flag is necessary to disable them when running in CI environments that might not have network connectivity.

Tachi107: Yes. As we now use test suites, we can exclude network tests with `--no-suite=network`. I'm updating Debian build scripts too

kiplingw: Ok, be sure to remove all references to it then. This includes the `README.md` and `debian/rules`. The latter uses it in the `override_dh_auto_configure` override.

kiplingw (approved):

Tachi107:
The tests will likely fail because they'll try to set the `PISTACHE_ENABLE_NETWORK_TESTS` Meson option, as #1053 hasn't been merged yet

PR #1054, opened 2022-04-20
Functions for Server Sent Events added

IHni3:
Like described in #1040. I added `isOpened()` and `isClosed(),` so it is possible to know when the client terminates the connection. I personaly use it for SSE.


tests/streaming_test.cc
```diff
@@ -126,7 +126,8 @@ namespace
 
 // from
 // https://stackoverflow.com/questions/6624667/can-i-use-libcurls-curlopt-writefunction-with-a-c11-lambda-expression#14720398
-auto curl_callback = +[](void* ptr, size_t size, size_t nmemb,
+typedef size_t (*CURL_WRITEFUNCTION_PTR)(void*, size_t, size_t, void*);
+auto curl_callback = [](void* ptr, size_t size, size_t nmemb,
```
Tachi107, line 130: Why is this change needed?

fleon-psi:
Hi,  Option to support SSEs seems to be very useful for my project.
I wanted to ask about status of this fix? Is there something I could do to help in going forward with this one?
Thanks!

kiplingw:
Thanks @fleon-psi for checking in. This PR still needs some more. It needs to at least be rebased and a unit test added.

PR #1058, opened 2022-04-21
WIP: POC: do not merge: investigating what's being built on CI 

muttleyxd:

PR #1060, opened 2022-04-21
ci(autopkgtest): drop uscan, use manual .orig tarball

### LINKED ISSUE #1059: actions: autopkgtest pipeline build latest release instead of code
When working on #1050 I noticed that my changes are not being applied on CI/autopkgtest and it was confusing for me - I had it working on Arch host, Debian VM and even autopkgtest image when I extracted it and built my code there manually.

I made a proof-of-concept PR #1058 to verify my suspicions and it turns out I'm right
I changed `src/client/client.cc` file, just to check what's being built by autopkgtest and what's provided by PR

Log from https://github.com/pistacheio/pistache/runs/6112900509?check_suite_focus=true

```
Run find src/ -name "*.cc" | xargs md5sum
...
0c208d2dec5d5f6a16139464f6e246ee  src/client/client.cc # from my PR
...
657a2dfdbe24c1a2f3a020f767a8fde5  src/client/client.cc # what's in pistache-* directory
```

What would be a good way to fix this? The way that I've got it working, on my test environment, was doing something like this
```
cd pistache
git format-patch origin/master..HEAD
cd ../pistache-*/
patch -p1 <../pistache/0001-* # for each patch
dpkg-source --commit

dpkg-source --build .
```

This works but idk if that's Debian way to do it, I'm not familiar with it, so you may have a better idea to how to fix this.

@kiplingw

Tachi107:
Fixes #1059
Closes #1058

### DIFF
--- a/.github/workflows/autopkgtest.yaml
+++ b/.github/workflows/autopkgtest.yaml
@@ -71,16 +71,20 @@ jobs:
         sudo apt update
         sudo apt autopurge '*yarn*' '*npm*' '*node*'
         sudo rm -rf /usr/local/lib/node_modules
-        sudo apt --assume-yes install autopkgtest/testing genisoimage qemu-system vmdb2/testing devscripts --no-install-recommends qemu-user-static/testing qemu-efi-aarch64 qemu-efi-arm ovmf ovmf-ia32 && sudo apt-mark auto qemu-user-static qemu-efi-aarch64 qemu-efi-arm ovmf ovmf-ia32
+        sudo apt --assume-yes install autopkgtest/testing genisoimage qemu-system vmdb2/testing dpkg-dev --no-install-recommends qemu-user-static/testing qemu-efi-aarch64 qemu-efi-arm ovmf ovmf-ia32 && sudo apt-mark auto qemu-user-static qemu-efi-aarch64 qemu-efi-arm ovmf ovmf-ia32
+
+    - name: Create .orig tarball
+      run: |
+        repo_name=$(echo ${{ github.repository }} | cut -d / -f 2)
+        debian_version=$(dpkg-parsechangelog --show-field Version | cut -d - -f 1)
+        cd ..
+        tar --exclude-vcs --exclude=debian -cvzf pistache_"$debian_version".orig.tar.gz "$repo_name"
+
+    - name: Build Debian package source
+      run: dpkg-source --build .
 
     - name: Build image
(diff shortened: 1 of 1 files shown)


.github/workflows/autopkgtest.yaml
```diff
@@ -34,6 +34,14 @@ jobs:
     - name: Checkout
       uses: actions/checkout@v2
 
+    # The version number needs to be the same as the one in
+    # https://github.com/pistacheio/pistache/blob/debian/changelog
+    - name: Create .orig tarball
+      run: |
+        cd ..
+        repo_name=$(echo ${{ github.repository }} | cut -d / -f 2)
+        tar  --gzip --exclude-vcs -cvf pistache_0.0.003.orig.tar.gz "$repo_name"
```
kiplingw, line 43: Mostly good, but we have a minor problem here. The version probably shouldn't be hardcoded into the tarball name in this file. Instead it should be parsed from `version.txt`. Actually, even better, shouldn't the build environment generate the tarball (e.g. CMake)?

kiplingw (changes requested):

.github/workflows/autopkgtest.yaml
Tachi107: Unfortunately no, as it has to be in sync with what is in the debian directory, regardless of the contents of the "upstream" files.
But thinking about it, I could directly parse the version from the debian directory itself

kiplingw: Yeah, I think that's the way to do it. awk(1) / sed(1) are your friends.

Tachi107: Also, to clarify, neither CMake nor Meson are safe to use to generate the tarball because its contents have to be exactly identical (excluding vcs files) to the ones already present in the checked out directory. So, to make things simple and robust, I simply archive the checked out dir so that I'm sure that I don't have extra files nor missing ones.

kiplingw: Ah yes, I recall now. We talked about this before when we were discussing the GNU Autotools versus other build environments.

Tachi107: > Yeah, I think that's the way to do it. awk(1) / sed(1) are your friends.

Luckily `dpkg-parsechangelog` already exists :D

kiplingw: Your Debian kung fu is growing stronger every day.

PR #1069, opened 2022-05-21
build: link with stdc++fs on old GCC versions

### LINKED ISSUE #1068: Problems when installing pistache
When i install pistache by using cmake, I found that  after I execute the "make install" command, it will occur an error. 
No rule to make target ”install'.  Stop.

Does anyone have this problem?

Tachi107:
Fixes #1068


Tachi107:
Merging as autopkgtest is based on Debian Testing and ships a new version of GCC that doesn't need `-lstdc++fs`

PR #1070, opened 2022-05-21
ci: build on multiple Linux distributions

Tachi107:
This new CI job builds and tests Pistache on different Linux distros, with different compilers, and under different sanitizers, and also re-adds Codecov coverage integration back.

Currently, Pistache is tested on Debian Stable, Debian Testing and Red Hat Enterprise Linux 8, and the code is built with GCC and Clang. Sanitizers are only used on Debian, because as far as I'm aware RHEL doesn't ship the various sanitizers in its base images.

ThreadSanitizer is also disabled for now, as it reports a lot of data races in some tests. Some of the bugs are in the unit tests themselves, but I haven't yet looked into solving them. @dennisjenkins75 I seem to remember that you opened a meta-issue about dace conditions in Pistache but I haven't found it.

~I'll add RHEL 9 as soon as Red Hat publishes the image on [Docker Hub](https://hub.docker.com/r/redhat/ubi9-minimal)~ added using Red Hat's own registry

It would be nice to add something like [Gentoo](https://hub.docker.com/r/gentoo/stage3) too; @dennisjenkins75 you use/used Gentoo, right? Would you be able to help?

~I've also fixed a bug in the [`streaming_test`](https://github.com/pistacheio/pistache/pull/1070/commits/f5aeaa87795e45e8802e134adf9554f91b30aea1) test that caused lock ups on old distros (@kiplingw this means that PPA builds for Ubuntu 20.10 and older should be fixed as well)~ pushed as 8ce07256dc220cbf96b3fd6f5991a3c68fd54aba

Lastly, I ~removed~ [reduced the scope of the autopkgtest](https://github.com/pistacheio/pistache/pull/1070/commits/cc2025c525a55a688dc010bbb025928432f56b15) job; you can find my rationale in the commit message.

PR #1071, opened 2022-05-23
test: rework ClientDisconnect test

Tachi107:
It was using a complex mix of condition variables, mutexes, synchronization stuff that was completely unneeded, and for some reason this made the test lock up when ran on old Linux distros like RHEL 8 or Ubuntu 20.10. This should fix the build failures in the Launchpad PPA.

Cherry-picked from #1070, as it is taking a while to get merged.


kiplingw:
Well done @Tachi107. After all CI passes, you can go ahead and merge.

Issue #908, opened 2021-04-19
[FEATURE REQUEST] (cmake install)


EMHussain:
## Install pistache via cmake.
Please add cmake install support i am successfully able to build with cmake but unable to install.

Tachi107:
Pistache has switched to the Meson build system a while ago; you can find instructions on how to install it from source on pistache.io.

Why do you need to be able to install the library using CMake?

EMHussain:
It want to build for android, I am not much familiar with meson build for android.

Tachi107:
What are you trying to achieve exactly? This isn't really a Pistache issue, since Meson supports cross-compiling, but (re)adding the install target wouldn't be difficult either.

EMHussain:
Ok alright I can re add install targets thanks.

PR #1073, opened 2022-05-26
build(cmake): add install error message

### LINKED ISSUE #1072: Custom cmake target "install" should print a useful error message and direct users to RTFM
Consider crafting a custom cmake target for "install" that would print a useful error message and direct users to the documentation on how to properly install pistachio from sources.

Tachi107:
When running the install target, users will now get the following error:

    $ cmake --install build
    -- Install configuration: "Release"
    CMake Error at build/cmake_install.cmake:46 (message):
      Please use Meson to install Pistache.

      See the README for details:
      https://github.com/pistacheio/pistache#building-from-source

Fixes #1072

(CI skipped as CMake isn't tested)


Tachi107:
Thank you for suggesting the enhancement in the first place! I hope we'll get fewer reports like #1068, #877, and #908

patrick-kings:
Is there a way to disable "Please use Meson to install Pistache" error since my setup works with the previous commit and FetchContent in cmake.

```cmake
FetchContent_Declare(pistache
    GIT_REPOSITORY https://github.com/pistacheio/pistache.git
    GIT_TAG 469bad6edf32d76bab3565d5d1f09b86807ff08c
    GIT_PROGRESS TRUE
    DOWNLOAD_EXTRACT_TIMESTAMP TRUE
)

FetchContent_MakeAvailable(pistache)

target_link_libraries(${PROJECT_NAME} PRIVATE pistache)
```

eli-schwartz:
add_subdirectory still invokes installation, and that's what FetchContent_MakeAvailable essentially is.

You could guard this error message by checking if it is the main project, I guess. :)

PR #1075, opened 2022-05-26
docs: use pistacheio.github.io/pistache/

Tachi107:
The pistache.io domain has been renewed by somebody outside of the Pistache organization, and there's not much we can do about it...


kiplingw:
Good eye @Tachi107. Don't forget `debian/{upstream/metadata,control,copyright}` too.

PR #1077, opened 2022-05-28
fix(mime): ignore locales when parsing qvalues

### LINKED ISSUE #1076: Accept header parsing is locale-dependent, fails for valid values
My default Firefox `Accept` header looks like this: `Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,*/*;q=0.8`.

I suspect it's similar for a lot of people.

Pistache fails to parse the header, because of the `q=0.9` (and `q=0.8` for the matter) quality factor.

Stepping through the debugger, I traced the problem here: https://github.com/pistacheio/pistache/blob/master/src/common/mime.cc#L243

`match_double` internally calls `strtod`, which is locale dependent. My locale has `,` as a decimal separator, which means `strtod` only consumes the "0" of "0.9", leaving the "." as the next token, and thus breaking the parser, resulting in the "Unfinished Media Type parameter".

The solution to this would be parsing the quality factor in a locale-independent way, as seems correct (see https://httpwg.org/specs/rfc7231.html#rfc.section.5.3.1).

To do so, I propose getting inspiration from g++ stdlib here: https://github.com/gcc-mirror/gcc/blob/master/libstdc%2B%2B-v3/src/c%2B%2B17/floating_from_chars.cc#L331, the underlying implementation of C++17's `from_chars` (i.e. temporarily changing the locale before parsing, and restoring it after).

I'd be happy to provide the fix if the solution seems ok.

Tachi107:
Qvalues are now parsed manually in `str_to_qvalue()` instead of using the locale-dependant `std::strtof`.

Using a custom function allowed me to add some more checks to the function; while `std::strtof` was previously accepting all floating point numbers, `str_to_qvalues()` makes sure that the supplied input conforms to the [HTTP standard](https://datatracker.ietf.org/doc/html/rfc7231#section-5.3.1)

A rapid "conformance check" reveals the following results:

    $ cat qvalues
    0
    0.0
    0.00
    0.000
    0.1
    0.12
    0.123
    1
    1.0
    1.00
    1.000
    0.
    00.0
    0.0.0
    .0
    .
    1.
    1.001
    .1
    0.1234
    a
    0.a
    1.a
    a.1
    0.00a
    0.000a

`std::strtof()`:

    $ for value in $(cat qvalues); do
          curl localhost:9080 --header "Accept: text/html;q=$value"
      done
    Hello World
    Hello World
    Hello World
    Hello World
    Hello World
    Hello World
    Hello World
    Hello World
    Hello World
    Hello World
    Hello World
    Hello World
    Hello World
    Unfinished Media Type parameter
    Hello World
    Invalid quality factor
    Hello World
    Hello World
    Hello World
    Hello World
    Invalid quality factor
    Unfinished Media Type parameter
    Unfinished Media Type parameter
    Invalid quality factor
    Unfinished Media Type parameter
    Unfinished Media Type parameter

`str_to_qvalue()`:

    $ for value in $(cat qvalues); do
          curl localhost:9080 --header "Accept: text/html;q=$value"
      done
    Hello World
    Hello World
    Hello World
    Hello World
    Hello World
    Hello World
    Hello World
    Hello World
    Hello World
    Hello World
    Hello World
    Invalid quality factor
    Invalid quality factor
    Invalid quality factor
    Invalid quality factor
    Invalid quality factor
    Invalid quality factor
    Invalid quality factor
    Invalid quality factor
    Invalid quality factor
    Invalid quality factor
    Invalid quality factor
    Invalid quality factor
    Invalid quality factor
    Unfinished Media Type parameter
    Unfinished Media Type parameter

`str_to_qvalue()` is far stricter, making MIME handling as a whole more reliable. It also makes responses more accurate, as in the new approach "Invalid quality factor" is sent when the old one wrongly sends "Unfinished Media Type parameter".

Fixes #1076

Please note that I'm not that familiar with how the Accept header works, so this might be completely wrong!

@rems4e could you confirm that this patch fixes your issue? Thanks

~TODO: tests~


kiplingw:
Looks fine to me. @dennisjenkins75?

src/common/mime.cc
```diff
@@ -15,6 +15,87 @@
...
+ *
+ *     qvalue = ( "0" [ "." 0*3DIGIT ] )
+ *            / ( "1" [ "." 0*3("0") ] )
+ *
+ * https://datatracker.ietf.org/doc/html/rfc7231#section-5.3.1
+ */
+static bool str_to_qvalue(const char* str, float* qvalue, std::size_t* qvalue_len)
+{
+    // ASCII numbers start at index 48
+    constexpr char offset = 48;
```
dennisjenkins75, line 31: Just use '0' instead of the ascii numeric value.

src/common/mime.cc
```diff
@@ -15,6 +15,87 @@
...
+    {
+        return false;
+    }
+
+    *qvalue = 0;
+
+    switch (*qvalue_len)
+    {
+    case 5:
+        *qvalue += (str[4] - offset) / 1000.0F;
```
dennisjenkins75, line 79: This can possibly have odd rounding errors, as its not a perfect binary fraction.

Suggestion: Parse the number as if it were an integer, then convert to float at the very end of the function.

dennisjenkins75 (commented):
Please add unit tests.  In addition to verifying taht `str_to_qvalue()` rejects malformed input, also verify that it correctly parses stuff like:

0.000
0.001
0.002
0.003
0.010
0.124
0.126
0.998
0.999


Pick values that are just above or below a perfect 1/(2^n) for integer n.

Tachi107:
> In addition to verifying taht `str_to_qvalue()` rejects malformed input, also verify that it correctly parses stuff like:

Thanks for mentioning, I can confirm that it does.

> This can possibly have odd rounding errors, as its not a perfect binary fraction.
> 
> Suggestion: Parse the number as if it were an integer, then convert to float at the very end of the function.

I'm not sure I understood you correctly, could you please check if my latest change is ok?

> Please add unit tests

I simply added a call to `std::locale::global(std::locale())` in the mime test to set the user-preferred locale when parsing, and it should be enough as `valid_parsing_test` already checks that `qvalue`s are parsed correctly. In fact, setting the locale while reverting this fix results in failed tests (this is only effective in systems where the preferred locale uses `,` as the decimal separator, but specifying a specific locale may be worse- this makes it possible to run tests with different locales without recompiling)

src/common/mime.cc
```diff
@@ -239,9 +321,14 @@ namespace Pistache::Http::Mime
 
                 if (match_literal('=', cursor))
                 {
-                    double val;
-                    if (!match_double(&val, cursor))
```
rems4e, line 243: The `match_double` function is now unused, so it should probably go, as  I don't see the use for that locale-dependent implementation in the future.

src/common/mime.cc
```diff
@@ -15,6 +15,88 @@
...
+{
+    constexpr char offset = '0';
+
+    *qvalue_len = 0;
+
+    // It is useless to read more than 6 chars, as the maximum allowed
+    // number of digits after the dot is 3, so n.nnn is 5.
+    // The 6th character is read to check if the user specified a qvalue
+    // with too many digits.
+    for (; *qvalue_len < 6; (*qvalue_len)++)
```
rems4e, line 38: Because you're already looping, probably you can parse the number as an integer in the loop as well. I understand that's also what @dennisjenkins75 meant: you simply do the classic `qvalue_int = qvalue_int * 10 + (str[i] - '0')`. Only one loop -> function simplified ; no precision loss involved by dealing with floating points.

src/common/mime.cc
Tachi107: You're right, but `match_double` is also part of the public API, and it may be used by someone already

src/common/mime.cc
Tachi107: Maybe my brain simply doesn't want to collaborate, but could you please explain me how you would do this?

src/common/mime.cc
rems4e: In which case, it's worth mentionning in the doc that it's locale dependent :)

Tachi107:
From my new commit:

I added some tests that verify that the `HttpError` exeption message is the expected one, i.e. "Invalid quality factor". This is done with GoogleTest's new `ThrowsMessage` matchers. Unfortunately they are not fully documented, but most information about them can be found in this comment: https://github.com/google/googletest/issues/952#issuecomment-638968347

This (new?) `EXPECT_THAT` syntax seems to be encouraged by GoogleTest, and I find it quite nice: https://google.github.io/googletest/faq.html#why-does-googletest-support-expect_eqnull-ptr-and-assert_eqnull-ptr-but-not-expect_nenull-ptr-and-assert_nenull-ptr

`EXPECT_THAT` is part of GMock, and the `ThrowsMessage` matcher is only available in GoogleTest 1.11.0 and newer. This means that the `debian/control` file in the `debian` branch needs to be updated.

dennisjenkins75 (approved):

kiplingw (approved):

Issue #1080, opened 2022-06-13
Setting a request attribute


andreastedile:
In the past I have used Java Servlets (now [Jarkarta Servlets](https://jakarta.ee/specifications/servlet/6.0/jakarta-servlet-spec-6.0.html)). With Servlets, I could set a "request attribute". To better explain what a request attribute is, it is easier if I provide a use case.

```
bool auth_middleware(Request &request, ResponseWriter &response) {
    const auto cookies = req.cookies();
    const auto userId = getUserId(&cookies);
    // use a service to retrieve the User entity corresponding to userId
    const std::option<User> user = UserService.getUser(userId);
    if (user.has_value()) {
        request.setAttribute("user", user->get());
        return true; // proceed along the chain
    } else {
        response.send(Code::Bad_Request);
        return false; // reject the request
    }
}
```

When the execution flow reaches the actual handler, one could do as follows:
```
void some_handler(const Rest::Request &request, Http::ResponseWriter response) {
    auto user = request.getAttribute("user");
}
```

The advantage of this is that any handler that needs to do something with the `User` entity does not need to repeat the logic of checking whether the user exists and is authenticated (separation of concerns).

Is it possible to do something like this in Pistache? Thank you! :smile:

Fabio3rs:
I think this is very useful, I did a pull request: #1085

PR #1085, opened 2022-07-24
Add set request attribute #1080

Fabio3rs:
The issue #1080 has some time and I don't saw pull requests, I think this is very useful, so I make some code that do it.

I created two commits:
- the code
- the ninja clang-format that the git commit scripts asks

Suggestions for more tests? Or names?
Thanks!

### DIFF
--- a/include/pistache/endpoint.h
+++ b/include/pistache/endpoint.h
@@ -127,7 +127,7 @@ namespace Pistache::Http
    * [2] https://en.wikipedia.org/wiki/CRIME
    */
         void useSSL(const std::string& cert, const std::string& key,
-            bool use_compression = false, int (*cb_password)(char *, int, int, void *) = NULL);
+                    bool use_compression = false, int (*cb_password)(char*, int, int, void*) = NULL);
 
         /*!
    * \brief Use SSL certificate authentication on this endpoint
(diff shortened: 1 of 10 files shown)


andreastedile:
Hi, I am not a contributor of Pistache, but I can provide the following feedback.

- `getData`, `tryGetData` and `getDataAs`: I would personally condense these functions into one single function returning a `std::optional`. The optional should be parametrized with the type of data I want it to hold (just like what `getDataAs` currently does).
- I prefer `getData` and `setData` instead of `getData` and `setData`. As a user, knowing that there is a `get` function, I would look for a `set` corresponding function rather than a `put` one.
- I am not 100% sure that `removeData` should throw an error if the key does not exist.
- In my opinion, the naming of the functions can be improved to reflect their intent. What we want to do is to set and get the attributes of a request: `setAttribute` and `getAttribute` are intuitive to me, whereas I would have more trouble understanding the intent of `setData` and `getData` (what does "data" refer to? it is too generic).

Fabio3rs:
Hi. Thank you!
I think your observations makes sense, I was trying to make similar to the Pistache::Tcp::Peer code https://github.com/pistacheio/pistache/blob/master/include/pistache/peer.h#L54

I am not sure how to proceed is this case.

PR #1091, opened 2022-08-19
CMake subproject fixes

Tachi107:
In this PR I've made some minor improvements to the CMake files so that Pistache plays more nicely when it is included as a subproject. See the individual commits for details.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,12 +6,6 @@ cmake_minimum_required (VERSION 3.11)
 
 set(CMAKE_BUILD_TYPE_INIT Release)
 
-# CMAKE Pin cxx compiler to CXX17
-set(CMAKE_CXX_STANDARD 17)
-
-set(CMAKE_CXX_STANDARD_REQUIRED ON)
-set(CMAKE_CXX_EXTENSIONS OFF)
-
 file(READ "version.txt" VERSION_FILE_RAW)
 string(STRIP "${VERSION_FILE_RAW}" VERSION_FILE)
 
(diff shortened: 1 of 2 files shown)


kiplingw:
Thanks @Tachi107. I'm fine with it, notwithstanding the two CI failures on: `NOTICE: You are using Python 3.6 which is EOL. Starting with v0.62.0, Meson will require Python 3.7 or newer`. We can merge now and fix that later, if you like, since that's not really anything to do with your PR.

Tachi107:
Il giorno ven 19 ago 2022 alle 08:21:35 -07:00:00, Kip 
***@***.***> ha scritto:
> Thanks @Tachi107. I'm fine with it, notwithstanding the two CI 
> failures on: NOTICE: You are using Python 3.6 which is EOL. Starting 
> with v0.62.0, Meson will require Python 3.7 or newer. We can merge 
> now and fix that later, if you like, since that's not really anything 
> to do with your PR.
> 

The Python 3.6 notice isn't actually the cause of the failure, which is 
explained a few lines above:

   meson.build:5:0: ERROR: Compiler clang++ can not compile programs.

I've fixed this in #1092, thanks for merging both!

-- 
OpenPGP key: 66DE F152 8299 0C21 99EF  A801 A8A1 28A8 AB1C EE49

Tachi107:
@patrick-kings, could you check if this fixes your issue with CMake FetchContent?

patrick-kings:
Thank you so much, this works fine

Tachi107:
Awesome, thanks for the feedback :D

PR #1097, opened 2022-09-14
Add vcpkg installation instructions

Adela0814:
`pistache` is available as a port in vcpkg, a C++ library manager that simplifies installation for `pistache` and other project dependencies. Documenting the install process here will help users get started by providing a single set of commands to build pistache, ready to be included in their projects.

We also test whether our library ports build in various configurations (dynamic, static) on various platforms (OSX, Linux, Windows: x86, x64, UWP, ARM) to keep a wide coverage for users.

I'm a maintainer for vcpkg, and [here is what the port script looks like](https://github.com/microsoft/vcpkg/blob/master/ports/pistache/portfile.cmake). We try to keep the library maintained as close as possible to the original library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -226,6 +226,21 @@ Some other Meson options:
 | PISTACHE_BUILD_EXAMPLES       | False   | Build all of the example apps                  |
 | PISTACHE_BUILD_DOCS           | False   | Build Doxygen docs                             |
 
+
+## Installing pistache using vcpkg
+
+You can download and install pistache using the [vcpkg](https://github.com/Microsoft/vcpkg) dependency manager:
+
+```bash
+git clone https://github.com/Microsoft/vcpkg.git
+cd vcpkg
+./bootstrap-vcpkg.sh
+./vcpkg integrate install
+./vcpkg install pistache
+```
+
+The pistache port in vcpkg is kept up to date by Microsoft team members and community contributors. If the version is out of date, please [create an issue or pull request](https://github.com/Microsoft/vcpkg) on the vcpkg repository.
+
 ## Example
 
 ### Hello World (server)


Tachi107:
Hi, thanks for your patch! How is installing the library with vcpkg different from installing with plain `meson setup build && meson install -C build`?

Tachi107:
> pistache will only work on Linux

Yes, and the vcpkg port [already accounts for this](https://github.com/microsoft/vcpkg/blob/master/ports/pistache/portfile.cmake#L1-L3)

kiplingw:
> > pistache will only work on Linux
> 
> Yes, and the vcpkg port [already accounts for this](https://github.com/microsoft/vcpkg/blob/master/ports/pistache/portfile.cmake#L1-L3)

But what is the point of adding support for it if it can't be built on Windows?

Tachi107:
Il giorno mer 14 set 2022 alle 09:49:59 -07:00:00, Kip 
***@***.***> ha scritto:
> But what is the point of adding support for it if it can't be built 
> on Windows?
> 

vcpkg ports can be useful for people using the vcpkg package manager in 
e.g. CMake, even if they are targeting Linux. I still don't understand 
the usefulness of installing Pistache using the commands added in this 
patch; I simply don't know what's the difference between that and 
`meson setup build && meson install -C build`

eli-schwartz:
> I still don't understand the usefulness of installing Pistache using the commands added in this patch; I simply don't know what's the difference between that and `meson setup build && meson install -C build`

It's no different from providing a fifteen line overview of how to `sudo apt install pistache` (if one assumed that Debian packaged it), complete with instructions on how to:
- install Debian, so you can use its package manager to install pistache
- report an out of date package to Debian

It would probably be a lot more useful to add a repology badge:
```
[![Packaging status](https://repology.org/badge/vertical-allrepos/pistache.svg)](https://repology.org/project/pistache/versions)
```
[Packaging status](https://repology.org/project/pistache/versions)

And then people could click there, and click a second time for whatever their favorite package manager is -- oh hey, it looks like it's in conan too. ;) Also some recipe collection for a package manager based on a fork of Bazel, so that's apparently... a thing?

Issue #1103, opened 2022-10-31
pistache server URL is not reatcheable after remove and add route


chorfa007:
Dear All,

I would like to implement a server where i add and remove routes  dynamically without rebooting server.
Below the example 

```
`#pragma once
#include <pistache/endpoint.h>
#include <pistache/http.h>
#include <pistache/router.h>

#include "rapidjson/document.h"
#include <csignal>
#include <iostream>

using namespace Pistache;
using namespace Rest;
class Server
{
public:
    enum Modes
    {
        mode_1,
        mode_2
    };    
    Server();
    virtual ~Server() = default;
    void SwitchMode();
    void StartServer(Server::Modes mode);
    void StopServer();

private:
    Router router_;
    Router empty_router_;
    std::shared_ptr<Http::Endpoint> endpoint_;
private:
    void SetupRoutes();
    void hello_fun_1_mode_1(const Rest::Request& request, Http::ResponseWriter response);
    void hello_fun_2_mode_1(const Rest::Request& request, Http::ResponseWriter response);
    void hello_fun_1_mode_2(const Rest::Request& request, Http::ResponseWriter response);
    void hello_fun_2_mode_2(const Rest::Request& request, Http::ResponseWriter response);
    Modes system_mode_ = Modes::mode_2;
};
#include "Server.hpp"

using namespace Pistache;
using namespace Rest;
Server::Server()
{
    Port port(9900);
    Pistache::Address addr(Pistache::Ipv4::any(), Pistache::Port(port));
    endpoint_ = std::make_unique<Pistache::Http::Endpoint>(addr);
    auto opts = Pistache::Http::Endpoint::options()
            .threads(2)
            .flags(Pistache::Tcp::Options::ReuseAddr);
    endpoint_->init(opts);
}

void Server::StartServer(Server::Modes mode)
{
    system_mode_ = mode;
    SetupRoutes();
    endpoint_->setHandler(router_.handler());
    endpoint_->serveThreaded();
}

void Server::StopServer()
{
    endpoint_->shutdown();
}

void Server::SetupRoutes()
{
    using namespace Pistache::Rest;
    if (system_mode_ == Modes::mode_1)
    {
        Routes::Get(router_, "/hello_fun_1_mode_1", Routes::bind(&Server::hello_fun_1_mode_1, this));
        Routes::Get(router_, "/hello_fun_2_mode_1", Routes::bind(&Server::hello_fun_2_mode_1, this));
    }
    else
    {
[...]

PR #1105, opened 2022-11-02
pistache server throw exception

chorfa007:
Reference github project https://github.com/pistacheio /
[pistache](https://github.com/pistacheio/pistache) issue #1103 :   std::remove throwing an instance of 'std::runtime_error' ( Requested does not exist) #1103 

I just create a UT that include a crash when removing route, just to let people get my PR onto their development box and dig into the problem.


chorfa007:
including @dennisjenkins75 @Tachi107

kiplingw:
Good work @chorfa007. Setting up a PR to start with reproducing the issue is the right way to go about this while the fix is in the works.

chorfa007:
thanks @kiplingw, being able to edit the routes live would be a really nice feature i hope that the fix can be ready soon

tests/remove_routes_issue.cc
```diff
@@ -0,0 +1,145 @@
...
+        endpoint_->shutdown();
+    }
+    Port getPort()
+    {
+        return endpoint_->getPort();
+
+    }
+
+private:
+    Router router_;
```
chorfa007, line 72: use a shared pointer here is mandatory if we want to make the route table mutable, with this implementation the route is not mutable that explain why   PR fails

chorfa007:
@arghness  thanks for the example,Good work !   could you please create a PR with your proposal ?
BTW i fixed my PR i will keep it as an increase of UT coverage because route remove was not unit tested

PR #1111, opened 2023-01-04
Fetch GoogleTest from its main branch

shlomnissan:
The [GoogleTest](https://github.com/google/googletest) repository renamed its primary branch from `master` to `main`, but CMake's `FetchContent` defaults to `master`, so fetching the library if it isn't installed locally fails. This PR adds a tag to the CMakeLists file to resolve this issue.


Tachi107:
@shlomnissan thanks for the fix!

@dennisjenkins75 thanks for the ping, but Meson uses a "wrap file" pinned to a specific release (currently 1.11) so it is not needed; see [subprojects/gtest.wrap](https://github.com/pistacheio/pistache/blob/master/subprojects/gtest.wrap)

shlomnissan:
@Tachi107 No worries :) Thanks for all your hard work!

@dennisjenkins75 I'm not familiar with commitlint. Is it a matter of updating the title/description? I did that but I can't verify/re-run the check.

Tachi107:
Il giorno mer 4 gen 2023 alle 12:05:18 -08:00:00, Shlomi Nissan 
***@***.***> ha scritto:
> I'm not familiar with commitlint. Is it a matter of updating the 
> title/description? I did that but I can't verify/re-run the check

You simply have to reformat you commit message into something like 
"build(cmake): fetch GoogleTest from its main branch". To rename your 
commit locally, you can do

    $ git commit --amend
    $ git push --force

Issue #1104, opened 2022-10-31
[BUG] Listener::~Listener()'s acceptThread.join() deadlocks


kiplingw:
This may be related to #708 or #761, but I've noticed an issue recently with one of my servers that appears to hang at the `acceptThread.join()` call here:

```
    Listener::~Listener()
    {
        if (isBound())
            shutdown();
        if (acceptThread.joinable())
            acceptThread.join();

        if (listen_fd >= 0)
        {
            close(listen_fd);
            listen_fd = -1;
        }
    }
```

This appears to happen occasionally on destruction via `Pistache::Http::Endpoint::~Endpoint()`. The `acceptThread.join()` is just a wrapper for `std::thread::join()`, of course.

I have two separate core dumps of the same process. One _before_ I asked it to terminate gracefully via SIGTERM signal and the other _after_ issuing the signal. In both I see a Pistache listener thread that appears to be hanging in OpenSSL's `BIO_read()`. I'm pretty sure this is the problem because in the coredump post SIGTERM after all other threads, besides main, have exited, still remains:

```
#0  __GI___libc_read (nbytes=5, buf=0x7ffb94008da3, fd=4) at ../sysdeps/unix/sysv/linux/read.c:26
        sc_ret = -512
        sc_cancel_oldtype = 0
        sc_ret = <optimized out>
#1  __GI___libc_read (fd=4, buf=0x7ffb94008da3, nbytes=5) at ../sysdeps/unix/sysv/linux/read.c:24
#2  0x00007ffb9d5f0ddd in  () at /lib/x86_64-linux-gnu/libcrypto.so.3
#3  0x00007ffb9d5e12eb in  () at /lib/x86_64-linux-gnu/libcrypto.so.3
#4  0x00007ffb9d5e42e5 in  () at /lib/x86_64-linux-gnu/libcrypto.so.3
#5  0x00007ffb9d5e4467 in BIO_read () at /lib/x86_64-linux-gnu/libcrypto.so.3
#6  0x00007ffb9cba934c in  () at /lib/x86_64-linux-gnu/libssl.so.3
#7  0x00007ffb9cbaaf19 in  () at /lib/x86_64-linux-gnu/libssl.so.3
#8  0x00007ffb9cbbcb36 in  () at /lib/x86_64-linux-gnu/libssl.so.3
#9  0x00007ffb9cbbe608 in  () at /lib/x86_64-linux-gnu/libssl.so.3
[...]

PR #1115, opened 2023-02-07
Fix: DoS vulnerability in SSL_accept() hanging indefinitely with clie…

kiplingw:
Fix: DoS vulnerability in SSL_accept() hanging indefinitely with clients that say nothing during TLS handshake as described in #1104...
version.txt: Bumped version from 0.1.0 to 1.0.0...


kiplingw:
Thankfully the patch is not long. Basically all I did was at a user configurable option for the SSL handshake timeout. This is set on the socket descriptor before `SSL_accept()` and disabled immediately after the handshake completes. Let me know what you think of the semantic changes. I can change formatting before we merge.

Tachi107:
As Dennis said, please try to keep the diff at a minimum; this will make it easier to backport the fix into the Debian package, among other things.

As for bumping the version, we should do that in another PR, for the same reasons mentioned above.

Other than that, thank you for the fix, well done!

Tachi107:
> Do we have a \".clang-format\" file? if not, we need one.

[Yes, we do](https://github.com/pistacheio/pistache/blob/master/.clang-format)

kiplingw:
Ok [take a look](https://github.com/pistacheio/pistache/pull/1115/files?diff=split&w=1) now guys. I've ran clang-format(1) on the affected source and pushed the new changes. The `version.txt` I didn't split into a separate PR because we have to bump the version anyways because this change would likely break the ABI because we've changed some public interfaces, and not simply added additional ones.

kiplingw:
Thanks @dennisjenkins75. @Tachi107, unless you have anything to add on my patch, I'll hold off on merging until your unit test is added and confirms a fail without this fix.

include/pistache/stream.h
```diff
@@ -83,10 +83,10 @@ namespace Pistache
     public:
         using Base = StreamBuf<CharT>;
 
-        explicit ArrayStreamBuf(size_t maxSize)
+        explicit ArrayStreamBuf(size_t _maxSize)
             : StreamBuf<CharT>()
             , bytes()
-            , maxSize(maxSize)
+            , maxSize(_maxSize)
```
Tachi107, line 89: Were these leading underscores added by mistake?

kiplingw: Intentionally added to suppress a compiler warning about the argument parameter shadowing a class attribute by the same name.

Tachi107:
On Fri Feb 10, 2023 at 6:55 PM CET, Kip wrote:
> Intentionally added to suppress a compiler warning about the argument parameter shadowing a class attribute by the same name.

If so, maybe it'd be better to keep the argument name as-is and rename
the private class attribute instead. All the other classes in stream.h
name their attributes with a trailing underscore, ArrayStreamBuf is the
only exception. Also, while in this case _maxSize is safe to use, it is
easy to violate the C++ standard when naming variables this way, which
says that identifiers starting with two underscores or with an
underscore and a capital letter are reserved and can't be used by
regular code.

kiplingw:
Yup. That's a good point. Fixed.

Tachi107:
While writing a test for this, I started to think that adding an `sslHandshakeTimeout` parameter to `Listener::init` isn't ideal, since:

1. Is a breaking change
2. Affects everybody, even ones not using Pistache's TLS support

It may be better to instead add it as another parameter to `Listener::setupSSL`, because:

1. It's a setting related to TLS, so it is conceptually in a better place
2. Isn't a breaking change (at least in terms of API compatibility, the ABI will still probably be changed)

Patching this issue in an API compatible way will also allow me to write a test that can compile on both old a new versions of Pistache, so that it is easy to test the behaviour both before and after this patch

Still related to this issue but unrelated to the above, isn't 60 seconds a bit too high? [Go's HTTP library](https://github.com/golang/go/blob/a7b75972f267aac5607b16b0e9826d56386a92af/src/net/http/transport.go#L52), for example, sets it to 10 seconds by default, which I think is a way safer default. No TLS-aware, well intentioned client will wait more than ten seconds before initiating the TLS handshake, and in the case you need to increase the default you can always do it. Having a timeout of 10 seconds by default will also make writing the test a bit more straightforward.

Last (and least), Kip could you please rebase and cleanup a bit the commit history? I'd like to have the fix in a single commit that minimizes the diff, so you could for example separate this PR in:

1. A commit which fixes the issue
2. A commit that bumps the version
3. A commit that contains the cosmetic changes like the `_maxSize` rename, `clang-format`'s changes, etc.

Tachi107:
@kiplingw I've reordered the git history with a force-push, hope this doesn't look like being rude :)

Tachi107:
Here's how I'd apply my suggestions above

```diff
diff --git a/include/pistache/config.h b/include/pistache/config.h
index f26579c..4ef1646 100644
--- a/include/pistache/config.h
+++ b/include/pistache/config.h
@@ -28,7 +28,7 @@ namespace Pistache::Const
     static constexpr auto DefaultHeaderTimeout       = std::chrono::seconds(60);
     static constexpr auto DefaultBodyTimeout         = std::chrono::seconds(60);
     static constexpr auto DefaultKeepaliveTimeout    = std::chrono::seconds(300);
-    static constexpr auto DefaultSSLHandshakeTimeout = std::chrono::seconds(60);
+    static constexpr auto DefaultSSLHandshakeTimeout = std::chrono::seconds(10);
     static constexpr size_t ChunkSize                = 1024;
 
     static constexpr uint16_t HTTP_STANDARD_PORT = 80;
diff --git a/include/pistache/listener.h b/include/pistache/listener.h
index 6c1888c..13c2bb3 100644
--- a/include/pistache/listener.h
+++ b/include/pistache/listener.h
@@ -60,11 +60,10 @@ namespace Pistache::Tcp
 
         explicit Listener(const Address& address);
         void init(size_t workers,
-                  Flags<Options> options                        = Flags<Options>(Options::None),
-                  const std::string& workersName                = "",
-                  std::chrono::milliseconds sslHandshakeTimeout = Const::DefaultSSLHandshakeTimeout,
-                  int backlog                                   = Const::MaxBacklog,
-                  PISTACHE_STRING_LOGGER_T logger               = PISTACHE_NULL_STRING_LOGGER);
+                  Flags<Options> options          = Flags<Options>(Options::None),
+                  const std::string& workersName  = "",
+                  int backlog                     = Const::MaxBacklog,
+                  PISTACHE_STRING_LOGGER_T logger = PISTACHE_NULL_STRING_LOGGER);
 
         void setTransportFactory(TransportFactory factory);
         void setHandler(const std::shared_ptr<Handler>& handler);
@@ -88,7 +87,8 @@ namespace Pistache::Tcp
         void pinWorker(size_t worker, const CpuSet& set);
 
         void setupSSL(const std::string& cert_path, const std::string& key_path,
-                      bool use_compression, int (*cb_password)(char*, int, int, void*));
+                      bool use_compression, int (*cb_password)(char*, int, int, void*),
+                      std::chrono::milliseconds sslHandshakeTimeout = Const::DefaultSSLHandshakeTimeout);
         void setupSSLAuth(const std::string& ca_file, const std::string& ca_path,
                           int (*cb)(int, void*));
         std::vector<std::shared_ptr<Tcp::Peer>> getAllPeer();
diff --git a/src/server/endpoint.cc b/src/server/endpoint.cc
index 36e5fbe..dd388ad 100644
--- a/src/server/endpoint.cc
+++ b/src/server/endpoint.cc
@@ -249,7 +249,7 @@ namespace Pistache::Http
 
     void Endpoint::init(const Endpoint::Options& options)
     {
-        listener.init(options.threads_, options.flags_, options.threadsName_, options.sslHandshakeTimeout_);
+        listener.init(options.threads_, options.flags_, options.threadsName_);
         listener.setTransportFactory([this, options] {
             if (!handler_)
                 throw std::runtime_error("Must call setHandler()");
@@ -294,7 +294,7 @@ namespace Pistache::Http
 #ifndef PISTACHE_USE_SSL
         throw std::runtime_error("Pistache is not compiled with SSL support.");
 #else
-        listener.setupSSL(cert, key, use_compression, pass_cb);
+        listener.setupSSL(cert, key, use_compression, pass_cb, options_.sslHandshakeTimeout_);
 #endif /* PISTACHE_USE_SSL */
     }
 
diff --git a/src/server/listener.cc b/src/server/listener.cc
index 51cbb24..38d2661 100644
--- a/src/server/listener.cc
+++ b/src/server/listener.cc
@@ -206,22 +206,20 @@ namespace Pistache::Tcp
     }
 
     void Listener::init(size_t workers, Flags<Options> options,
-                        const std::string& workersName,
-                        std::chrono::milliseconds sslHandshakeTimeout,
-                        int backlog, PISTACHE_STRING_LOGGER_T logger)
+                        const std::string& workersName, int backlog,
+                        PISTACHE_STRING_LOGGER_T logger)
     {
         if (workers > hardware_concurrency())
         {
             // Log::warning() << "More workers than available cores"
         }
 
-        options_             = options;
-        sslHandshakeTimeout_ = sslHandshakeTimeout;
-        backlog_             = backlog;
-        useSSL_              = false;
-        workers_             = workers;
-        workersName_         = workersName;
-        logger_              = logger;
+        options_     = options;
+        backlog_     = backlog;
+        useSSL_      = false;
+        workers_     = workers;
+        workersName_ = workersName;
+        logger_      = logger;
     }
 
     void Listener::setTransportFactory(TransportFactory factory)
@@ -622,7 +620,8 @@ namespace Pistache::Tcp
     void Listener::setupSSL(const std::string& cert_path,
                             const std::string& key_path,
                             bool use_compression,
-                            int (*cb_password)(char*, int, int, void*))
+                            int (*cb_password)(char*, int, int, void*),
+                            std::chrono::milliseconds sslHandshakeTimeout)
     {
         SSL_load_error_strings();
         OpenSSL_add_ssl_algorithms();
@@ -636,7 +635,8 @@ namespace Pistache::Tcp
             PISTACHE_LOG_STRING_FATAL(logger_, e.what());
             throw;
         }
-        useSSL_ = true;
+        sslHandshakeTimeout_ = sslHandshakeTimeout;
+        useSSL_              = true;
     }
 
 #endif /* PISTACHE_USE_SSL */
```

kiplingw:
@Tachi107, your diff looks perfect and appears to address all the issues. I was going to do that later today, but you've saved me some time.

Are you able to commit directly into my branch?

kiplingw:
@Tachi107, I've integrated your prudent suggestions. I'm still trying to figure out how to split up the changes logically into separate commits.

kiplingw:
Ok I think I've figured out how to do the rebasing:

- The `maxSize_` cosmetic / compiler warning fix was [already](https://github.com/pistacheio/pistache/pull/1115/commits/e10db45f9828ef3fb7c139220bbbfe9afe280f18) in it's own commit;
- The `version.txt` change is in its [own commit](https://github.com/pistacheio/pistache/pull/1115/commits/ebc96a4f2a9d7b07fa8a02727412f41db672a504) and no longer reflecting an ABI change;
- All the substantive changes for the DoS fix itself are in their [own commit](https://github.com/pistacheio/pistache/pull/1115/commits/2ec7a485faf61794fdbd65332472b14dd5182484);
- The `clang-format(1)` aesthetic changes are in their [own commit](https://github.com/pistacheio/pistache/pull/1115/commits/01a18ad01962c90fcd654554eb4814e37dac2c6b).

If it looks good to you, and your unit test PR works, we can merge if everyone's happy.

Tachi107:
@kiplingw can I squash the last commits? After that, I'll merge

kiplingw:
I just gave it a try. Not sure if I did it correctly. But feel free to correct. I've never squashed before.

Tachi107:
Uhm, it seems that you've duplicated the last few commits :)

No worries, I can do it myself. <https://git-rebase.io/#squash> might be helpful though. It's written by the the awesome people behind SourceHut

Tachi107:
Done. I've also shortened the commit message titles to make the commitlint check happier (it still doesn't like the fact that "DoS" starts with a capital "D", since it expects things that are after "fix: " to be lower case, but whatever).

Tachi107:
In order to keep the ABI backwards compatible, the new `sslHandshakeTimeout_` class member should be placed at the end. I wasn't sure if this was the case though, and while searching for confirmation I've found [this nice slide show](https://events.static.linuxfound.org/sites/events/files/slides/Binary_Compatibility_for_library_devs.pdf) that contains a lot of useful information in little text (just in case you're interested).

Tachi107:
@kiplingw if you're ok with my last change (moving `sslHandshakeTimeout_`) I'll merge

kiplingw:
Yes, it all looks good to me. Go ahead and merge. The issue of the `meson.build` hardcoding the project version is still an issue, but one that can be addressed in a separate PR.

PR #1120, opened 2023-02-18
ci: automatically update debian/changelog

Tachi107:
This CI job will run when `version.txt` is updated in the `master` branch, and will push a new commit to the `debian` branch updating the changelog.

`DEBFULLNAME` and `DEBEMAIL` are set by "parsing" the control file. If you know a true debian/control parser that could be used instead, preferably part of dpkg or devscripts, please let me know.

This script currently only updates the single existing changelog entry. It could be possible to make it add a new entry each time instead, but it'd be a bit more involved and we would have to first check if it'd be compatible with our current PPA setup, which I personally don't know very well.

Closes #1119

### DIFF
--- a/.github/workflows/debian-changelog.yaml
+++ b/.github/workflows/debian-changelog.yaml
@@ -1,0 +1,45 @@
+# SPDX-FileCopyrightText: 2023 Andrea Pappacoda
+# SPDX-License-Identifier: Apache-2.0
+
+name: debian-changelog
+
+on:
+  push:
+    branches:
+    - master
+  paths:
+    - version.txt
+
+defaults:
+  run:
+    shell: sh
+
+jobs:
+  debian-changelog:
+    runs-on: ubuntu-latest
+
+    steps:
+
+    - name: Install dependencies
+      run: |
+        apt update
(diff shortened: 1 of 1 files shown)


Tachi107:
Forgot to mention that this would also change how `debian/changelog` records the version. We currently only update it when bumping the major, minor or patch versions, while this updates and writes to it also when bumping the date. Is this OK for the PPA? If I recall correctly this shouldn't cause issues, and might even be better than before.

kiplingw:
Hey @Tachi107. Nice job (again). You might want to try experimenting with the [dctrl-tools](https://packages.ubuntu.com/search?keywords=dctrl-tools&searchon=names&suite=all&section=all) package which I think can do exactly that - parse `debian/control` files.

I don't know if it will break in the PPA, but worst case is the Launchpad builder will let us know.

Tachi107:
I've looked at dctrl-tools, but what they do is extract the `Maintainer` field as-is, you cannot split it in two variables like `DEBFULLNAME` and `DEBEMAIL`. Not a big deal though, `cut` works fine.

And no, I don't need to extract the version from d/changelog, because I have to overwrite it with what's in `version.txt`

PR #1123, opened 2023-02-19
ci(debian-changelog): last fixes

Tachi107:
These should be the last fixes needed to make this work. It does work now in my fork, see <https://github.com/Tachi107/pistache/actions/runs/4217592497/jobs/7321484202>


Tachi107:
After this, the next time version.txt is bumped `debian/changelog` will be updated by CI

Tachi107:
Ops, I forgot about this PR for some reason. Merging.

PR #1130, opened 2023-04-24
Support multiple values for query params

sanjayatn:


kiplingw:
And rebase too.

sanjayatn:
@Tachi107 @kiplingw I didn't realized you can see this branch but I love to submit it as a proper PR! Coming up today!

sanjayatn:
@Tachi107 @kiplingw I see two things failing in the checks. Which one should focus on? If you can give me 2 sentence summary of what abidiff / abi step's goal, then I have better idea to fix it.

kiplingw:
Hey @sanjayatn. For commitlint you can find what it's complaining about [here](https://github.com/pistacheio/pistache/actions/runs/4811284454/jobs/8565705074?pr=1130). For abidiff my guess is you may have missed [this](https://github.com/pistacheio/pistache#versioning).

sanjayatn:
@kiplingw thanks!

Tachi107:
Hi @sanjayatn, thanks for the added description and sorry for the delayed response.

I'm personally not a fan of multiple query parameters because, as you say, the behaviour is not standardized, and differences in how different software components handle them can be a source of security vulnerabilities, so much that this class of vulnerabilities got a name: "HTTP parameter pollution".

That being said, aligning to what others do is probably fine, as long as the others do sensible stuff.

Could you please add a couple of test cases showing this new behaviour? Thanks!

PR #1131, opened 2023-05-21
build(cmake): fix configure_file version header

shlomnissan:
A relative path for `configure_file` `<output>` is treated with respect to the value of `CMAKE_CURRENT_BINARY_DIR`. This PR adds an absolute path to the output parameter so `version.h` is generated and stored in the include directory.


kiplingw:
Thanks @shlomnissan. Your PR looks mostly fine to me, but I have two comments.

(1) I am curious why `include/pistache/version.h` has been added to your PR since it's suppose to be machine generated at configure time from the `include/pistache/version.h.in` template. Perhaps this was by accident?

(2) Does the same issue you are trying to address also affect the Meson build environment?

@Tachi107, any feedback?

Tachi107:
Hi @shlomnissan, thanks for your patch. I believe that the current behaviour is the intended one, i.e. the generated file should reside in the a build directory that gets added to the include list by CMake or Meson. Why do you think that the version file should be generated in the source directory?

kiplingw:
I actually didn't realize that he was trying to output it into the source directory. But yes, @Tachi107 is correct. That's how build environments normally are suppose to work. Since the source directory in some situations might be read only, that's the only way to allow remote builds.

shlomnissan:
Interesting... I didn't know that the compiler driver includes the build folder in its header search paths.

Suppose I want to use the version variables defined in `version.h` to update the user agent header value in the client so that it's not hard-coded (`client.cc:33`). How would I include it?

kiplingw:
The `version.h` ships with the binary package with the versioning baked into it. You'd just use the variables that are in it already in your application. But note that those will only say the version it was compiled against. If the library is still binary compatible then it's possible the application could be linking at runtime with a version newer than what it saw at compile time.

shlomnissan:
Thank you for the feedback, and I apologize for the delayed response. I’m going to close the PR. However, I am still uncertain about how to use `version.h` in my application with my existing workflow. Specifically, I am including the source code in my project and adding it with CMake's `add_subdirectory`. The current binary directory is not part of the default header search path, so my project can’t find it.

kiplingw:
Hey @shlomnissan,

I strongly recommend not including the project source within your project. If you want to make source level changes, then by all means do that. But my guess is you probably just want to use the library.

If the latter, at configure / build time you want to use `pkg-config(1)`, as discussed [here](https://github.com/pistacheio/pistache/#use-via-pkg-config). The pkg-config manifest ships with the binary packages that you can find [here](https://github.com/pistacheio/pistache/#precompiled-packages).

As for using `version.h`, you're going to do something like this:

```
    #include <pistache/version.h>

    cout << Pistache::Version::Major << "." << Pistache::Version::Minor << "." << Pistache::Version::Patch << ".-git" << Pistache::Version::Git);
```

Hopefully that's helpful.

PR #1134, opened 2023-06-08
tests/net_test: don't try to resolve external names

daniloegea:
The test case "address_creation" relies on resolving the name www.example.com to succeed. It means that testing the project on environments without access to a DNS server will fail. Use the name "localhost" instead so the local resolver will not need to query a DNS server.

### DIFF
--- a/tests/net_test.cc
+++ b/tests/net_test.cc
@@ -106,8 +106,8 @@ TEST(net_test, address_creation)
     ASSERT_EQ(address14.family(), AF_INET);
     ASSERT_EQ(address14.port(), 80);
 
-    Address address15("www.example.com");
-    ASSERT_EQ(address15.host(), "93.184.216.34");
+    Address address15("localhost");
+    ASSERT_EQ(address15.host(), "127.0.0.1");
     ASSERT_EQ(address15.family(), AF_INET);
     ASSERT_EQ(address15.port(), 80);
 


Tachi107:
Hi @daniloegea,  thanks for your patch. While using `localhost` looks like it'd make sense, Pistache is [currently hardcoded](https://github.com/pistacheio/pistache/blob/a68ad0902d2cfc23f69fc16e26747ac77bc2f123/src/common/net.cc#L441-L444) to return `127.0.0.1` if `localhost` is specified, meaning that dropping `www.example.com` would reduce test coverage. Does it make sense to you? Or would it be better to let `getaddrinfo()` handle `localhost` itself?

The `net_test` is special and opt-in because it requires internet access; how did you run into issues with it?

daniloegea:
Hi @Tachi107, thanks for replying.

I came across this issue while investigating why the Pistache Debian package was failing to build on Ubuntu. The reason is that the Ubuntu builders don't have internet access. And it's not rare to see CI pipelines with restricted network access.

In my opinion it would be a better approach to let the system's libc resolve names, even for localhost, otherwise the library won't follow whatever the user added to /etc/hosts. Or at least the call to the system's resolver should be mocked.

Besides, is it guaranteed that www.example.com will not start resolving to a different IP in the future?

kiplingw:
@daniloegea, in case it helps, unit tests that require network connectivity are [disabled](https://github.com/pistacheio/pistache/blob/abbce651ee35937188841772b37213d71c82df72/rules#L40) when the Debian package is being built.

Tachi107:
@daniloegea I do agree with you and I think that Pistache shouldn't be special casing `localhost`, but it should simply always use the system's resolver with `getaddrinfo()`. I'm going to submit a patch to do just that pretty soon.

By the way, what do you mean by building the Debian package on Ubuntu? Are you trying to package Pistache in Ubuntu's main repos or are you just trying to build it in a "third party" repository?

daniloegea:
@kiplingw thanks for the suggestion. That's interesting, Debian doesn't really use this rule: https://tracker.debian.org/media/packages/p/pistache/rules-0.0.5ds-3. And I just realized that @Tachi107 is the package's maintainer :)

@Tachi107 No, Ubuntu already has the package (https://launchpad.net/ubuntu/+source/pistache). Ubuntu basically copies the package from Debian as it is and rebuilds it. We probably hit this issue because debian/rules is not skipping the test suit that requires network as @kiplingw mentioned.

Tachi107:
> That's interesting, Debian doesn't really use this rule:

Weird, I probably forgot to turn it on. I'll fix it as soon as possible

Tachi107:
Fixed in [version 0.0.5+ds-4](https://salsa.debian.org/tachi/pistache/-/compare/debian%2F0.0.5+ds-3...debian%2F0.0.5+ds-4?from_project_id=72793&straight=false). It will soon hit the Debian archive

Tachi107:
Note: with this merged, tests doesn't cover the getaddrinfo() portion of the code. I intend to change that, but I didn't do it yet.

PR #1136, opened 2023-06-23
Add support for custom headers with dash in their names

sjoubert:
- Request both the class name and header name in the macro as lots of headers have a '-' but their name can't also be used for the class
- Prefix the new macro with PISTACHE_ to avoid potential name conflicts


sjoubert:
I'm not sure what is your backward compatibility policy. In doubt I've kept the `CUSTOM_HEADER` macro, but it should probably be deprecated in favor of the new one. I don't really know how to mark a macro as deprecated though.

include/pistache/http_header.h
```diff
@@ -628,19 +628,19 @@ namespace Pistache::Http::Header
         std::string ua_;
     };
 
-#define CUSTOM_HEADER(header_name)                                                  \
-    class header_name : public Pistache::Http::Header::Header                       \
+#define PISTACHE_CUSTOM_HEADER(header_class, header_name)                           \
+    class header_class : public Pistache::Http::Header::Header                      \
```
Tachi107, line 632: Is renaming the class required? Couldn't you change the macro parameter name instead? It feels like it would cause less backwards compatibility issues to me

include/pistache/http_header.h
```diff
@@ -654,6 +654,8 @@ namespace Pistache::Http::Header
         std::string value_;                                                         \
     };
 
+#define CUSTOM_HEADER(header_name) PISTACHE_CUSTOM_HEADER(header_name, #header_name)
```
Tachi107, line 657: Pistache already defines several unprefixed macros, and fixing it is out of the scope of this patch, in my opinion.

Tachi107 (changes requested):
Looking at the relevant RFCs I couldn't find a reason not to allow dashes in headers, so this patch is probably ok. [RFC 9110](https://www.rfc-editor.org/rfc/rfc9110#fields.registry) even explicitly mentions that dashed headers are allowed to be registered as "official" headers.

Still, I have a few questions below.

include/pistache/http_header.h
sjoubert: Do you mean renaming the "macro"? In order to support different class name and header name we'll need 2 different parameter since header names with '-' are not valid class names. So unless compatibility is dropped I need a second macro for the new use case.

include/pistache/http_header.h
sjoubert: See my answer for the other point. We need a second macro name if we want to be backward compatible (unless there is a way I'm not seeing). So prefixing the macro seemed a good practice for the new name. Otherwise it would have to be `CUSTOM_HEADER_2` or something like that.

Tachi107: Sorry, I misread the diff, you're right :)

tests/headers_test.cc
```diff
@@ -679,14 +679,14 @@ TEST(headers_test, server_test)
...
-CUSTOM_HEADER(TestHeader)
+PISTACHE_CUSTOM_HEADER(TestHeader, "Test-Header")
 
 TEST(headers_test, macro_for_custom_headers)
 {
     TestHeader testHeader;
     std::ostringstream os;
 
-    ASSERT_TRUE(strcmp(TestHeader::Name, "TestHeader") == 0);
+    ASSERT_TRUE(strcmp(TestHeader::Name, "Test-Header") == 0);
```
Tachi107, line 689: Please use `EXPECT_STREQ(TestHeader::Name, "Test-Header")` instead

Tachi107 (commented):
Apart from a nitpick, this looks good to me. If you can rebase, I'll merge :)

Tachi107 (approved):

Tachi107:
Thanks for the patch, @sjoubert!

PR #1139, opened 2023-06-27
fix: drop memset() calls

Tachi107:
Using `memset()` to zero-initialize variables is kinda ugly, and not always correct. Use instead C++ and C23's empty brace list, which correctly zero-initializes any kind of object.

See the following docs for reference:

- <https://thephd.dev/ever-closer-c23-improvements#consistent-warningless-and-intuitive-initialization-with-->
- <https://www.open-std.org/jtc1/sc22/wg14/www/docs/n2900.htm>


kiplingw:
I think for consistency sake, it's a good idea to use parentheses with `sizeof`. As @dennisjenkins75 pointed out, sometimes it's actually required.

Tachi107:
To be clear, I did not add any new `sizeof expression` uses - I actually removed one. Forget what I said in my previous comment, I misread the example.

Anyway, unless there are other feedbacks, this can be merged

PR #1145, opened 2023-07-07
Make rapidjson dependency optional

radistmorse:
rapidjson is not really used anywhere in the server itself, only in one include file which the user might not even want to use. And in tests, which again, most users are probably not that interested in. If you want pistache just as a server lib - you don't need rapidjson.


kiplingw:
Just to be clear, are we talking about removing it as a build time or as a run time dependency? Because if it's the former, he still needs to update `debian/control`. If we're talking about just at runtime, at least on Debian based systems, `dh_shlibdeps` is not going to list it as a runtime dependency if it doesn't detect any symbols dependent on it.

radistmorse:
I'm talking about using pistache as a subproject. Obviously, if you're using the debian package you shouldn't care which compile flags pistache does or doesn't have. It only matters if you compile it yourself.
As for changing debian control, that's not for me to decide whether you want to include rapidjson support in your debian package.

kiplingw:
@radistmorse, the `Build-Depends-Arch` stanza doesn't in itself necessitate that RapidJSON will be a runtime dependency. It just means that the `rapidjson-dev` package is necessary in order to compile the package. But even if it's necessary for compile time, it still doesn't mean that it will be a runtime requirement though. For instance, some build dependencies could be header only libraries. Others might have a shared library, but no symbols were actually linked and thus `dh_shlibdeps` doesn't make it a hard requirement at runtime.

I personally don't recommend using Pistache as a subproject, but use the actual `libpistache-dev` Debian package. Including the project source in your tree is more maintenance for you. If you need to make source level changes, then by all means feel free to include it as a subproject.

So I still need to know whether `debian/control` needs to be modified.

radistmorse:
@kiplingw no, rapidjson won't be a runtime dependency, since no pistache code ever calls any functions from rapidjson. Looking at libpistache-dev (https://packages.debian.org/sid/libpistache-dev) I can see that it does depend on rapidjson-dev, which is logical, since the include file in pistache in turn includes rapidjson.
The switch that I propose to introduce has nothing to do with that though. It doesn't change the default behavior of the build: rapidjson is still used. So you don't need to update debian control, unless you want to switch off the rapidjson support for libpistache-dev, which you probably dont.
And to answer your first question: it removes the dependency during the build time, since rapidjson was never a runtime dependency to begin with.

Tachi107:
Hi all, sorry for not replying earlier.

While I have nothing against this patch, I'd prefer to use Meson's `feature` option type instead of `boolean` for things like these, so that if rapidjson isn't installed the build doesn't fail, but falls back to not using the Swagger thing. It is of course possible to use `feature`s like booleans, and force the feature to be enabled or disabled with, well, `-DPISTACHE_USE_RAPIDJSON=enabled` and `=disabled`.

You can find more info about this at <https://mesonbuild.com/Build-options.html#features>

@radistmorse, is it OK for you if I change the new option to be a feature instead?

radistmorse:
@Tachi107 I'm ok with anything, as long as I can disable rapidjson in my project :)
But I wonder if this really makes any sense, since rapidjson has a fallback option, so ```auto``` will always behave exactly like ```enabled```. Introducing ```feature``` for e.g. openSSL would be more reasonable, but I really can't think of a usecase where user would want to make openSSL ```auto```. You either need HTTPS support or you dont and in both cases using ```auto``` may bring you some problems.

Tachi107:
It's more of a consistency thing than anything. Using feature types 
makes sense to specify, well, optional features. It easily makes it 
possible for e.g. distro packagers to turn on all optional features, 
and not to cause build failures when optional dependencies without 
fallbacks are missing.

So yeah, I'd like to change all the relevant option types to features.

Tachi107:
Il giorno lun 10 lug 2023 alle 04:40:47 -07:00:00, radistmorse 
***@***.***> ha scritto:
> You either need HTTPS support or you dont and in both cases using 
> auto may bring you some problems

You may simply not care about it, and things will work out of the box 
without issues. If you don't need it but link to OpenSSL it's not an 
issue.

radistmorse:
>If you don't need it but link to OpenSSL it's not an issue.

It might be, if you compile it for an environment without openSSL. This is not like with rapidjson, libpistache will require libssl and libcrypto even if you do not create any ssl endpoints. The server will simply not start without these libraries present.

JnCrMx:
Could we also get a similar option for CMake? :D

PR #1146, opened 2023-07-07
build: set soname to major.minor

Tachi107:
Until the 1.0 release, setting the soname to `version_major.version_minor` gives us more freedom to make changes without being too restricted by ABI compatibility.

We never actually really commited to ABI stability before the 1.0 relase - the README states that at the bottom - but I made this decision clearer by updating the "versioning" section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ SPDX-License-Identifier: Apache-2.0
 [![codecov](https://codecov.io/gh/pistacheio/pistache/branch/master/graph/badge.svg)](https://codecov.io/gh/pistacheio/pistache)
 [![REUSE status](https://api.reuse.software/badge/github.com/pistacheio/pistache)](https://api.reuse.software/info/github.com/pistacheio/pistache)
 
-Pistache is a modern and elegant HTTP and REST framework for C++. It is entirely written in pure-C++17[*](#linux-only) and provides a clear and pleasant API.
+Pistache is a modern and elegant HTTP and REST framework for C++. It is entirely written in pure-C++17[\*](#linux-only) and provides a clear and pleasant API.
 
 ## Documentation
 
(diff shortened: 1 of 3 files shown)


kiplingw:
Hey @Tachi107. So sorry for leaving this to collect dust. I totally forgot about it.

The issue here is that SONAME versioning is not supposed to mirror the library's versioning. That's the general best practice. SONAME versioning is just supposed to reflect the interface.

@jcalcote wrote an excellent book, _Autotools: A Practioner's Guide to GNU Autoconf, Automake, and Libtool_. There's a section in chapter 7, _Library Versioning Is Interface Versioning_ which deals with this exact topic.

Tachi107:
> The issue here is that SONAME versioning is not supposed to mirror the 
> library's versioning. That's the general best practice. SONAME 
> versioning is just supposed to reflect the interface.
>
> @jcalcote wrote an excellent book, _Autotools: A Practioner's Guide to 
> GNU Autoconf, Automake, and Libtool_. There's a section in chapter 7, 
> _Library Versioning Is Interface Versioning_ which deals with this 
> exact topic.

I'm aware of that (and read that part of the book!), but I really don't 
want to be this careful this versioning. It'd be great to do so, but 
I don't have the time. Stating that our interface changes whenever our 
library minor version changes is a fine requirement and better than the 
current situation. Not ideal, but better.

kiplingw:
Can we merge now as a temporary measure, but pledge to implement Calcote's method later before we do a stable release?

Tachi107:
> Can we merge now as a temporary measure, but pledge to implement 
> Calcote's method later before we do a stable release?

That's fine for me. We cannot use libtool, but we can trivially store 
the SONAME in a small text file and have Meson read it from there.

Issue #1148, opened 2023-07-25
[Feature Request] Integrating compression in ResponseWriter (gzip and deflate)


kiplingw:
Hello everyone,

I'd like to put a feature request in as a starting point for discussion before I spend time implementing it.

Currently many clients, including Python's request library, [by default](https://requests.readthedocs.io/en/latest/community/faq/#encoded-data) send an [Accept-Encoding](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Encoding) header requesting at least `gzip` and `deflate` (the two most common codecs). The idea of clients expecting the server to be capable of sending a compressed response is so common it would not make sense for us to make every Pistache application go and re-implement the necessary changes when this could be implemented directly in the library.

What I was thinking was the following:
* Add necessary build dependencies to Pistache build environment that can be optionally toggled;
* Pre-compiled packages would by default have compression support;
* Application is responsible for parsing the client's [Accept-Encoding](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Encoding) request headers using existing API;
* Application toggles appropriate setting in `ResponseWriter` to ensure whatever user passes its `send()` method is automatically compressed;
* Application is responsible for setting response [Content-Encoding](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding) header.

Thoughts?

dennisjenkins75:
This is reasonable, so long as it can be disabled at runtime as well.

kiplingw:
To disable at runtime user would do nothing. If they hadn't configured `ResponseWriter` to do anything different, then its default behaviour would be to simply not compress what it's provided.

Tachi107:
This sounds like a more then reasonable request, and I support its implementation. Still, I have heard multiple times that TLS + compression can create significant security issues, even though I am not sure how or why. We should definitely investigate - or, in case you already know why this can happen, please write it here so that I can better understand the issue :)

@Ne02ptzero, since you've implemented TLS support for pistache in the past, do you know what issues compression with encryption can create?

kiplingw:
I did a bit of [research](https://www.feistyduck.com/library/bulletproof-tls-guide/online/configuration/http-and-application-security/issues-with-http-compression.html) this morning and it looks like the security issue arose when using integrated compression with TLS. But I don't think this is a problem for us since we're proposing to handle it at the Pistache library layer (although I'll do more research).

[...]

PR #1151, opened 2023-07-27
Partial fix for #1148 by adding deflate content encoding support...

kiplingw:
Partial fix for #1148 by adding deflate content encoding support...


src/common/http.cc
```diff
@@ -852,6 +856,25 @@ namespace Pistache::Http
...
 
+#ifdef PISTACHE_USE_CONTENT_ENCODING_GZIP
+
+        // User requested GZip compression...
+        if(contentEncoding_.has_value() &&
+           contentEncoding_.value().encoding() == Http::Header::Encoding::Gzip)
+        {
+            // Initialize constructor...
+            std::stringbuf compressorBuffer;
+            zstr::ostream compressor(&compressorBuffer);
```
Tachi107, line 867: I'm personally not a fan of using an (obscure?) C++ wrapper for Zlib, especially considering that our usage is pretty basic. The `libhpptools` Debian package [has never been updated](https://tracker.debian.org/pkg/libhpptools/news/) to a new upstream version, and contains the old `zstr` version available in the <https://github.com/mateidavid/hpptools> repository, and not the more recent <https://github.com/mateidavid/zstr>.

While I could update the Debian package myself, I believe we should simply use the regular zlib functions to compress the stream, especially considering how simple it is and how little we use it.

```cpp
size_t compressed_size = compressBound(size);
std::unique_ptr compressed = std::make_unique<unsigned char>(compressed_size);
compress(compressed.get(), &compressed_size, data, size);
return putOnWire(compressed.get(), compressed_size);
```

That being said, I think there's another issue here.

By default, zlib compresses data in the `zlib` format ([RFC 1950](https://datatracker.ietf.org/doc/html/rfc1950)), while the `gzip` content encoding expects data in the `gzip` format ([RFC 1952](http://tools.ietf.org/html/rfc1952)). I don't think there is an easy one-way zlib function like `compress()` that produces `gzip`-formatted data, but we could simply choose to handle the `deflate` content encoding instead, which quite conveniently expects `zlib`-formatted data.

tests/http_server_test.cc
```diff
@@ -855,3 +864,221 @@ TEST(http_server_test, client_multiple_requests_disconnects_handled)
...
+    ASSERT_TRUE(server.isBound());
+
+    // Log server coordinates...
+    const std::string server_address = "localhost:" + server.getPort().toString();
+    LOGGER("test", "Server address: " << server_address);
+
+    // Initialize client...
+
+        // Construct and initialize...
+        Http::Experimental::Client client;
```
Tachi107, line 969: Just wanted to say thanks for using our own http client for writing tests, so that we're more incentivized in making it better :)

tests/http_server_test.cc
```diff
@@ -855,3 +864,221 @@ TEST(http_server_test, client_multiple_requests_disconnects_handled)
...
+
+        // Storage for decompressed data...
+        std::vector<int> newlyDecompressedData;
+
+        // Keep decompressing bytes while there are some...
+        while(decompressor) {
+            int currentInteger = -1;
+            if(decompressor.read(reinterpret_cast<char *>(&currentInteger), sizeof(currentInteger)))
+                newlyDecompressedData.push_back(currentInteger);
+        }*/
```
Tachi107, line 1055: Probably forgot to remove these lines

Tachi107 (changes requested):
Some considerations about `setCompression()`:

1.  Having it as part of `ResponseWriter()` is more flexible than having a compression option in `Http::Endpoint::Options`, and since Pistache is not an httpd but a toolkit to build REST APIs, where each route can have its quirks, the ability to specify compression per-route (and per-request) is probably essential. For the record, nginx and Apache2 allow enabling compression as a global setting, much like the `Http::Endpoint::Options`. So well done with that!

2.  In my opinion, giving responsibility to the user to set the correct `Content-Encoding` header is undesirable. The [HTTP RFC](https://www.rfc-editor.org/rfc/rfc9110#field.content-encoding) is pretty clear:

    > If one or more encodings have been applied to a representation, the sender that applied the encodings MUST generate a Content-Encoding header field that lists the content codings in the order in which they were applied. Note that the coding named "identity" is reserved for its special role in [Accept-Encoding](https://www.rfc-editor.org/rfc/rfc9110#field.accept-encoding) and thus SHOULD NOT be included.

    There's no scenario where letting the user choose the `Content-Encoding` header for compression makes sense, so it should be handled inside the `setCompression()` function to make mistakes harder.

Tachi107:
And I've just noticed that the Debian testing clang package changed something in a manner that breaks our CI again... Going to investigate that after merging this patch.

kiplingw:
> Some considerations about `setCompression()`:
> 
>     1. Having it as part of `ResponseWriter()` is more flexible than having a compression option in `Http::Endpoint::Options`, and since Pistache is not an httpd but a toolkit to build REST APIs, where each route can have its quirks, the ability to specify compression per-route (and per-request) is probably essential. For the record, nginx and Apache2 allow enabling compression as a global setting, much like the `Http::Endpoint::Options`. So well done with that!

I agree. There's a lot of situations where the user will want compression on some endpoints and not on others. For instance an endpoint that spits out lots of JSON will compress nicely whereas an endpoint that spits out compressed audio or video is just going to waste CPU cycles trying to compress what won't compress well. This gives them more control.

>     2. In my opinion, giving responsibility to the user to set the correct `Content-Encoding` header is undesirable. The [HTTP RFC](https://www.rfc-editor.org/rfc/rfc9110#field.content-encoding) is pretty clear:
>        > If one or more encodings have been applied to a representation, the sender that applied the encodings MUST generate a Content-Encoding header field that lists the content codings in the order in which they were applied. Note that the coding named "identity" is reserved for its special role in [Accept-Encoding](https://www.rfc-editor.org/rfc/rfc9110#field.accept-encoding) and thus SHOULD NOT be included.
>
>        
>        There's no scenario where letting the user choose the `Content-Encoding` header for compression makes sense, so it should be handled inside the `setCompression()` function to make mistakes harder.

I was on the fence about this, but since you pointed out the RFC there's really only one correct expected behaviour for the server if it returns a compressed encoding. Fixed in today's PR.

tests/http_server_test.cc
kiplingw: It was a pain in the ass to figure out and it does have some weird quirks, but we got it working at least in the unit test.

src/common/http.cc
```diff
@@ -852,7 +865,50 @@ namespace Pistache::Http
...
+                   &compressedSize,
+                    reinterpret_cast<const unsigned char *>(data),
+                    size,
+                    contentEncodingDeflateLevel_);
+
+                    // Failed...
+                    if(compressionStatus != Z_OK)
+                        throw std::runtime_error(
+                            std::string("compress2() failed, returning: ") +
+                            std::to_string(compressionStatus));
```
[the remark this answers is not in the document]
kiplingw: Fixed.

src/common/http.cc
```diff
@@ -786,6 +795,10 @@ namespace Pistache::Http
         , buf_(DefaultStreamSize, other.buf_.maxSize())
         , transport_(other.transport_)
         , timeout_(other.timeout_)
+        , contentEncoding_(Http::Header::Encoding::Identity)
+#ifdef PISTACHE_USE_CONTENT_ENCODING_DEFLATE
+        , contentEncodingDeflateLevel_(Z_DEFAULT_COMPRESSION)
+#endif
```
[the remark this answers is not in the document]
kiplingw: Fixed.

kiplingw:
> [personal opinion] Please avoid adding a dependency on an external C++ zlib wrapper library.

We already removed that.

tests/http_server_test.cc
kiplingw: Fixed.

include/pistache/http.h
```diff
@@ -472,6 +479,20 @@ namespace Pistache
                 return nullptr;
             }
 
+            // Compress using the requested content encoding, if supported,
+            //  before sending bits to client. User responsible for setting
+            //  Content-Encoding header...
```
Tachi107, line 484: _"User responsible for setting Content-Encoding header"_ this is no longer the case

Tachi107 (commented):
Only one minor nitpick about a code comment. After that's addressed, and if Dennis isn't opposed to it, I'll merge. Good job Kip!

Tachi107:
And one more thing: if you could squash your commits so that commitlint passes it'd be great

include/pistache/http.h
kiplingw: Fixed.

kiplingw:
> And one more thing: if you could squash your commits so that commitlint passes it'd be great

I always forget how to do this. Can you remind me the command? I know what it does, I just can't remember the CLI for this whole PR.

Tachi107:
> I always forget how to do this. Can you remind me the command? I know what it does, I just can't remember the CLI for this whole PR.

It's `git rebase --interactive <first-commit-parent>`, and in this case the parent of this feature branch is `master`, so you can do:

    git rebase --interactive master

That'll open a text editor with something like this:

    pick 6e4135d Partial fix for #1148 by adding gzip content encoding support...
    pick 02e6255 contentEncoding renamed => contentEncoding_ to be consistent with class attribute naming convention...
    pick 9174955 Fixed some comments in content encoding unit test...
    pick 63cd6ed {meson.build, src/meson.build}: This PR is actually more correctly, as Andrea rightfully pointed out, implements support for deflate and not gzip, as similar as they may be... include/pistache/http.h: Added a setCompressionDeflateLevel() so user can change zlib's compression level... meson.build: Remove dependency on libhpptools-dev package which is not well maintained and use zlib directly... version.txt: Bumped versioning metadata...
    pick 0501399 src/common/http.cc: setCompression() now sets Content-Encoding header automatically, consistent with RFC 9110... {src/common/http.cc,include/pistache/http.h}: Use in-class initializer for contentEncoding_ and contentEncodingDeflateLevel_ members... include/pistache/{http,http_header,http_headers}.h: Ran clang-format... tests/http_server_test.cc: Ran clang-format...
    pick 57a7560 include/pistache/http.h: Clarify that Content-Encoding automatically set by setCompression()...

From there, you can change the "`pick`" words into "`squash`" (or "`s`" for short) of all the commits you want to squash, leaving `pick` only on the top one. Like this:

    pick 6e4135d Partial fix for #1148 by adding gzip content encoding support...
    squash 02e6255 contentEncoding renamed => contentEncoding_ to be consistent with class attribute naming convention...
    squash 9174955 Fixed some comments in content encoding unit test...
    squash 63cd6ed {meson.build, src/meson.build}: This PR is actually more correctly, as Andrea rightfully pointed out, implements support for deflate and not gzip, as similar as they may be... include/pistache/http.h: Added a setCompressionDeflateLevel() so user can change zlib's compression level... meson.build: Remove dependency on libhpptools-dev package which is not well maintained and use zlib directly... version.txt: Bumped versioning metadata...
    squash 0501399 src/common/http.cc: setCompression() now sets Content-Encoding header automatically, consistent with RFC 9110... {src/common/http.cc,include/pistache/http.h}: Use in-class initializer for contentEncoding_ and contentEncodingDeflateLevel_ members... include/pistache/{http,http_header,http_headers}.h: Ran clang-format... tests/http_server_test.cc: Ran clang-format...
    squash 57a7560 include/pistache/http.h: Clarify that Content-Encoding automatically set by setCompression()...

Now you can save and close the file. This will fire up another text editor from where you can edit the commit message of the single commit you'll leave on the branch. So you can write something like:

    # This is a combination of 6 commits.
    # This is the 1st commit message:
    
    feat: add deflate content encoding support
    
    This patch adds "Content-Encoding: deflate" support by using zlib. Users
    can enable compression per-route and specify the compression level too.
    Partial fix for #1148
    
    # This is the commit message #2:
    
    
    # This is the commit message #3:
    
    
    # This is the commit message #4:
    
    
    # This is the commit message #5:
    
    
    # This is the commit message #6:

Be sure to remove all the unwanted non-comment lines! And after saving and closing the file, you'll rebase will be complete.

kiplingw:
Ok done. Let me know how that looks.

@dennisjenkins75, also please have a look over if you can and test it out.

Tachi107:
> Ok done. Let me know how that looks

I believe you forgot to force-push

kiplingw:
> > Ok done. Let me know how that looks
> 
> I believe you forgot to force-push

I tried that:

```
$ git push --force
Everything up-to-date
```

Tachi107:
> I tried that:
> 
> ```
> $ git push --force
> Everything up-to-date
> ```

Maybe you didn't actually perform the rebase? What do `git show` and `git log`  print?

Tachi107:
It's probably GitHub's fault. My other PR isn't showing new commits too, and today GitHub had issues with pull requests: <https://www.githubstatus.com/incidents/l59z35rhzdky>

kiplingw:
Ok I think I got it to work this time, albeit the squash commit message could have been cleaner.

Tachi107 (approved):
Yeah the commit message isn't perfect, but you'll do better next time ;)

kiplingw:
So I just realized that there is a problem. The `setCompressionDeflateLevel()` method in the `ResponseWriter` in `http.h` is only publicly available if `PISTACHE_USE_CONTENT_ENCODING_DEFLATE` is defined by the preprocessor. It is defined during compile time of the library, but not for the user.

Before anyone tries to fix this, we should probably talk about how. @Tachi107?

Tachi107:
> So I just realized that there is a problem. The `setCompressionDeflateLevel()` method in the `ResponseWriter` in `http.h` is only publicly available if `PISTACHE_USE_CONTENT_ENCODING_DEFLATE` is defined by the preprocessor. It is defined during compile time of the library, but not for the user.
> 
> Before anyone tries to fix this, we should probably talk about how. @Tachi107?

Yes, I thought about this problem when working on my patch too. I was thinking about a more flexible method, where the user passes in a custom function or something similar which mutates the compression level, but I have nothing concrete yet.

kiplingw:
Yeah it needs some thinking. Or perhaps having some kind of generic way to set the compression level because the different schemes have different scales, such as a normalized 0 to 1 floating point range.

PR #1152, opened 2023-07-27
debian/control: Build with deflate content encoding enabled by default...

kiplingw:
Continues work on #1148...

debian/rules: Build with deflate content encoding enabled by default...
debian/changelog: Bumped versioning metadata...

### DIFF
--- a/changelog
+++ b/changelog
@@ -1,5 +1,5 @@
-pistache (0.2.1.20230726-1) UNRELEASED; urgency=low
+pistache (0.2.1.20230727-1) UNRELEASED; urgency=low
 
   * Bumped version.
 
- -- Kip Warner <kip@thevertigo.com>  Wed, 26 Jul 2023 23:39:00 +0000
+ -- Kip Warner <kip@thevertigo.com>  Thu, 27 Jul 2023 16:56:00 +0000
--- a/changelog
+++ b/changelog
@@ -1,5 +1,5 @@
-pistache (0.2.0.20230707-1) UNRELEASED; urgency=low
+pistache (0.2.1.20230726-1) UNRELEASED; urgency=low
 
   * Bumped version.
 
- -- Kip Warner <kip@thevertigo.com>  Sun, 09 Jul 2023 17:38:20 +0000
+ -- Kip Warner <kip@thevertigo.com>  Wed, 26 Jul 2023 23:39:00 +0000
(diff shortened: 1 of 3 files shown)


Tachi107:
> debian/control: Replace pkg-config with pkgconf, the former deprecated on Debian...

Is it? pkg-config is still the official interface to use, pkgconf is just the new implementation. As far as I know, people should be depending on pkg-config still

Tachi107:
Sorry, I wasn't clear. The pkg-config package on Debian is provided by pkgconf, as the "old" freedesktop.org implementation has been removed completely. Since most tools use the pkg-config interface (binary name, flags, etc.) it makes sense to me to depend on the general pkg-config package which itself pulls in the most desirable implementation (in Debian's case, right now, pkgconf) 

Il 27 luglio 2023 19:17:33 CEST, Kip ***@***.***> ha scritto:
>On Thu, 2023-07-27 at 03:47 -0700, Andrea Pappacoda wrote:
>> Is it? pkg-config is still the official interface to use, pkgconf is
>> just the new implementation. As far as I know, people should be
>> depending on pkg-config still
>
>They're mostly interchangeable. pkgconf is a drop in replacement for
>pkg-config. The latter is deprecated on Debian[0] and Fedora[1] now.
>The former is actively maintained. 
>
>There's a number of reasons for it, but one of them is pkgconf has no
>glib dependency. 
>
>Users shouldn't need to make any changes to their build environment at
>all, other than to ensure pkgconf is installed.
>
>[0] <https://debconf22.debconf.org/talks/86-pkg-config-pkgconf-bof/>
>[1] <https://fedoraproject.org/wiki/Changes/pkgconf_as_system_pkg-config_implementation>
>

kiplingw:
On Thu, 2023-07-27 at 10:29 -0700, Andrea Pappacoda wrote:
> Sorry, I wasn't clear. The pkg-config package on Debian is provided
> by pkgconf, as the "old" freedesktop.org implementation has been
> removed completely. Since most tools use the pkg-config interface
> (binary name, flags, etc.) it makes sense to me to depend on the
> general pkg-config package which itself pulls in the most desirable
> implementation (in Debian's case, right now, pkgconf) 

Good call. Fixed.

-- 
Kip Warner
OpenPGP signed/encrypted mail preferred
https://www.thevertigo.com

PR #1153, opened 2023-07-28
ci(linux): fix Debian Testing Clang jobs

Tachi107:
Work in progress.


Tachi107:
This is a temporary issue caused by Debian's GCC 13 transition, it's not worth a workaround since it'll be fixed in a couple of days

PR #1140, opened 2023-06-28
refactor!: switch to sockaddr_storage, IPv6 and resolver improvements

Tachi107:
Using a `struct sockaddr_storage` allows us to get rid of the spooky union of `sockaddr_in` and `sockaddr_in6`, and to do IPv4 and IPv6 as recommended by RFC 3493.

The code is not that cleaner, but it should be more robust and less "dangerous". As discussed recently on IRC, Kip and I were quite [puzzled by the various constructors of class IP](https://github.com/pistacheio/pistache/blob/e884d6d198593716be0441fc28e271993556a302/src/common/net.cc#L155-L161), as it wasn't immediatly clear why they were `memcpy`ing `INET_ADDRSTRLEN` bytes into a smaller `s_addr` array.

Needless to say, this is a breaking change. I also removed a few now useless functions, but they could be kept if desired.

Note: this is still experimental and incomplete.

PR #1155, opened 2023-07-31
Unix domain socket support

win32asm:
Adds support for unix domain listener sockets.

Makes minimal changes to do so, avoiding the temptation to refactor the IP and Address classes.

To bind a unix domain listener one can either use an Address constructors that takes a string or char* argument or use the Address::fromUnix factory method with a struct sockaddr_un* argument to obtain an Address that can be used to create an Endpoint.  In either case, the argument must conform to the criteria listed in the Address Format section of the unix(7) manual page.

Adds clauses to code interrogating address family values to handle AF_UNIX.

Code using unix domain addressing shouldn't care about ports or hostnames.  Rather than throw exceptions, queries for these values return 0 and "localhost" respectively.

Refactors Listener::bindListener to bypass calls to getaddrinfo for unix domain addresses.

Adds an addrLen_ member to the Address class to allow setting the proper address length value in calls to bind(2) for unix domain sockets.


win32asm:
> Thanks @win32asm. I'm not very familiar with UNIX domain sockets. Could you let us know what the principle benefit or benefits are to the user?

Access to unix domain sockets is provided based on the socket file properties/permissions, in a regular unix fashion - instead of firewall rules.
Otherwise it`s an analog of having multiple distinct ways to address the server running on the same (localhost) system.

src/common/peer.cc
```diff
@@ -68,22 +68,34 @@ namespace Pistache::Tcp
...
-            sa.sin_family = AF_INET;
-            if (inet_pton(AF_INET, addr.host().c_str(), &sa.sin_addr) == 0)
+            if (addr.family() == AF_UNIX)
             {
-                hostname_ = addr.host();
+                //
+                // Communication through unix domain sockets is constrained to
+                // the local host.
+                //
+                hostname_.assign("localhost");
```
dennisjenkins75, line 77: I have slight concerns about this.  Maybe return an empty string, or a value that is NOT a valid hostname, like "**unix" or similar.

win32asm: "[Unix]" ?
We already use "[Abstract]" for linux abstract sockets.
However, "localhost" matches the expectations as in - all communication for unix domain sockets happens within localhost.
Any idea how that could be used in the way other than for presentation purposes?

src/common/net.cc
```diff
@@ -337,6 +370,41 @@ namespace Pistache
 
     void Address::init(const std::string& addr)
     {
+        // Handle unix domain addresses separately.
+        if (isUnixDomain(addr))
+        {
+            struct sockaddr_un unAddr;
```
Tachi107, line 376: Please initialize any `struct sockaddr_un` used. Citing the [POSIX standard](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_un.h.html):

> Although applications are required to initialize all members (including any non-standard ones) of a sockaddr_in6 structure (see [<netinet/in.h>](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/netinet_in.h.html)), the same is not required for the sockaddr_un structure, since historically many applications only initialized the standard members. Despite this, applications are encouraged to initialize sockaddr_un structures in a manner similar to the required initialization of sockaddr_in6 structures.

You should use `struct sockaddr unAddr = {}`, as per POSIX:

> Although it is common practice to initialize a sockaddr_in6 structure using:
> 
>     struct sockaddr_in6 sa;
>     memset(&sa, 0, sizeof sa);
> 
> this method is not portable according to this standard, because the structure can contain pointer or floating-point members that are not required to have an all-bits-zero representation after default initialization. Portable methods make use of default initialization; for example:
> 
>     struct sockaddr_in6 sa = { 0 };

`= {}` is recommended as opposed to `= { 0 }` since its a bit safer to use than the latter. For details, see C23's [N2900](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n2900.htm) accepted proposal.

src/common/net.cc
```diff
@@ -337,6 +370,41 @@ namespace Pistache
...
+            unAddr.sun_family = AF_UNIX;
+
+            // See unix(7) manual page; distinguish among unnamed, abstract,
+            // and pathname socket addresses.
+            auto size = std ::min(addr.size(), sizeof unAddr.sun_path);
```
Tachi107, line 381:
```suggestion
            const auto size = std::min(addr.size(), sizeof(unAddr.sun_path));
```

src/common/net.cc
```diff
@@ -337,6 +370,41 @@ namespace Pistache
...
+            if (size == 0)
+            {
+                addrLen_ = sizeof unAddr.sun_family;
+            }
+            else if (addr[0] == '\0')
+            {
+                addrLen_ = static_cast<socklen_t>(
+                    sizeof unAddr.sun_family + size);
+                ::memcpy(unAddr.sun_path, addr.data(), size);
```
Tachi107, line 390: If you really want to specify the namespace of `memcpy()` and `strncpy()` then use `std::memcpy`, since technically speaking the `cstdlib` header only provides functions in the `std` namespace

src/server/endpoint.cc
```diff
@@ -88,7 +88,7 @@ namespace Pistache::Http
             if (entry.getTag() == Polling::Tag(timerFd))
             {
                 uint64_t wakeups;
-                ::read(timerFd, &wakeups, sizeof wakeups);
+                [[maybe_unused]] auto rv = ::read(timerFd, &wakeups, sizeof wakeups);
```
Tachi107, line 91: Why? If you want to ignore the result value, just cast the result to `void`

tests/listener_test.cc
```diff
@@ -241,6 +245,35 @@ TEST(listener_test, listener_bind_ephemeral_v6_port)
     ASSERT_TRUE(true);
 }
 
+TEST(listener_test, listener_bind_unix_domain)
+{
+    // Avoid name conflict by binding within a fresh temporary directory.
+    using charArray        = char[];
```
Tachi107, line 251: I'm not sure this type alias makes the code more readable

tests/listener_test.cc
```diff
@@ -241,6 +245,35 @@ TEST(listener_test, listener_bind_ephemeral_v6_port)
...
+    charArray nameTemplate = "/tmp/bind_test_XXXXXX";
+    auto tmpDir            = ::mkdtemp(nameTemplate);
+    ASSERT_TRUE(tmpDir != nullptr);
+    auto ss = std::stringstream();
+    ss << tmpDir << "/unix_socket";
+    auto sockName = ss.str().c_str();
+
+    struct sockaddr_un sa;
```
Tachi107, line 259:
```suggestion
    struct sockaddr_un sa = {};
```

Tachi107 (changes requested):
Hi @win32asm, thank you very much for this patch, it looks really interesting! What are your use cases for this? The first thing that comes to my mind is to use a unix socket when putting a Pistache service behind a reverse proxy, since there's no need to use TCP there, but I'm interested in hearing your feedback.

> Makes minimal changes to do so, avoiding the temptation to refactor the IP and Address classes.

What refactor do you have in mind? After my recent patch in #1140 I've been thinking a lot about changing the design of the two classes, since I find them a bit messy. It's nice to know that somebody sees them as I do :)

I have left some feedback below.

glenn-skinner:
I didn't see a place to reply to @Tachi107's inital comment above, so I'll do it here.

Here's some motivation (it meshes with the reverse proxy comment above):

---
unix domain sockets provide a host-local communication mechanism that provides socket semantics.  Since communication is confined to the local host, many potential security problems become non-issues.  Naming is done through the file system name space rather than through IP addresses.

(Additionally, one can use unix domain sockets to pass state from one process to another that is well-defined only within the scope of the local host.  The canonical example of such state is file descriptors; for example, one can open a file and pass a descriptor for it to a peer process that might not itself have permission to do so.)

The specific use case that’s motivated us to add unix domain socket support to Pistache concerns the VMware ecosystem.  VMware provides APIs to allow third party vendors (such as us) to interpose on IO operations made by virtual machines running on ESX hosts, allowing those vendors to provide services such as caching, backup and migration, and disaster recovery.  These services run partially within the target VMs and partially in separate processes running as appliances, either on-host or off-host.

Communication with these appliances is structured as REST interfaces, which is where Pistache enters the picture.  VMware has recently chosen to funnel all off-host communication through a proxy that makes the peer available as a unix domain socket.  To maintain the REST communication abstraction we need a way to allow our server to bind to a unix domain socket listener that the proxy can connect to.
---

As far as class refactoring is concerned, if left with a free hand, I think the designers of the Unix socket addressing structs largely got it right and I would restructure to mirror them.  I think the Address class should be thought of as roughly equivalent to struct_sockaddr, providing services that are common to all addresing schemes (or address families in socket terminology).

At the implementation level, I would push the addr_ member currently in the IP class (a struct sockaddr_storage) up into the Address class.  The IP class would then either become a subclass of Address or have an Address member.  (I haven't thought about it enought to decide whether an is-a of a has-a relationship would work better, although I'm leaning toward is-a.)  It would rely on Address to store its socket address.  There would be another class sitting alongside IP for unix domain addresses. That refactoring would remove some of the awkwardness present in the current patch request.

I haven't had a chance to look at more specific comments yet.  I'll start working through them today.

src/common/net.cc
```diff
@@ -319,14 +352,14 @@ namespace Pistache
 
     Address Address::fromUnix(struct sockaddr* addr)
     {
-        if ((addr->sa_family == AF_INET) or (addr->sa_family == AF_INET6))
+        auto family = addr->sa_family;
+        if (family == AF_INET or family == AF_INET6 or family == AF_UNIX)
```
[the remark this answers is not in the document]
glenn-skinner: Will do.

src/common/net.cc
```diff
@@ -163,6 +171,12 @@ namespace Pistache
         {
             return ntohs(reinterpret_cast<const struct sockaddr_in6*>(&addr_)->sin6_port);
         }
+        else if (addr_.ss_family == AF_UNIX)
+        {
+            // Ports are a meaningless concept for unix domain sockets.  Return
+            // an arbitrary value.
+            return 0;
```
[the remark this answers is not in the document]
glenn-skinner: I agree about the class messiness (see my comment elsewhere).  Provided that we devise a good way to specify a unix domain address, the low-level mechanism for recording that address (a sockaddr_un) should be an implementation detail.  My first attempt at finding a way to set up a unix domain address was to piggyback on the Address::fromUnix method, as that exposed a struct sockaddr to use.  Then @win32asm and I realized that we could apply heuristics to the string Address constructor to do that job.  Perhaps the fromUnix method should be removed.

As to your original point, I have mixed feelings about throwing an exception.  As you suggested, refactoring to avoid the issue altogether is a better solution.  In the absence of that, I'm willing to go along with reviwer's consensus.  Does anyone else have an opinion here?

include/pistache/net.h
```diff
@@ -146,28 +153,57 @@ namespace Pistache
...
         std::string host() const;
         Port port() const;
         int family() const;
 
+        // Returns the address length to be used in calls to bind(2); valid
+        // only when family() == AF_UNIX.
+        socklen_t addrLen() const
+        {
+            return addrLen_;
+        }
```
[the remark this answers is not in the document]
dennisjenkins75: Personally, I have mixed feelings on the use of C++ exceptions in a library like Pistache; but they existed here before any of the current maintainers were ever involved.  However, in this specific case, I would opt to just return a 0 for addrlen if the address family is not AF_UNIX.  Throwing an exception in a getter seems odd.  Its fine for the docs to state "results undefined when family != AF_UNIX" and then do ${whatever}.  So in the end, I do not have a strong opinion here.  I've not looked at this code today: Is "addrLen_" even initialized for other address families?  If not, that is a bug, and a static analyzer should flag it.

include/pistache/net.h
```diff
@@ -146,28 +153,57 @@ namespace Pistache
...
+        // These two constructors can be used to create unix domain addresses
+        // as well as IP-based addresses.  If the addr argument meets one of the
+        // criteria for naming a unix domain address, it will generate such an
+        // address.  Note that matching such a criterion implies that addr would
+        // be invalid as an IP-based address.
+        //
+        // The criteria are:
+        // - addr is empty
+        // - addr[0] == '\0'
+        // - addr contains a '/' charcter
```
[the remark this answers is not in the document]
glenn-skinner: Will do.

include/pistache/net.h
```diff
@@ -146,28 +153,57 @@ namespace Pistache
...
         Address(IP ip, Port port);
 
         Address(const Address& other) = default;
         Address(Address&& other)      = default;
 
         Address& operator=(const Address& other) = default;
-        Address& operator=(Address&& other) = default;
+        Address& operator=(Address&& other)      = default;
 
+        // Supports the AF_UNIX address family as well as AF_INET and AF_INET6.
```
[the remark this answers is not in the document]
glenn-skinner: Reworded to "Supports the AF_INET, AF_INET6, and AF_UNIX address families."  Comment style modified.

include/pistache/net.h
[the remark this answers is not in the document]
glenn-skinner: An alternative would be to extend it so supply address lengths for AF_IP and AF_IP6 sockets.  (They're constant sizeof ... values.)  Would you prefer that?

src/common/net.cc
glenn-skinner: Fixed the space before :: and the missing auto.  But the formatting for sizeof triggers one of my pet peeves.  sizeof is not a function; it's a keyword operator and keywords should be followed by a space.  Omitting parens around its operand make it clear that the operand is an object as opposed to a type.

Having said that, if your style guide says that it should be formatted as you've suggested, I'll do so.

src/server/endpoint.cc
glenn-skinner: The compiler on one of our build systems (I don't remember whether it was a recent version of g++ or clang) was not happy about the cast.  The only way I could find to avoid the warning was to add the attribute.

src/common/net.cc
glenn-skinner: Done.

include/pistache/net.h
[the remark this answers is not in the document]
glenn-skinner: I've modified it accordingly.

src/common/net.cc
```diff
@@ -171,9 +185,28 @@ namespace Pistache
...
+        {
+            auto& unAddr = reinterpret_cast<const struct sockaddr_un&>(addr_);
+            if (unAddr.sun_path[0] == '\0')
+            {
+                // The socket is abstract (not present in the file system name
+                // space).  Its name starts with the byte following the initial
+                // NUL.  As the name may contain embedded NUL bytes and its
+                // length is not available here, simply note that it's an
+                // abstract address.
+                return std::string("[Abstract]");
```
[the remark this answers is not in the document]
glenn-skinner: Supporting Linux abstract sockets doesn't preclude extending the code base to additional platforms.  If one should attempt to create an abstract socket on a platform that doesn't support them, the bind() system call will fail and the error will propagate upward.  But no additional harm is done; it's simply a runtime error.

As a usefulness data point, it turns out that the VMware user world implementation of unix domain sockets treats the pathname variant as if it were abstract.  That is, the implementation doesn't register them in the file system name space.  (This is a security measure of dubious utility to hide sockets from attackers.)

Let me add a request for adding portability to epoll-like interfaces.  It also turns out that VMware's user world emulation of Linux interfaces omits the timerfd_* functions, but does include epoll.  That means that we won't be able to use this patch in our product without persuading VMware to add these functions to their emulation.  (That will be an uphill battle, but it will be greatly aided by this patch being accepted.  In the interim we're forced to use a hacked version of Pistache predating its use of the timerfd_* functions that supports only unix domain sockets.  This situation disgusts us and we want to fix it.)

src/common/net.cc
glenn-skinner: Fixed.

tests/listener_test.cc
glenn-skinner: Fixed.

It turns out that line 257 was the cause of the test failures. The call to str() created a temporary that the c_str() call depended on.  But the temporary didn't survive long enough.

tests/listener_test.cc
glenn-skinner: Reverted to C-style array declaration syntax

kiplingw: If possible, try std::array.

kiplingw:
Great work @glenn-skinner. You've got a very interesting use case for us in production code.

tests/listener_test.cc
kiplingw: We currently don't use `std::filesystem::path` in our unit tests, but you're certainly welcome to if you think it will help.

src/common/net.cc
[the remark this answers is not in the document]
Tachi107: To be clear: I kinda hate exceptions, but most other methods currently throw one in cases like these. But I don't have a strong preference.

tests/listener_test.cc
glenn-skinner: Yes, that's doable.  Changed accordingly.

glenn-skinner:
I think I've resolved all review comments.  Please let me know if I've missed any or if the resolution is unsatisfactory.

Is the ABI chnage worrisome?

Is it time to remove draft status from this pull request?

Tachi107 (approved):
Looks good to me!

Thank you @win32asm and @glenn-skinner very much for this patch, I really appreciate it.

With regards to ABI breakage, this isn't a big concern for me. I'll become more important once we tag a stable release.

As for the class refactoring, I think you're right. I'll look into it in the future, keeping your feedback as a reference.

Thanks again! If there are no objections, I'll merge.

kiplingw:
LGTM

Tachi107:
Merging. Thanks again!

PR #1162, opened 2023-08-29
build: Meson options for custom OpenSSL lib/hdr path

win32asm:
Required in case the openssl library is not native to the build environment (e.g. part of vcpkg fetched dependencies).


kiplingw:
Thanks @win32asm. Can you also remember to [document](https://github.com/pistacheio/pistache/blob/master/README.md#building-from-source) the proposed new build flags in our `README.md`?

kiplingw (approved):

kiplingw:
I'm fine with it. @dennisjenkins75, @Tachi107?

Tachi107:
I'm not a fan of this workaround. Can't you simply override the pkg-config search directories to include your custom OpenSSL build?

win32asm:
> I'm not a fan of this workaround. Can't you simply override the pkg-config search directories to include your custom OpenSSL build?

Let us see if we can use a pkg-config-based detection way.

kiplingw:
> I'm not a fan of this workaround. Can't you simply override the pkg-config search directories to include your custom OpenSSL build?

That's actually a really good point.

Tachi107:
Nice to hear! Happy to help :)

PR #1165, opened 2023-10-22
feat: add Request::getBestAcceptEncoding() function

Tachi107:
The new getBestAcceptEncoding() function makes using the new compression support introduced in commit b0b7d3076dcdba48935cfe14049b76809d0cd256 way easier, as it reports to the caller the most sensible content encoding to apply to the response, falling back to the Identity encoding (i.e. no actual compression) in case no encoding is supported by both the server and the client.

Fixes https://github.com/pistacheio/pistache/issues/1148


kiplingw (approved):
Solid work @Tachi107. LGTM. @dennisjenkins75, have a look over. If you're fine with it lets merge.

Tachi107:
Added some more tests and fixed a corner case discovered with the newly added tests. Should be ready now :)

kiplingw:
@Tachi107, if your PR passes in CI and @dennisjenkins75 has had a chance to review, I'm totally fine with it. Only thing I'd suggest changing is to bump the patch version.

Tachi107:
> Only thing I'd suggest changing is to bump the patch version.

Done

Tachi107:
Debian testing CI runs are failing the Coverage step because of Debian bug <https://bugs.debian.org/1053252>, it's unrelated to this patch set.

include/pistache/http.h
```diff
@@ -217,6 +217,16 @@ namespace Pistache
 
             std::chrono::milliseconds timeout() const;
 
+            /*
+             * Returns the "best" encoding to use to encode (tipically compress)
```
dennisjenkins75, line 221: "typically"

dennisjenkins75 (changes requested):

include/pistache/http.h
Tachi107: Thanks, fixed!

Tachi107:
Thanks for the review and merge :)

PR #1176, opened 2023-11-30
fix: make zlib a private dependency

Tachi107:
Before this patch, the `zlib.h` header was included by `pistache/http.h`, meaning that users had to install the development headers on zlib to build projects which depend on Pistache (in other terms, it was a transitive, public, API dependency). Since `zlib.h` was included by http.h just to get the value of zlib's default compression level, I simply hardcoded it to `6`; it's unlikely to ever change.


kiplingw:
But is this a good idea? Is there any particular reason why the include needs to be removed? Was it having a noticeable impact on preprocessor burden at compile time? You are right that the constant is unlikely to change, but it still could. zlib has gone through [many changes](https://www.zlib.net/ChangeLog.txt) and likely still will.

Tachi107:
Adding a normalized 1..0 value independent of the underlying compression algorithm used seems like a better solution (which still needs to be implemented). Closing.

PR #1178, opened 2023-12-03
Add Brotli compression support…

kiplingw:
Adds Brotli compression support. Continues #1148...


include/pistache/http.h
```diff
@@ -494,6 +498,15 @@ namespace Pistache
             //  automatically set to the requested encoding, if supported...
             void setCompression(const Pistache::Http::Header::Encoding _contentEncoding);
 
+#ifdef PISTACHE_USE_CONTENT_ENCODING_BROTLI
+            // Set the compression level for Brotli algorithm. Defaults to
+            //  BROTLI_DEFAULT_QUALITY...
+            void setCompressionBrotliLevel(const int _contentEncodingBrotliLevel)
```
Tachi107, line 504: Before adding new functions which are conditionally compiled with `#ifdef`s, I'd like to see if there's a better way to support custom compression levels. For example, if I understand correctly, a user cannot enable both Brotli and Deflate compression for the same request, so having two `contentEncoding*Level_` variables for request surely can be avoided, right?

Tachi107 (changes requested):

include/pistache/http.h
Tachi107: As an aside, having an underscore before the parameter name seems odd to me, and if I recall correctly we don't do it anywhere else, so `_contentEncodingBrotliLevel` should be `contentEncodingBrotliLevel`

kiplingw: Hey @Tachi107. The underscore isn't really my preference either. I was just trying to emulate the existing convention that appeared to be in effect.

Yes, you're correct the user can't enable multiple compression schemes, but can only request them with the intention being only one of them will be used (e.g. via q-values).

I agree that there should be a generic way to support compression levels across different codecs. I think one proposal is to have a normalized scale of [0.0, 1.0] where 0 is lowest and 1 is maximum compression.

meson.build
```diff
@@ -55,6 +55,19 @@ if get_option('PISTACHE_USE_RAPIDJSON')
...
+# Support Brotli compressed Content-Encoding responses...
+if get_option('PISTACHE_USE_CONTENT_ENCODING_BROTLI')
+
+    # Need Brotli encoder for library...
+    brotli = dependency('libbrotlienc')
+    deps_libpistache += brotli
+
+    # ...decoder for unit testing...
+    brotli = dependency('libbrotlidec')
+    deps_libpistache += brotli
```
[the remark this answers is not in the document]
kiplingw: I'm not sure how to refactor `meson.build` / `tests/meson.build`, but if you can fix that we can get this done. It's only a few lines. I'm not very familiar with meson.

PR #1181, opened 2023-12-15
Initial version for BSD support

alex0x08:
So below is the initial version, just to observe and get feedback. 
I'll do more work on testing soon (and this pull-request will be updated). Also need to check SSL, it compiles but failing on tests.


meson.build
```diff
@@ -51,6 +51,13 @@ deps_libpistache = [
 	date_dep
 ]
 
+if host_machine.system() in ['freebsd', 'netbsd', 'openbsd', 'dragonfly']
```
kiplingw, line 54: I recommend not using this approach. The best practice philosophy for portability is to test for specific features on whatever platform the user might be on and not test for the platforms themselves. So if a necessary function is suspected to be available on a particular platform and not on others, test for the presence of that function rather than testing for the platform. Otherwise the assumption about the platform might be wrong, or right today but wrong tomorrow in a subsequent release of the platform. 

I think what you want to do here is test for and set a macro for HAVE_BSD_SENDFILE, HAVE_BSD_SOCKTOOPT, HAVE_BSD_WAIT, which is typically what you do with autoconf.

kiplingw (changes requested):

meson.build
alex0x08: Done, anyway I planned to remove it, kept only as example.

kiplingw: It looks like you are still checking for the platform rather than the presumed functions themselves.

alex0x08: There is no simple way to check for such availability, at least I don't know such.  For example the 'wait()' function requires dedicated header file on FreeBSD, but function itself is in libc. Another story with 'sendfile()', which exists on FreeBSD too but has incompatible signature.

kiplingw: @alex0x08, you need to look for whatever the Meson equivalents to `AC_CHECK_FUNC` and `AC_CHECK_HEADER` are and use those.

alex0x08: Hmm..  found this one:
`
compiler.has_function('iconv_open', prefix : '#include <iconv.h>')
`
but looks like there is no exact equivalents to AC_CHECK_FUNC and AC_CHECK_HEADER:
https://github.com/mesonbuild/meson/issues/437
https://github.com/mesonbuild/meson/issues/292

I could try to adopt my check for wait() this way, but no idea how to deal with sendfile(), which presents in both Linux and BSD but with different signature.

kiplingw: You will want to use whatever the equivalent of `AC_RUN_IFELSE`, `AC_LANG_PROGRAM`, and probably `AC_DEFINE_UNQUOTED`. Example for Autoconf:

```m4
    # Check for system provided π constant...
    AC_MSG_CHECKING([whether system's cmath defines M_PI])
    AC_COMPILE_IFELSE(
        [AC_LANG_PROGRAM([#include <cmath>], [ double f = M_PI; (void)(f);])],
        [AC_MSG_RESULT([yes])],
        [
            # Alert user we're falling back...
            AC_MSG_RESULT([no, using our own])

            # This trick should work in any programming language...
            m_pi_workaround="(4 * ::atan(1.0))"
            AC_DEFINE_UNQUOTED(
                [M_PI],
                [$m_pi_workaround],
                [Workaround if no system definition for M_PI...])
            AC_SUBST([M_PI], [$m_pi_workaround])
        ])
```

kiplingw:
@dgreatwood, is this PR moot now given your macOS changes?

dgreatwood:
@kiplingw, yes, superseded by the libevent support and specifically by PR #1221 (FreeBSD, OpenBSD and NetBSD support), merged on Aug 16 2024.

PR #1182, opened 2024-01-04
Fix missing compiler flags in pkgconfig manifest needed for #1148...

kiplingw:
Fix missing compiler flags in pkgconfig manifest needed for #1148...


dennisjenkins75 (approved):
I do not know how Meson works.  I will assume that your changes are good.

Tachi107:
Looks good! Could be cleaned up a bit, but I'll do it together with other meson cleanups in a future patch :)

kiplingw:
Thanks @Tachi107. I agree that there is likely a way to improve it. I was trying to find a way to get the already defined compiler flags to put in `extra_cflags` by querying Meson, rather than manually tracking them, but I couldn't seem to find it.

PR #1183, opened 2024-01-06
ci(linux): workaround Debian bug #1053252

Tachi107:
Fixes jobs on Debian Testing. See <https://bugs.debian.org/1053252>.


Tachi107:
lcov's Debian maintainer [just merged my fix](https://salsa.debian.org/mckinstry/lcov/-/merge_requests/1) and [uploaded a new package version](https://bugs.debian.org/1053252#15), meaning that this bug is going away in ~10 days. It doesn't make sense to merge it anymore :)

PR #1188, opened 2024-01-13
ci(linux): small coverage fixes

Tachi107:
This PR finally fixes the CI Coverage step on Debian Testing, and also switches from gpgv to a sop implementation. See the individual commit messages for details :)


Tachi107:
...and this doesn't actually fix the coverage step on Debian Testing, lcov 2.0 is creating more issues than expected. I'll try to see if we can ditch it and directly use gcov instead.

PR #1191, opened 2024-02-12
Improve fuzzer

tyler92:
The current fuzzer target covered only basic things and didn't cover the most interesting parts of HTTP request parsing. I've extended fuzzer with more targets and added a corpus. As a result, coverage has been increased significantly and several memory issues have been found; I will report them separately to avoid mixing them with this MR.

Notes:

1. Due to the reuse policy, I had to add ".license" files for each file in the corpus. I'm unsure if it's a good idea and would appreciate your suggestions. 
2. The "router" target can't be covered via public interface because the Router class requires dependencies that send messages over a network. So Router is covered via the SegmentTreeNode class.
3. The 'removeRoute' method is excluded from the fuzzer due to crashes (there are no null pointer checks). Probably we can enable it later when it's fixed


kiplingw:
Good work @tyler92. We'll get back to you as soon as we've had a chance to review.

Tachi107:
Hi @tyler92, thanks for the patch! Fuzzing is definitely an interesting area where we could do better.

> Due to the reuse policy, I had to add ".license" files for each file in the corpus. I'm unsure if it's a good idea and would appreciate your suggestions.

What you can do instead is add a `.reuse/dep5` file in the project root which contains licensing information for all those small files, for example:

```dep5
Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/

Files: tests/fuzzers/corpus/http-*.txt
Copyright: 2024 Mikhail Khachayants
License: Apache-2.0

tyler92:
@Tachi107 Thanks for the hint with `.reuse/dep5`, it works!

tyler92:
@kiplingw Just in case - there are no more changes expected on my part, the MR is ready for review. Please let me know if something has to be fixed or improved.

PR #1204, opened 2024-04-14
Streaming test - Avoid memory-leak github test failure report

dgreatwood:
Fixed issue of apparently leaking Fds (per github test runners) in the test StreamingTest.ClientDisconnect. Absence the fix, the leak is visible in libevent mode.


kiplingw:
Great work again @dgreatwood.

Issue #1193, opened 2024-02-12
Out of bound read in std::strtol while parsing HTTP requests


tyler92:
The `std::strtol` requires the input string to be a zero-terminated string, but the HTTP parsing procedure works with a binary buffer and there is no guarantee that the last byte is zero. It may lead to out-of-bound read, which is an undefined behavior and might cause it to crash. The list of affected functions:

```
http.cc:291          State ResponseLineStep::apply(StreamCursor& cursor)
http.cc:460          BodyStep::Chunk::Result BodyStep::Chunk::parse(StreamCursor& cursor)
http_header.cc:229   void CacheControl::parseRaw(const char* str, size_t len)
```

Example of sanitizer report used in [fuzzing test](https://github.com/pistacheio/pistache/pull/1191):

```
==2060939==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x50c0008718c0 at pc 0x55892bade8af bp 0x7ffeeea689b0 sp 0x7ffeeea68170
READ of size 4 at 0x50c0008718c0 thread T0
    #0 0x55892bade8ae in StrtolFixAndCheck(void*, char const*, char**, char*, int) asan_interceptors.cpp.o
    #1 0x55892baca271 in strtol (pistache/build/tests/fuzzers/fuzz_parser+0x1d8271) (BuildId: b2cb0c20f13ffb84cf16359224e2ad7c4dcc6bc8)
    #2 0x55892bb6f65f in Pistache::Http::Private::BodyStep::Chunk::parse(Pistache::StreamCursor&) pistache/src/common/http.cc:460:35
    #3 0x55892bb6e85f in Pistache::Http::Private::BodyStep::parseTransferEncoding(Pistache::StreamCursor&, std::shared_ptr<Pistache::Http::Header::TransferEncoding> const&) pistache/src/common/http.cc:507:44
    #4 0x55892bb6d8db in Pistache::Http::Private::BodyStep::apply(Pistache::StreamCursor&) pistache/src/common/http.cc:397:24
    #5 0x55892bb7075a in Pistache::Http::Private::ParserBase::parse() pistache/src/common/http.cc:545:36
    #6 0x55892bb378c3 in fuzz_request_parser(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>> const&)::$_0::operator()() const pistache/tests/fuzzers/fuzz_parser.cpp:82:48
[...]

PR #1208, opened 2024-04-26
Fix for out of bound read in std::strtol while parsing HTTP requests

alexprabhat99:
Fixes: #1193 
This is an attempt to better a previous PR #1201.
@kiplingw This PR is ready for review. Thanks!


kiplingw:
@Tachi107, I can't tell, but are these CI failures innocuous?

src/common/http.cc
```diff
@@ -287,9 +288,18 @@ namespace Pistache::Http
...
-            if (*end != ' ')
+            const char* beg = codeToken.rawText();
+            const char* beg_trim = beg;
+            while (*beg_trim == '+' || *beg_trim == ' ')
+            {
+                if (*(beg_trim + 1))
+                    ++beg_trim;                    
+            }
+            std::string_view view(beg_trim, codeToken.size());
+            const char* end = view.data() + view.size();
```
Tachi107, line 299: Why is a string_view being used here?

Tachi107: Citing cppreference:

>     constexpr basic_string_view( const CharT* s, size_type count );
>
> After construction, [data()](https://en.cppreference.com/w/cpp/string/basic_string_view/data) is equal to s, and [size()](https://en.cppreference.com/w/cpp/string/basic_string_view/size) is equal to count.

So isn't your code exactly the same as:

```suggestion
            const char* end = beg_trim + codeToken.size();
```

src/common/http.cc
```diff
@@ -455,10 +465,18 @@ namespace Pistache::Http
...
-                if (*end != '\r')
+                const char* beg = chunkSize.rawText();
+                const char* beg_trim = beg;
+                while (*beg_trim == '+' || *beg_trim == ' ')
+                {
+                    if (*(beg_trim + 1))
+                        ++beg_trim; 
+                }
+                std::string_view view = beg_trim;
+                const char* end = view.data() + view.size();
```
Tachi107, line 476: I'm generally a fan of string_view, but here I personally find it unnecessary. I'd suggest using `strlen()` explicitly, like:

```suggestion
                const char* end = beg_trim + std::strlen(beg_trim);
```

Tachi107 (changes requested):
This patch is supposed to be fixing an out of bounds read due to using `std::strtol` on a non-`\0`-terminated string, but unless I'm reading this wrong the bug is still present; instead of over-reading with `strtol`, you're over-reading with `std::from_chars`, since the `end` pointer is derived by `beg_trim + strlen(beg_trim)`, which will point to beyond the end of the string when it isn't null-terminated since `strlen` looks for the string terminator to determined the size.

If you have a char pointer and the pointed data isn't null-terminated *and* you don't even have a length/size variable, then you really have no way of determining its length.

Tachi107:
Il giorno ven 26 apr 2024 alle 10:52:45 -07:00:00, Kip 
***@***.***> ha scritto:
> @Tachi107, I can't tell, but are these CI failures innocuous?

It seems an error in the runners themselves, so not related to this 
patch in particular. I've just rescheduled them, and they should 
hopefully pass.

-- 
OpenPGP key: 66DE F152 8299 0C21 99EF  A801 A8A1 28A8 AB1C EE49

kiplingw:
Fixed with #1251.

PR #1209, opened 2024-04-27
fix(server): Crash if remove not existing route

tyler92:
One more crash was found [by fuzzer](https://github.com/pistacheio/pistache/pull/1191) in `router.cc`. I've added unit tests that cover the "removeRoute" function, uncommented a corresponding line in the fuzzer, and renamed corpus files to their SHA1 hashes.


tests/router_test.cc
```diff
@@ -248,6 +248,17 @@ TEST(router_test, test_route_head_request)
...
 
+TEST(router_test, test_remove_not_existing)
+{
+    SegmentTreeNode routes;
+
+    ASSERT_THROW(routes.removeRoute("/v1/hello"), std::runtime_error);
+    ASSERT_THROW(routes.removeRoute("/v1/hello/:name/"), std::runtime_error);
+    ASSERT_THROW(routes.removeRoute("/get/:key?/bar"), std::runtime_error);
+    ASSERT_THROW(routes.removeRoute("/say/*/to/*"), std::runtime_error);
+    ASSERT_THROW(routes.removeRoute("*/api"), std::runtime_error);
```
Tachi107, line 259: I'd prefer this to check for the exact exception that's expected to be thrown so that it is more obvious what this is testing, using `EXPECT_THAT` and `testing::ThrowsMessage`, like this (you also have to include the `gmock/gmock-matchers.h` header):

```suggestion
    using testing::ThrowsMessage;

    ASSERT_THAT(
        [&] { routes.removeRoute("/v1/hello"); },
        ThrowsMessage<std::runtime_error>("Requested route does not exist."));
    ASSERT_THAT(
        [&] { routes.removeRoute("/v1/hello/:name/"); },
        ThrowsMessage<std::runtime_error>("Requested route does not exist."));
    ASSERT_THAT(
        [&] { routes.removeRoute("/get/:key?/bar"); },
        ThrowsMessage<std::runtime_error>("Requested route does not exist."));
    ASSERT_THAT(
        [&] { routes.removeRoute("/say/*/to/*"); },
        ThrowsMessage<std::runtime_error>("Requested route does not exist."));
    ASSERT_THAT(
        [&] { routes.removeRoute("*/api"); },
        ThrowsMessage<std::runtime_error>("Requested route does not exist."));
```

Tachi107 (changes requested):
Thanks for the fix! Just one minor request regarding the newly added test. Otherwise, this looks good to me :)

Tachi107:
It seems that this patch is causing test failures on RHEL 9 with the `router_test.test_remove_not_existing` test.

tyler92:
@Tachi107 Thanks for the suggestion with `tetsing::ThrowsMessage`. I noticed that @kiplingw committed it, and I committed include + fix for the error message. Sometimes we throw `Requested does not exist`, and sometimes `Requested route does not exist`. I made this error the same. Please let me know if you have a better suggestion here.

tyler92:
Sorry, accidentally pushed my local changes for fuzzing, reverted them.

Tachi107:
> Sometimes we throw `Requested does not exist`, and sometimes `Requested route does not exist`. I made this error the same. Please let me know if you have a better suggestion here.

Yeah making the error message the same is definitely a good choice! Still, the new test is consistently failing on RHEL 9, do you think you know why this is the case?

tyler92:
There was an issue with `fragment[0]` expression in `router.cc`. Formally it's not a memory issue because it's `std::string_view` and `fragment[0]` always points (in our case) to a valid memory region. But for the RHEL environment there was an assert in the standard library that failed the test. I've added a check

Tachi107:
The fix looks good to me too. Merging!

PR #1211, opened 2024-04-27
ci(linux): only run Coverage step on Debian Stable

Tachi107:
This is a temporary workaround to the Coverage failures on Debian Testing.

### DIFF
--- a/.github/workflows/linux.yaml
+++ b/.github/workflows/linux.yaml
@@ -92,7 +92,7 @@ jobs:
       run: meson test -C build --verbose
 
     - name: Coverage
-      if: ${{ !contains(matrix.os, 'redhat') }}
+      if: ${{ matrix.os == 'debian:stable' }}
       run: |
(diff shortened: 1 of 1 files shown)


kiplingw:
@Tachi107, `http_server_test.client_multiple_requests_disconnection_on_timeout_from_single_threaded_server` is bailing in the `linux / linux (debian:testing, clang, address, false) (pull_request)` runner. Was that to be expected?

Tachi107:
On Sat Apr 27, 2024 at 7:46 PM CEST, Kip wrote:
> @Tachi107, `http_server_test.client_multiple_requests_disconnection_on_timeout_from_single_threaded_server` is bailing in the `linux / linux (debian:testing, clang, address, false) (pull_request)` runner. Was that to be expected?

No, it's not, but it is unrelated to this patch. Looking at the logs it
seems a bug in the http client code?

Anyway, this shouldn't be a blocker for this patch in particular.

PR #1216, opened 2024-06-07
Support for the Last-Modified header.

yvesdm3000:
Support for the "Last-Modified" header.

To note is that this header is required to be in GMT timezone. The "date" library used by Pistache doesn't support timezones so the stdlib version of gmtime() on a time_t is used to convert the time to GMT. Without this conversion it is difficult to have this value in the older GMT since most users work with the identical UTC tz but that would result in an incorrect header string ("UTC" instead of "GMT".

Complete with a unit test.


Tachi107 (approved):
Looks good to me to! Merging.

Tachi107:
Thanks for the patch :D

PR #1218, opened 2024-07-16
Only install githooks for local checkouts

### LINKED ISSUE #1126: build should not silently override git hooks configuration
meson.build has the following logic:
```
if not meson.is_subproject()
	git = find_program('git', required: false)
	if git.found()
		run_command(git, 'config', '--local', 'core.hooksPath', meson.source_root()/'.hooks', check: false)
	endif
endif
```
Which isn't great when building pistache not from a git clone but inside another git checkout as it will silently override the git hooks configuration of the parent checkout. An example of this is source based distribution build systems like Buildroot (http://buildroot.org) which has a pistache package.

Maybe check for a .git directory or explicitly pass --git-dir=.git to ensure that only the local git checkout is modified?

vfazio:
Some tools, such as Buildroot, will use packed tarballs of cloned git repositories and trigger builds for the sources.

Currently, pistache will install githooks via `git config --local` for any build, even if it's not done from a git checkout. Since the source/build directory is not an actual git checkout, the command traverses up the directory hierarchy and will corrupt the git config for the Buildroot checkout.

```
Buildroot/
  .git/config  # <- this gets corrupted
  output/
    build/
      unpacked-pistache-@-commit-hash-without-.git-directory/
```

Now, we check that the source root directory has a `.git` directory before attempting to install githooks.

Fixes #1126

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -17,6 +17,8 @@ project(
 	meson_version: '>=0.53.2'
 )
 
+fs = import('fs')
+
 #macOS host_machine.system() is 'darwin'
 
 if host_machine.system() != 'linux' and host_machine.system() != 'darwin'
@@ -137,7 +139,7 @@ endif
 
 version_array = []
 if meson.version().version_compare('>=0.57.0')
-	version_array = import('fs').read('version.txt').strip().split('.')
+	version_array = fs.read('version.txt').strip().split('.')
 else
 	# Ugly workaround for reading a file
 	version_array = run_command(
@@ -178,7 +180,7 @@ endif
 
 if not meson.is_subproject()
 	git = find_program('git', required: false)
-	if git.found()
+	if git.found() and fs.is_dir(meson.source_root()/'.git')
 		run_command(git, 'config', '--local', 'core.hooksPath', meson.source_root()/'.hooks', check: false)
 	endif
 endif
--- a/meson.build
+++ b/meson.build
@@ -14,7 +14,7 @@ project(
 		'b_lto=false',
 		'warning_level=3'
 	],
-	meson_version: '>=0.50.0'
+	meson_version: '>=0.53.2'
 )
 
 #macOS host_machine.system() is 'darwin'


vfazio:
If `version.txt` needs a bump, lmk and I can do that.

kiplingw (approved):

kiplingw:
Thanks @vfazio. It looks fine to me, albeit githooks isn't an area of expertise for me. Since you aren't making any source level changes you probably don't need to bump `version.txt`.

I'll let @Tachi107 have a quick look and then we can merge if CI (still running) doesn't complain.

vfazio:
The commitlint complaint is a little odd because it looks like its complaining about a commit prior to the ones in this PR?

kiplingw:
It's alright. Don't worry about that one. The rest of the CI seems fine. Let's wait for @Tachi107 to have a final look over.

Tachi107:
Hi all!

Thanks for the patch! That was an oversight on my part, sorry :)

Is there a reason you require Meson 0.53.2 specifically? If not, just 
use 0.53.0. Also, and this is more important, you should use 
meson.current_source_dir() instead of meson.source_root(), since the 
latter would refer to the root project even if Pistache is used as 
a subproject.

With this minor fixes, this looks good. Thanks again!

vfazio:
> Hi all! Thanks for the patch! That was an oversight on my part, sorry :) Is there a reason you require Meson 0.53.2 specifically? If not, just use 0.53.0. Also, and this is more important, you should use meson.current_source_dir() instead of meson.source_root(), since the latter would refer to the root project even if Pistache is used as a subproject. With this minor fixes, this looks good. Thanks again!

When i was doing naive research, there looked to be issues with 0.53.0 and i didn't really want to risk any issues with the bump. 0.53.2 is also rhe version used by Ubuntu 20.04 so it seemed to naturally fit. The Debian control file has already been merged to use 0.53.2 i believe? 

I can change the meson directory function, but that function was already being used in this area and there is a test to make sure pistache is not a subproject before using source_root

Lmk if you still want me to make these changes and spin up another PR to fix thr control file

Tachi107:
On Wed Jul 17, 2024 at 12:34 PM CEST, Vincent Fazio wrote:
> When i was doing naive research, there looked to be issues with 0.53.0 and i didn't really want to risk any issues with the bump. 0.53.2 is also rhe version used by Ubuntu 20.04 so it seemed to naturally fit. The Debian control file has already been merged to use 0.53.2 i believe? 

Not a big deal, 0.53.2 is fine too. I just find it a bit odd :)

> I can change the meson directory function, but that function was already being used in this area and there is a test to make sure pistache is not a subproject before using source_root

Oh right, I forgot about the is_subproject() check. Looks good then!

vfazio:
@Tachi107 I'll be happy to change it to 0.53.0 if thats your preference, just lmk. that is technically the minimum version for the `fs` module and i didn't test building on it to see if pistache was affected by the linker issues

kiplingw (approved):

PR #1224, opened 2024-08-03
build: fix can't find rapidjson/prettywriter.h

### LINKED ISSUE #1223: failed to build example
I use meson and setup with below option:
```shell
meson setup build \
    -DPISTACHE_USE_RAPIDJSON=true   \
    -DPISTACHE_BUILD_EXAMPLES=true  \
    -DPISTACHE_BUILD_DOCS=true          \
    -DPISTACHE_USE_CONTENT_ENCODING_DEFLATE=true  \
    --prefix=`pwd`/install
```
but after I execute `meson compile -C build`. I got:
```
[20/42] Compiling C++ object examples/runrest_description.p/rest_description.cc.o
FAILED: examples/runrest_description.p/rest_description.cc.o 
c++ -Iexamples/runrest_description.p -Iexamples -I../examples -Iinclude -I../include -fdiagnostics-color=always -DNDEBUG -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Wnon-virtual-dtor -Wextra -Wpedantic -std=c++17 -O3 -Wconversion -Wno-sign-conversion -Wno-missing-field-initializers -pthread -DPISTACHE_USE_CONTENT_ENCODING_DEFLATE -MD -MQ examples/runrest_description.p/rest_description.cc.o -MF examples/runrest_description.p/rest_description.cc.o.d -o examples/runrest_description.p/rest_description.cc.o -c ../examples/rest_description.cc
In file included from ../examples/rest_description.cc:17:
../include/pistache/serializer/rapidjson.h:15:10: fatal error: rapidjson/prettywriter.h: No such file or directory
   15 | #include <rapidjson/prettywriter.h>
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
[33/42] Compiling C++ object src/libpistache.so.0.3.1.p/server_router.cc.o
ninja: build stopped: subcommand failed.
```

From compile command, I can't find corresponding `-I` option. but I checked meson configure and do not find any problem.

Crazyokd:
fix #1223 

I found the `rapidjson_dep` only used in example. And the `meson` seems to discard unused parameter information.

So add `rapidjson_dep` to `examples/meson.build` directly maybe a better choice?


kiplingw:
If it needs to be added as a build dependency then you also need to add it to `debian/control`. However, if I recall correctly, there was a good reason @Tachi107 had raised ages ago why it was never an explicit build dependency. But I can't remember the reason.

Crazyokd:
now if I need pass tests, I should modify the workflows, and add `-DPISTACHE_USE_RAPIDJSON=true` in the stage of configuring meson.

before that expect your better reason.

Tachi107:
Hi @Crazyokd, I've pushed #1225 which should fix the issue you've encountered. Also, note that RapidJSON is not used only in the examples, but also in `include/pistache/serializer/rapidjson.h`. This is the reason why this patch broke CI.

@kiplingw, rapidjson-dev is already listed in debian/control :)

Crazyokd:
@Tachi107 Yeah, that's a better fix.

PR #1226, opened 2024-08-06
C++20 compatibility

radistmorse:
While the project is aiming at C++17, it should be possible to use it with newer versions of the standard. There are two problems preventing it.

First is the hinnant-date library. It was included in the C++20 standard, but with quirks, so using it as-is is problematic. On some compilers it just fails. This PR introduces the wrapper, which replaces the lib with the standard found in `std::chrono` if the language version is high enough.

Second is the implicit lambda capture of `this`. It is deprecated since C++20, you are supposed to use `[=, this]` instead. The problem is, this syntax is only introduced in C++20, and it doesn't work in the previous versions. The additional problem is that in pistache the `[=]` capture is used everywhere, even when it doesn't make any sense. For example in the [Transport::asyncWrite](https://github.com/pistacheio/pistache/blob/master/include/pistache/transport.h#L65) function we capture `buffer` by value (copy the content once), and then pass it to the `BufferHolder` constructor which copies it again. I tried to understand the logic behind `BufferHolder` but I couldn't. Why does `BufferHolder::detach` copies the content of the buffer before passing it to the constructor that copies it again? And same with the `Connection`, why do you use refs in a lambda that captures everything by-value? Argh! My head hurts! But anyway, this is outside the scope of this PR. I replaced the implicit captures with explicit ones in places where it caused problems and tried to keep everything exactly as-is. Fixing the memory management should be a task for somebody more qualified than me.


meson.build
```diff
@@ -36,25 +36,42 @@ if get_option('b_coverage')
...
 ]
 
+#Check if compiler supports C++20 date
+
+cxx_chrono_date_check_code = '''
+#include <sstream>
+#include <chrono>
+int main() {
+    std::stringstream ss;
+	ss << std::format("%FT%T%z", std::chrono::system_clock::now());
```
Tachi107, line 50: I would avoid using `std::format`, since it isn't well supported everywhere and we're just checking for C++20 chrono anyway. Also, the code isn't ran, so you can simply call the functions you want to check for, without actually writing something that "makes sense".

radistmorse: If we want to replace `date::to_stream` from hinnant-date we kinda have no other option. This is the way to turn dates into strings in C++20, no way around it. The `fmt` lib is also available as a sort of preview of C++20 `std::format` like hinnant-date is a preview of the date changes. So a fallback could be introduced, but it would be a rare cornercase for something that is already a rare cornercase (building pistache for C++20).

Tachi107:
Il giorno mar 6 ago 2024 alle 11:13:16 -07:00:00, George Sedov 
***@***.***> ha scritto:
> If we want to replace date::to_stream from hinnant-date we kinda have 
> no other option

Makes sense, I wasn't thinking about the new date::to_stream function 
you've introduced. It's OK to keep this as is then :)

src/client/client.cc
```diff
@@ -186,7 +186,7 @@ namespace Pistache::Http::Experimental
         void registerPoller(Polling::Epoll& poller) override;
         void unregisterPoller(Polling::Epoll& poller) override;
 
-        Async::Promise<void> asyncConnect(std::shared_ptr<Connection> connection,
+        Async::Promise<void> asyncConnect(const std::shared_ptr<Connection>& connection,
```
Tachi107, line 189: Is this related to the modified lambda captures? If not, please either split this in a separate commit explaining why this is needed or drop this change

radistmorse: Hm, no, it's just an error. When I zoned out for a little and just started pressing "yes" to the CLion suggestions to fix things.

Finkman: It is not an error. It is only chasing the reference counter of `std::shared_ptr` more than needed. Best would be to use rvalue-ref or to move it into the `ConnectionEntry`.

kiplingw:
@radistmorse, can you please rebase in the mean time? Someone will eventually get to reviewing your PR.

radistmorse:
Done.

I also changed the way the feature presence is tested. The general `__cplusplus` can cause some problems since the compilers adopt the features in different pace. The feature tests should be better for such cases.

dgreatwood:
I had a quick scan, no objections from what I could see. 

Though they look fine to me, it would be good if @Tachi107 could look over the date and meson.build changes.

JFYI, in light of the syntactic tweaks, I cloned the branch to a OpenBSD machine. It compiles and runs fine there. 👍

The branch will need to be brought up to date with Master again, with a version.txt update. I think that latest change to Master was just the removal of the macOS-12 workflow, now GitHub has removed support for that version of macOS. So should be an easy branch update.

Thanks much!

PR #1235, opened 2024-08-28
Add Accept header serialization

EmilienKia:
Hello there,

Here is a small naive implementation of serialization (write) of Accept header.
It can be useful to use this header object at client side.

Please not that I have split in multiple commits, including the modification of the version.txt and some small modifications of a test file according to clang-format recommendation, to not block commits, even if it is not related to my modifications.

Bests.


Tachi107 (approved):
Hi Emilien, thanks for the patch! This looks good to me.

kiplingw:
Thanks @EmilienKia!

Issue #1200, opened 2024-04-07
[FEATURE REQUEST] Add support for zstd compression


kiplingw:
We currently have `br` (brotli) and `deflate` compression codecs among our list of  [Content-Encoder](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding#directives) support. [Zstd](https://en.wikipedia.org/wiki/Zstd) is the new kid on the block. It is a fast lossless compression codec created by Facebook. We don't currently have support for it.

To implement support should be fairly straightforward in recycling the now already familiar pattern for the two other codecs in Pistache. The [reference](https://raw.githack.com/facebook/zstd/release/doc/zstd_manual.html) C implementation is available on Debian based systems as `libzstd-dev`.

A PR needs to update the build environment, ensuring building with support for it is optional; the `d/control` and `d/rules` would also need to modified with the new build dependency, our `README` to mention it, and a unit test added.

EdgarModesto23:
@kiplingw Hi! Is this currently being worked on? I'd love to help with it.

kiplingw:
Hey @EdgarModesto23. Thanks for showing initiative.

It is not currently being worked on, but we would certainly welcome a PR. You could take a look at #1177 and #1178 for an example of how we added Brotli support, informed the build environment of the required dependencies, and added a unit test to make sure it all works.

EdgarModesto23:
@kiplingw Awesome!, I'll work on a PR.

PR #1247, opened 2024-10-06
Added support for zstd compression

EdgarModesto23:
Following #1200

This patch adds support for zstd compression and it updates build scripts and the main README to include the new -DPISTACHE_USE_CONTENT_ENCODING_ZSTD flag. I mostly followed the style suggested by @kiplingw on #1177 and #1178.

It uses version 1.5.6 of the [C implementation of zstd library](https://github.com/facebook/zstd).


src/common/http.cc
```diff
@@ -922,6 +923,33 @@ namespace Pistache::Http
...
 
+#ifdef PISTACHE_USE_CONTENT_ENCODING_ZSTD
+
+        case Http::Header::Encoding::Zstd: {
+            // Get max compressed size
+            size_t estimated_size = ZSTD_compressBound(size);
+            // Allocate a smart buffer to contain compressed data...
+            std::unique_ptr compressedData = std::make_unique<std::byte[]>(estimated_size);
+
+            // Compress data using compresion_level = 5: https://facebook.github.io/zstd/zstd_manual.html#Chapter5
```
kiplingw, line 934: We currently don't have a public API for the user to specify the compression level (each compressor can use a different range). Until we figure out an elegant way to do that, it might be best to set the compression level consistent with whatever we're using on the others (I think max).

version.txt
```diff
@@ -1 +1 @@
-0.4.7.20240930
+0.4.7.20241005
```
kiplingw, line 1: Since you are adding to an interface, e.g. the Zstd enumerated constant, you should bump the patch version.

tests/meson.build
```diff
@@ -77,7 +81,8 @@ foreach test_name : pistache_test_files
 				gmock_dep,
 				curl_dep,
 				cpp_httplib_dep,
-				brotli_dep
+				brotli_dep,
+        zstd_dep
```
kiplingw, line 85: Remember that `libzstd-dev` also needs to be added to `Build-Depends` in `debian/control`. That's in a separate branch and so you'll need a separate PR for that.

src/common/http.cc
dgreatwood: I believe max is _really_ max on zstd. We could also set compression level dynamically, or partially dynamically, taking into account available CPU bandwidth, available network bandwidth, number of cores (zstd is good with multithreading) etc.

But maybe a middling value is wisest until we do a well-thought-through API/algorithm for compression level?

tests/meson.build
dgreatwood: Hi @EdgarModesto23 -

I see a number of the github workflows (https://github.com/pistacheio/pistache/actions) are failing, e.g. with:
`tests/meson.build:71:1: ERROR: Unknown variable "zstd_dep".
`
Some of that may be related to the need for an update to debian/control. Certainly the way brotli is dealt with should be a good model.

The workflows _linuxflibev_, _linux_ and _macOS libevent_ should all pass (occasionally the debian:testing tests under linux workflow will fail if debian:testing is in a funny state, but in fact it seems OK at present). The _conventional-commits_ and _abidiff_ are less important (just make sure you bump version number if _abidiff_ changes).

dgreatwood (commented):
Regarding http_server_test.cpp -
There are a lot of formatting changes that amount to indenting code that lies within encapsulating (scope limiting) curly braces, like:
```
{ // encapsulate
    // Indent this line by 4 more spaces
    // Etc.
}
```
The extra indentation that has been added is 100% logical, however it was not added before so that the encapsulating braces can be added or removed without causing a "diff" for each of the enclosed lines.

Would it be practical to revert the formatting changes from http_server_test.cpp, leaving only the substantive changes to be reviewed?

src/common/http.cc
EdgarModesto23: @dgreatwood @kiplingw I agree. Using a middle value makes the most sense to me. At least until we find a better solution as both of you guys mentioned.

kiplingw:
@Tachi107, your feedback is appreciated.

Tachi107 (changes requested):
I've left some feedback in a few different comments, check it out!

Also, the [1.5.6 release of zstd](https://github.com/facebook/zstd/releases/tag/v1.5.6) seems to have specific improvements to usage in web contexts. You might want to see what changed and evaluate if tweaking your implementation is worthwhile.

include/pistache/http.h
```diff
@@ -530,6 +534,14 @@ namespace Pistache
                 contentEncodingBrotliLevel_ = _contentEncodingBrotliLevel;
             }
 #endif
+#ifdef PISTACHE_USE_CONTENT_ENCODING_ZSTD
+
+            void setCompressionZstdLevel(const int _contentEncodingZstdLevel)
+            {
+                contentEncodingZstdLevel = _contentEncodingZstdLevel;
+            }
```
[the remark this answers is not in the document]
EdgarModesto23: Thank you for the comment!. I just fixed this.

include/pistache/http.h
```diff
@@ -562,6 +574,10 @@ namespace Pistache
             int contentEncodingBrotliLevel_ = BROTLI_DEFAULT_QUALITY;
 #endif
 
+#ifdef PISTACHE_USE_CONTENT_ENCODING_ZSTD
+            int contentEncodingZstdLevel = ZSTD_lazy2;
```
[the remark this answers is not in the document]
EdgarModesto23: I'm sorry, the official documentation confused me a little bit. This was intended to be a default int value for the compression level. I changed it to ZSTD_defaultCLevel().

tests/http_server_test.cc
```diff
@@ -1297,6 +1301,12 @@ struct ContentEncodingHandler : public Http::Handler
         switch (encoding)
         {
 
+#ifdef PISTACHE_USE_CONTENT_ENCODING_ZSTD
+        case Http::Header::Encoding::Zstd:
+            writer.setCompressionZstdLevel(ZSTD_btultra2);
```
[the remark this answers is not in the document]
EdgarModesto23: Same as the ZSTD_lazy2 issue above. I got a bit confused with the documentation. I changed it to be ZSTD_maxCLevel() as you suggested.

kiplingw (approved):

kiplingw:
@Tachi107 and @dgreatwood, looks fine to me.

kiplingw:
@EdgarModesto23, please review our PPA's build log. Your PR seems to break on all architectures on jammy. Here's is the [log](https://launchpadlibrarian.net/758241566/buildlog_ubuntu-jammy-amd64.pistache_0.4.17.20241110+git20241110.2e98845~ubuntu22.04.1_BUILDING.txt.gz) for amd64.

Can you look into your build dependencies? If we can't satisfy Jammy, we will have to either drop it, or at least find a way to disable support for it in Jammy in `debian/rules`.

EdgarModesto23:
@kiplingw I'll fix it right away.

EdgarModesto23:
@kiplingw I made a PR that should fix the problem. I'll stay around in case anything comes up. Thank you for reaching out!

dgreatwood:
Hi @EdgarModesto23 -

May I raise another question / concern, this time about the test code?

In your http_server_test.cc/server_with_content_encoding_zstd, you have:
```
const auto compressionStatus = ZSTD_getFrameContentSize(newlyDecompressedData.data(), newlyDecompressedData.size());

const auto decompressed_size = ZSTD_decompress((void*)newlyDecompressedData.data(), compressionStatus, newlyCompressedResponse.data(), newlyCompressedResponse.size());
```
Here are my concerns:

1. Per spec, ZSTD_getFrameContentSize's first parameter points "to the start of a ZSTD encoded frame". However,  newlyDecompressedData doesn't contain a ZSTD frame. It doesn't actually contain anything yet, since we have yet to decompress the data when ZSTD_getFrameContentSize is called.
2. If ZSTD_getFrameContentSize fails, compressionStatus may be an error code. But compressionStatus is not checked for an error.
3. Given (1) above, ZSTD_getFrameContentSize actually does fail, and compressionStatus actually is an error code.
4. The only reason the subsequent call to ZSTD_decompress works is that the error code is a very large number (as an unsigned), and so it appears we're passing a sufficiently large - in fact very large - buffer to ZSTD_decompress.

In short, it seems that ZSTD_getFrameContentSize should act on newlyCompressedResponse, and then ZSTD_getFrameContentSize's return code should be checked. I included this code:
```
// Decompress...
auto decompressedSzFromFrame = ZSTD_getFrameContentSize(newlyCompressedResponse.data(), newlyCompressedResponse.size());
if (ZSTD_isError(decompressedSzFromFrame))
{
    LOGGER("test", "getFrameContentSize result: " <<
           ((decompressedSzFromFrame == ZSTD_CONTENTSIZE_UNKNOWN) ?
            "Content Size Unknown" :
            (decompressedSzFromFrame == ZSTD_CONTENTSIZE_ERROR) ?
            "Content Size Error" : "Other"));
    decompressedSzFromFrame = newlyDecompressedData.size();
}
const auto decompressed_size = ZSTD_decompress(reinterpret_cast<void*>(newlyDecompressedData.data()), decompressedSzFromFrame, newlyCompressedResponse.data(), newlyCompressedResponse.size());
```
(cf. https://github.com/dgreatwood/pistachefork/blob/windows/tests/http_server_test.cc)

Could you please review the comments above and likewise my proposed code?

Thanks much!

PR #1250, opened 2024-10-12
Add zstd compression support to package build dependencies

EdgarModesto23:
Following #1247 and #1200

Adds libzstd-dev to build dependencies

### DIFF
--- a/control
+++ b/control
@@ -11,6 +11,7 @@ Build-Depends: debhelper-compat (= 13), meson (>= 0.53.2)
 Build-Depends-Arch: libbrotli-dev,
                     libcurl4-openssl-dev,
                     libevent-dev,
+                    libzstd-dev,
                     libgmock-dev (>= 1.11.0) <!nocheck> | googletest (>= 1.11.0) <!nocheck>,
                     libssl-dev,
                     netbase <!nocheck>,
@@ -40,6 +41,7 @@ Section: libdevel
 Architecture: any
 Depends: libpistache0 (= ${binary:Version}),
          libbrotli-dev,
+         libzstd-dev,
          libssl-dev,
          rapidjson-dev,
          zlib1g-dev | libz-dev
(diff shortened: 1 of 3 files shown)


control
```diff
@@ -11,6 +11,7 @@ Build-Depends: debhelper-compat (= 13), meson (>= 0.53.2)
 Build-Depends-Arch: libbrotli-dev,
                     libcurl4-openssl-dev,
                     libevent-dev,
+                    libzstd-dev,
```
kiplingw, line 14: @Tachi107, does he need it in the `Build-Depends-Arch` stanza as well?

rules
```diff
@@ -38,6 +38,7 @@ override_dh_auto_configure:
 		-DPISTACHE_USE_SSL=true \
 		-DPISTACHE_USE_CONTENT_ENCODING_BROTLI=true \
 		-DPISTACHE_USE_CONTENT_ENCODING_DEFLATE=true
```
kiplingw, line 40: Missing trailing \

kiplingw (changes requested):

rules
EdgarModesto23: @kiplingw My bad haha. Fixing now

control
Tachi107: Sure. Adding both in `Build-Depends` and `Depends` means "libzstd-dev is needed both for building and consuming the libpistache package"

kiplingw: No, sorry, I meant specifically `Build-Depends-Arch` as opposed to just `Build-Depends`.

Tachi107: Build dependencies listed in the `Build-Depends-Arch` filed need to be satisfied when building architecture-dependent packages (i.e. packages marked with `Architecture: any`), while deps listed in `Build-Depends` need to also be present when building any kind of package, or when just running the `clean` target.

So `Build-Depends-Arch` is more correct here, since pistache doesn't build arch-independent packages and `libstd-dev` isn't needed to run the `clean` target.

This comes from [Debian Policy section 7.7](file:///usr/share/doc/debian-policy/policy.html/ch-relationships.html#relationships-between-source-and-binary-packages-build-depends-build-depends-indep-build-depends-arch-build-conflicts-build-conflicts-indep-build-conflicts-arch) (need to install the `debian-policy` package to open the link)

PR #1251, opened 2024-10-13
fix: buffer overflow with strtol

tyler92:
Closes https://github.com/pistacheio/pistache/issues/1193

The buffer overflow is fixed and checked with the fuzzing test but I would still appreciate feedback about the correctness of the parsing part. The main differences between `strtol` and `from_chars` as it was said [here](https://github.com/pistacheio/pistache/pull/1201#issuecomment-2053803517):

1/ from_chars does not ignore leading white space, strtol does.
2/ from_chars does not recognize a leading "+" sign (or signs), whereas strtol treats a leading plus (or pluses) as indicating that the value is positive, which is the default in any case.

I didn't add ignoring leading spaces because the code does it before `from_chars`. Also, I didn't find in many RFCs that leading `+` is allowed.


Tachi107 (approved):
Looks good to me! This will also close #1208 when merged

kiplingw:
@tyler92, I'm almost ready to merge. But can you please bump the patch version in `version.txt` because you've changed the internal implementation?

tyler92:
> can you please bump the patch version in version.txt because you've changed the internal implementation?

Done, 0.4.11.20241018

PR #1252, opened 2024-10-13
Avoid using compiler.has_function('dladdr')

dgreatwood:
It would be nice to use compiler.has_function('dladdr') (as per prior attempt, commit e9cb3a5d27e1e995b55fd309a09c63cdec69d31d) to test if we need to add libdl, but unfortunately that approach seems to break certain Red Hat builds, specifically the Red Hat builds that use ubi-8. In such cases, it appears meson creates a test file that includes the comment:

> With some toolchains ... the compiler provides various builtins which are not really implemented... [If] the user provides a header, including the header didn't lead to the function being defined, and the function we are checking isn't a builtin itself, we assume the builtin is not functional and error out

To avoid generating such an error, we take the simpler approach of trying to add libdl but making it optional (i.e. not required).


kiplingw (approved):

meson.build
```diff
@@ -154,14 +154,23 @@ endif
...
+#   which are not really implemented... [If] the user provides a
+#   header, including the header didn't lead to the function being
+#   defined, and the function we are checking isn't a builtin itself,
+#   we assume the builtin is not functional and error out
+# To avoid generating such an error, we take the simpler approach of
+# trying to add libdl but making it optional (i.e. not required).
+if meson.version().version_compare('>=0.62.0')
+  deps_libpistache += dependency('dl', required: false)
+else
+  deps_libpistache += compiler.find_library('dl', required: false)
```
eli-schwartz, line 173: `dependency('dl')` explicitly does `#include <dlfcn.h>` here.

What was the problem with ubi-8? Did they define a broken builtin that got used by `has_function('dladdr')` without a header match? Are you expecting for `__builtin_dladdr` or `dladdr` to exist??

dgreatwood: > What was the problem with ubi-8? 

It was a while ago, but, from memory, meson throws an error with ubi-8 if meson.build has a line like:
`... compiler.has_function('dladdr')`


> Did they [ubi-8] define a broken builtin that got used by has_function('dladdr') without a header match?

Per the comment added to meson.build, it appears that meson detects/thinks that "the builtin is not functional". I did not attempt to verify whether the builtin really is "not functional" or if something else is going on. 


> Are you expecting for __builtin_dladdr or dladdr to exist??

The actual function invoked in the Pistache source is `dladdr(...)`; we also use the struct `DL_info`; and the header included in our source is _dlfcn.h_, which of course is the standard header for `dladdr(...)` per POSIX and Linux documentation. Again, I haven't dug into meson's own methodology here. Also, as I think you know but for the record, often the dl library is not needed in fact since dladdr is commonly provided in libc.

eli-schwartz: I feel like you're not really getting what I'm trying to say here at all. :(

Your previous check was definitely suboptimal -- you used this:

```meson
cc.has_function('dladdr')
```

which doesn't say anything about [`prefix: '#include <dlfcn.h>'`](https://mesonbuild.com/Reference-manual_returned_compiler.html#has_function_prefix) and thus meson had to jump through a few more logical hoops to check various different possibilities and isn't, in fact, checking at all whether it is the POSIX-defined dlfcn.h interface. It's probably inconceivable that there is a dladdr defined anywhere else, but it turns out it does matter for the builtins check...

`dependency('dl')` does handle that though. :) 



> It was a while ago, but, from memory, meson throws an error with ubi-8 if meson.build has a line like:
> `... compiler.has_function('dladdr')`

Meson certainly should not throw an *error* unless something very wrong happened. It should simply return a boolean false as the value of `has_dladdr_func`. I think it's important to actually define what's really happening using precise terms that help to better understand what is happening.

You don't need to "avoid generating such an error" if the error is an internal compiler diagnostic whose only purpose is as part of a *series* of checks probing system state before returning a well-formed meson not-error. You just need internal compiler error diagnostics to reliably occur when it is accurate for them to occur, and not occur when they shouldn't occur... and you haven't defined which of those is the case.



> Per the comment added to meson.build, it appears that meson detects/thinks that "the builtin is not functional". I did not attempt to verify whether the builtin really is "not functional" or if something else is going on.

If dladdr exists as a regular symbol meson will never proceed so far as to check if it is a builtin. So it's quite relevant to ask whether it's a builtin or not.

But much, much worse, is the fact that the code you cite in your comment ***cannot*** run -- it is part of a C preprocessor statement defined to `#if ! 1` -- in pistache, since pistache doesn't use [`prefix: '#include <dlfcn.h>'`](https://mesonbuild.com/Reference-manual_returned_compiler.html#has_function_prefix)

So (speaking as a Meson maintainer) it seems evident to me that this PR is based on a mistaken understanding of what Meson is doing. I'm interested to hear more, because it is very much not the intention of Meson that `cc.has_function('dladdr')` should return false when there is in fact a dladdr symbol in libc.

> Again, I haven't dug into meson's own methodology here. Also, as I think you know but for the record, often the dl library is not needed in fact since dladdr is commonly provided in libc.

Yes, I'm aware. I'm the one who reviewed and merged meson's `dependency('dl')`, which based on my advice first checks to see if `dlopen()` is provided in libc, and then if it is not provided in libc, checks again to see if it is provided in `-ldl`.

You can use `dl_dep.type_name() == 'builtin'` to check whether meson detected it in libc. The only valid type_name() values are "builtin" (libc) and "system" (`-ldl`).

eli-schwartz: Again -- if dladdr *does* exist but `cc.has_function('dladdr')` returns "no, sorry, it doesn't exist" then that feels like meson is doing something wrong and I'd like to understand how it occurs and hopefully fix it.

dgreatwood: Hi Eli -

Thanks for digging in.

> Meson certainly should not throw an error unless something very wrong happened...

Ah, but Meson did generate an error - I mean a build error, not an internal meson error - in this case. The meson.build [code](https://github.com/pistacheio/pistache/pull/1246/files) generating the error:
```
has_dladdr_func = compiler.has_function('dladdr')
if not has_dladdr_func
  libdl_dep += dependency('dl')
  has_dladdr_func = compiler.has_function('dladdr', dependencies: libdl_dep)
  if not has_dladdr_func
    warning('Unable to find dladdr(), even when trying to link to libdl')
  endif
  libpistache_deps += libdl_dep
endif
```

The github workflows results are [here](https://github.com/pistacheio/pistache/actions/runs/11113243656). You can see that the ubi8 builds all failed. Ignore the debian:testing failures, that was an unrelated issue. The actual output log is gone I think (I believe GitHub discards them after 90 days by default) but you can [see](https://github.com/pistacheio/pistache/actions/runs/11113243656/usage) that the ubi-8 builds all failed in less than a minute, whereas the ubi-9 builds succeed in 3-4 minutes. The ubi-8 build failures came from `meson setup` throwing a build error.

I went ahead and recreated a fork branch corresponding to commit bbc70155ee1981081a3687eb28e54daefa32f6c0, which had this issue. You can see the workflow output log [here](https://github.com/dgreatwood/pistachefork/actions/runs/12897967298/job/35964104897).

From my reading, the first `compiler.has_function('dladdr')` works fine, returning false in this case as you suggested. But the second `has_function('dladdr', ...)` leads to the the build error.

Does that help?

eli-schwartz: Thanks, this error message is illuminating.

```
Checking for function "dladdr" : NO 
Run-time dependency dl found: NO (tried pkgconfig and cmake)

meson.build:159:2: ERROR: Dependency "dl" not found, tried pkgconfig and cmake
```
Very suspicious! "dl" is not available via either pkg-config or cmake. The only valid tries are "builtin" and "system".


And also,

```
The Meson build system
Version: 0.61.5
Source dir: /__w/pistachefork/pistachefork
Build dir: /__w/pistachefork/pistachefork/build
Build type: native build
Project name: pistache
Project version: 0.3.0
```

This failed to work because in order to use `dependency('dl')` you must bump your minimum required meson to 0.62.0

> From my reading, the first `compiler.has_function('dladdr')` works fine, returning false (though I wonder if the correct answer would actually be 'true') in this case as you suggested. But the second `has_function('dladdr', ...)` leads to the the build error.

This branch only executes `compiler.has_function` a single time within your workflow results. Within the meson.build code, the following procedural logic is performed:

- `has_dladdr_func = compiler.has_function('dladdr')`
- `libdl_dep += dependency('dl')`
- `has_dladdr_func = compiler.has_function('dladdr', dependencies: libdl_dep)`

At step 1, for ubi-8 systems dladdr is not available in libc (you need `-ldl` for this).

At step 2, the workflow aborted because a required dependency was not found (and Meson was too old).

Step 3 would be the second has_function, but meson did not get that far.

eli-schwartz: I believe the talk about compiler builtins is a red herring.

You require, I guess, that the Meson version shipped as an rpm on ubi-8 is sufficient to build pistache. That means you cannot (yet) use the custom dependency. Instead, you have to manually code it yourself. Here is a meson.build reimplementation of `dependency('dl')`:

```meson
if compiler.has_function('dladdr', prefix: '#include <dlfcn.h>')
    dl_dep = declare_dependency()
else
    dl_dep = compiler.find_library('dl', required: true)
endif
```

Issue #1257, opened 2024-10-24
`std::string Pistache::Http::Header::Authorization::getBearerToken()` . Where art thou?


mowijo:
I feel `std::string Pistache::Http::Header::Authorization::getBearerToken()` is missing while its siblings 

 * `std::string Pistache::Http::Header::Authorization::getBasicUser()`
 * `std::string Pistache::Http::Header::Authorization::getBasicPassword()`

exist.

Is there a design reason for lack of that method or has it just not been implemented yet? It it is the latter, I shall get on with it immediately.

kiplingw:
The latter. I implemented basic authorization a long time ago in #725.

mowijo:
# 🥳

mowijo:
> The latter. I implemented basic authorization a long time ago in #725.

But it wasn't merged. How comes?

dgreatwood:
Hi @mowijo -

Just to confirm what I think you already understood - #725 implemented BasicUser/BasicPassword, and that support was merged.

If you would like to propose a PR related to bearer token use, then that sounds good to me and I'm sure we'd be happy to take a look.

Existing mentions of bearer in the code base you can see here:
https://github.com/search?q=repo%3Apistacheio%2Fpistache%20bearer&type=code

Thanks.

winkler-pixop:
I  got it , thank you, and I am on it!

Thanks, Morten

mowijo:
There is a PR for that now: https://github.com/pistacheio/pistache/pull/1258 

Please let me know what you think about that.

PR #1258, opened 2024-10-27
Feat: Add method to extract HTTP Bearer token from Authorisation header.

mowijo:
This PR Adds a method to extract HTTP Bearer token from the Authorisation header, thus closing #1257.


src/common/http_header.cc
```diff
@@ -450,6 +450,16 @@ namespace Pistache::Http::Header
         return true;
     }
 
+    std::string Authorization::getBearerToken() const
+    {
+        // Verify bearer authorization method was used...
+        if (!hasMethod<Authorization::Method::Bearer>())
+            throw std::runtime_error("Authorization header does not use Bearer method.");
+
+        const std::string token(value_.begin() + std::string("Bearer ").length(), value_.end());
```
kiplingw, line 459: This is a problem. You need to make sure that there was a properly formatted token after `Bearer `. Otherwise you could end up with a scenario where the client sends a malformed request, perhaps intentionally, and it takes down the server.

kiplingw (changes requested):
You also need to bump the patch version in `version.txt` because you are adding new interfaces.

src/common/http_header.cc
mowijo: Ok. I see the need for security, but I don't understand how to check if the token is properly formatted after `Bearer ` since we don't know what sort of token people might want to exchange. All I can think of is, to verify that the value of the header value matches this `^Bearer [!-~]+$`. (Only printable ascii chars except space).
 
Do you think that will suffice? If not, what sort of check did you have in mind?

kiplingw: Well as it is right now, unless I'm reading it wrong, you're not checking for the presence of anything which could lead to a segfault.

winkler-pixop: There is a check that the string starts with `"Bearer "`. That is done by the 

`if (!hasMethod<Authorization::Method::Bearer>())` statement.

1. Would you prefer that I check for that myself in the method body? It is not a problem at all to do so. I have been unable to cause any crashes with any partiality of the header value.
3. Would you like that added as testcases?

kiplingw: Isn't it not checking for anything after "Bearer " though before trying to initialize the std::string's internal buffer?

dgreatwood: I would check for a minimum token size, and maximum, and reject if prospective length falls outside that length.

Minimum - definitely the size needs to be > 0. A common minimum is 16 bytes. To be conservative, I would probably reject any token of less than 4 bytes in length.

Maximum - most web servers won't except more than 8KB of headers, so that forms an effective max length. Again, being conservative, we could reject anything with length > 16KB.

Would it be illegal to have a null as part of the bearer token? If so, I would check for that before forming the std::string and reject any token that contains same.

Should we reject attempts to authenticate with bearer token when SSL is _not_ in use?

Not sure what else might be not allowed, but you get the idea.

Yes, unit tests for any such validity check would be great.

Thanks!

winkler-pixop: > I would check for a minimum token size, and maximum, and reject if prospective length falls outside that length.
> Minimum - definitely the size needs to be > 0. A common minimum is 16 bytes. To be conservative, I would probably reject any token of less than 4 bytes in length.
> Maximum - most web servers won't except more than 8KB of headers, so that forms an effective max length. Again, being conservative, we could reject anything with length > 16KB.

I dont think it shold be up to a framework to enforce a minimum size of a token.  Likewise, no minimum size for the password for Basic is implemented in Pistache and 

Limiting to a specific length for security reasons sounds like a good idea, but would hold true for all headers.
 
> Would it be illegal to have a null as part of the bearer token?

Yes that would indeedbe illegal. According to RFC7230 section 3.2, the content of a header field must be only visible ASCII chars and optional folds, but I don't see how that logic belongs in the Authorization-specialization.

>  If so, I would check for that before forming the std::string and reject any token that contains same.

Indeed, but that holds true for any header and not only the Authorization header. 
 
> Should we reject attempts to authenticate with bearer token when SSL is _not_ in use?

No. That would make it annoyingly difficult to run a Pistache server behind a SSL terminating reverse proxy. 
 
> Not sure what else might be not allowed, but you get the idea.

Honestly, no. I do not get the idea.

It seems like all the issues  above are issues that are general for all http headers. Placing such checks in the Authorization header seems like too low in the abstraction to me. If these checks should be performed (and they should) , it should be at a higher level so the individual header specializations 

> Yes, unit tests for any such validity check would be great.
Will do!

I would like to suggest that

 1. I add additional tests for the token extraction
 2. We create an issue for several of the issues above to be addressed in a central "Header-parsing-place-that-I-dont-know -where-is-or-exists" and solve them for all headers in one implementation
 3. We merge the logic for Bearer extraction in this PR.

dgreatwood: I'm fine with most of this except that I would have your code at least check for a non-zero token size, i.e. reject a zero-sized token, if you're not willing to reject a token of less than 4 bytes as a stricter test.

I agree that pathologically large headers, or headers containing a null, are applicable to all headers and would best be dealt with at a higher layer, just as you said. Would be great if you would create issues for same - and for any other header validity check you think we might include but which is not covered today.

Thanks again!

mowijo: Sure. That sounds good. I can do that.

Work just took up a lot of my spare time the couple of days ;-)

kiplingw: @mowijo, can you update us on your progress?

kiplingw (changes requested):
Please check for zero length token and generate an error in that scenario.

Issue #1128, opened 2023-03-30
Windows Builds


Rxann:
Could builds for windows be released?

kiplingw:
Thanks @Rxann, but this has been asked many times already. We don't have any plans on porting the library to legacy systems.

Rxann:
Thanks

Tachi107:
Hi @Rxann, yeah as Kip said Windows is not currently a focus for the project. You can find more detailed info in issue #6, it's not trivial to port Pistache to non-Linux OSes

Issue #869, opened 2021-03-01
[FEATURE REQUEST] Windows Support


kyro-dev:
Is the pistacheio team planning to add support for windows? I'm developing on windows and I would really like to use your library.

Tachi107:
As mentioned in #6, #538 and #649, it would not be an easy task, especially because the issue is related to architectural decisions. In the meantime you could use WSL, or maybe try out Linux ;) 

Anyway, I don't think that opening another issue will improve the situation, also because #538 is still open; I think you can close this, thanks

kyro-dev:
Ok, thanks anyways.

Issue #649, opened 2020-01-04
Windows support soon?


AdavizeTheFirst:
This is a very great framework I'd like to know whether to expect any support for windows soon. 
The website says "No support for windows yet".

kiplingw:
Hey @AdavizeTheFirst. I can't speak for the other maintainers, but as for myself I don't see any reason for triaging resources on legacy systems. Windows isn't just disappearing on the desktop, but it's already virtually gone in the server world which is where Pistache always had its primary application.

Asimov500:
I am not botherd if there is a windows version but windows isn't disappearing on the desktop anywhere I have been. I am a web developer at work and windows is certainly not disappearing. I also run a windows 2012 server. I have seen no sign of windows disappearing anywhere?? I am also a 3D modeller and it isn't something I would like to do on a linux computer that can't run 3ds Max. Windows is far from being dead.

kiplingw:
If you are making websites then you don't have any need for Pistache. If you are developing a high performance server written in C++ with a REST API, then you have a use case. 

As for CentOS, Pistache should build on any GNU system. The pre-compiled packages are prepared for Ubuntu, but the source is portable across different GNU systems. The packages are just convenience packages to save from having to build and run DEP-8 tests.

Asimov500:
I do web development but I also write C++ code. We have php apis which are  on our centos server. We are currently converting some of our apis to C++ for extra speed. We have got it working in Ubuntu but had a lot of trouble trying to compile it for Centos. I totally understand why you don't wish to support windows, but it is far from being dead.

kiplingw:
@Asimov500, if this is a company you work for, tell them to reach out to me and I can help them port and package it for CentOS.

hydratim:
There is a fundamental network stack reason we don't support windows.
Please search for tickets before raising a new one: https://github.com/oktal/pistache/issues/6

Issue #538, opened 2019-05-15
Do you have any plan to support Windows ?


portsip:
Do you have any plan to support Windows ?

hydratim:
Not at the moment, the network stack doesn't support it (see the answer from @oktal in issue #6) due to architectural decisions.

In theory we could potentially add a compile-time option the replaces the network stack with boost (thus changing from a read-ready to a read-complete model), thus allowing windows support, but it's not likely to happen soon.

The latest Windows 10 will come with a full Linux Kernel in WSL though, so in the mean time you could potentially use that for developing with pistache.

Aidoboy:
Pistache works great in WSL1 on Windows 10, that's how I've used it.

seghcder:
The reason I chose Pistache was to avoid the dependency on Boost, and to have a fully native, fast, standalone ReST server. I am also keen to see a Windows native version of Pistache. 

I'd be willing to sponsor/donate a few hundred US dollars to support this project and particularly this specific goal. Would anyone else be willing to support?

I've raised a new issue re that topic... https://github.com/oktal/pistache/issues/569

Is it reasonably achievable? Or are those architectural decisions too fundamental to Pistache?

seghcder:
FYI regarding Travis - https://blog.travis-ci.com/2018-10-11-windows-early-release and https://docs.travis-ci.com/user/reference/windows/ (in reference question in #569)

mheyman:
https://github.com/piscisaureus/wepoll may provide a good if limited option - a scalable drop-in replacement for `epoll()` on windows. Wepoll doesn't do `EPOLLET` (edge-triggered) mode and only works with sockets.

Tachi107:
I would close this and reopen #6, since it contains more information about the subject and was opened first

Tachi107:
As mentioned in my previous comment, I'm closing this issue in favour of #6. It contains far more information than this one and was opened earlier

PR #1259, opened 2024-11-09
Windows Support

dgreatwood:
Addresses: Issue #6, #649, #869, #538, #1128. 

The changes for Windows support fall into 5 main buckets:
1. Build and environment: _meson.build_, environment configuration (_winscripts/*.ps1_), and github runner (_windows.yaml_).
2. Windows logging support
3. Syntactic and API difference support
4. Event-handling support in Windows.
5. Misc/Other

Taking each in turn:

### Build and environment
The meson.build files are adjusted to link Windows libraries; to compile and install Pistache's Windows logging methods (uses custom_target); and to export symbols from the DLL (including _dump2def_ subproject).

Environment configuration: We presume development will be carried out via a PowerShell terminal (local or remote). The files _winscripts/msvcsetup.ps1_ and _winscripts/gccsetup.ps1_ are provided to configure development with Visual Studio or GCC respectively. The file _Building on Windows.txt_ gives Windows-specific build instructions.

windows.yaml: Leverages _msvcsetup.ps1_ and _gccsetup.ps1_ to set up the build environment and install dependencies. Otherwise works similarly to other Pistache runners (_linux.yaml_ etc.). 

### Windows logging support
We log using the Windows-preferred "EWT" methodology. The _meson.build_ methods provide for the compilation and installation of the necessary logging manifest etc. At the level of the Pistache code, the `PS_LOG_xxxx` methods are used, just as per Linux, macOS etc.

_pist_syslog.cc_: In addition to invoking the Windows logging functions, it also uses the Windows API to monitor the registry property HKCU:\Software\pistacheio\pistache\psLogToStdoutAsWell; by setting this property to 1 (or back to 0), the user can turn on (or off) verbose logging to stdout.

### Syntactic and API difference support
There are differences between Linux and Windows in type and constant definitions, API call definitions, and C++ syntax. This causes a lot of small changes across the Pistache code base.

The header file _winornix.h_ defines pre-processor macros and header-file names. By using these definitions, in most cases in the code base we **avoid** patterns such as:
```
#ifdef _IS_WINDOWS
    doThis();
#else
    doThat();
#endif
```
As an example, there is no `ssize_t `type in Windows. We define `PST_SSIZE_T` in _winornix.h_, which works in both Windows and *nix code, and reduce to `ssize_t` in the *nix case. 

In a number of cases, we have provided wrappers (or occasionally larger implementations) for Windows so that we can have Linux-like function interfaces for use in the broader Pistache code base. These "Linux-like functions in Windows" are usually fronted with a preproccessor macro to abstract the Windows vs. *nix differences. As an example, the macro `PST_SOCK_BIND` is used the same way in Windows and *nix; in Windows it invokes our function `pist_sock_bind` and in the *nix case it reduces to `::bind`. Files that provide these kind of wrapper include:
- pist_clock_gettime
- pist_fcntl
- pist_filefns (read, open)
- pist_ifaddrs
- pist_resource (getrusage)
- pist_sockfns (connect, read, close etc.)
- pist_strerror_r
- ps_basename
- ps_sendfile
- ps_strl (strlcpy / strlcat)
- pst_errno.h (to wrap errno.h)

### Event-handling support in Windows
Most of the differences between event handling in Windows vs. *nix are already abstracted away by the previously-implemented libevent support. The major remaining item is the lack of edge-triggered events in Windows; Windows is level-triggered only. Code in _eventmeth.cc_ emulates edge-triggered behaviour sufficiently for Pistache, essentially by deactivating a activated event, and then, if appropriate, adding it back to the event polling loop ready for the next activation.

### Misc/Other Items of Note
C-style casts. Having introduced additional C-style casts during the Windows-support implementation, we then went through to replace them with C++-style casts. It turned out to be easier to do that replacement for all C-style casts, whether newly introduced or not, so that's what we did.

_client.cc_: Reworked Transport::handleIncoming; its logic around receiving in multiple pieces (cf. use of totalBytes) didn't work before. Windows seems more prone to using multiple receive events to receive a single incoming datagram.

_eventmeth.cc_
- Some rework of how the shutdown of an event-loop works - don't use "event finalize later" when not needed.
- Deal with lack of CLOEXEC in Windows (mostly by ignoring it)

_reactor.cc_
- Additional protection (shutdown_mutex_) during shutdowns. This was an issue exposed in Windows.
- Code for setting thread description. Has to dynamically load the symbol SetThreadDescription from KernelBase.dll in older versions of Windows.

_transport.cc_ - improvements when using TCP_NODELAY, which Windows requires.

_listener.cc_ - use an atomic counter, not the file descriptor number, to generate an index into the array of handlers for a new connection. File descriptor numbers don't have the nice "increment by one each time" pattern in Windows that they do in Linux.

_dump2def_ - subproject utility to assist in generating symbols for DLL export. This is needed becasue, by default, Windows DLLs do not export their symbols, unlike in Linux.

_http_server_test.cc_
- Got rid of class SimpleLogger; improved ScopedLogger. This makes it easier to integrate with Windows logging.
- Deal with the different temporary-file mechanisms in Windows
- Normalize some indentation (been avoiding this because it shows as a lot of changes, but it's time)


kiplingw (approved):

kiplingw (approved):

Tachi107 (commented):
Hi Duncan, thanks for your incredible work. Windows support is something many people wanted and this patch set will make them quite happy. I don't know how many people will actually run Pistache on a production environment with Windows, but still, portability is a worthwhile goal in itself, especially when it comes to supporting non-Unix systems. Unix made lot of things right, but it's not perfect (my bachelor's thesis is partly about this, in fact).

With that being said, I'm not going to review this. There are more then ten thousand changed lines, spanning over 300 commits, and it is hence impossible for anyone to review such work in one's free time. I would have preferred small, organized changes sent over time via multiple pull requests.

Kip, I leave this to you :)

Thanks again for your incredible work!

dgreatwood:
Hi @Tachi107 -

Thank you for your consideration :-)

Yes, indeed, this PR is essentially a large number of tiny changes to address syntactical and API matters, combined with some lumps of PowerShell script to configure the dev environment.

Should you have additional comments related to the changes in future, I will of course be more than happy to take a look.

Meanwhile, all the best!

src/meson.build
```diff
@@ -89,12 +381,42 @@ if meson.version().version_compare('>=0.54.0')
...
+            # to libraries may be harmful since it makes pistachelog
+            # appear on the .pc file's "Libs" line, and we don't
+            # actually want applications to try and link with
+            # pistachelog.
+            # ref: https://mesonbuild.com/Pkgconfig-module.html
+            name: 'Pistache',
+            description: 'An elegant C++ REST framework',
+            url: 'https://pistacheio.github.io/pistache/',
+            version: '@0@-git@1@'.format(version_str, version_git_date),
+            filebase: my_filebase,
```
Tachi107, line 409: Why is filebase set to "pistache" on Windows?

dgreatwood:
> Can I change it back to being "libpistache" everywhere?

I don't see a problem with that, so long as it passes the automated tests. Thanks.

PR #1260, opened 2024-11-10
Fix: For RedHat, Explicitly Install lld as a Dependency

dgreatwood:
RedHat-Ubi8-Clang builds had started to fail because lld (linker) was not being found. Now we explicitly include "lld" on the "microdnf install" line along with $compiler and the other items, and it works again.


Tachi107:
Makes sense. Thanks!

PR #1261, opened 2024-11-12
Fix libzstd-dev on debian build

EdgarModesto23:
From #1247 

This patch fixes issues mentioned by @kiplingw on https://github.com/pistacheio/pistache/pull/1247#issuecomment-2469344358.

The issue was caused because the patch that adds zstd support uses the _ZSTD_defaultCLevel()_ function. which is [only supported by libzstd v1.5.0+](https://raw.githack.com/facebook/zstd/release/doc/zstd_manual.html#Chapter3). The current configuration uses libzstd v1.4.8

This PR fixes that by specifying that pistache depends on libzstd >= 1.5.4.

### DIFF
--- a/control
+++ b/control
@@ -11,7 +11,7 @@ Build-Depends: debhelper-compat (= 13), meson (>= 0.53.2)
 Build-Depends-Arch: libbrotli-dev,
                     libcurl4-openssl-dev,
                     libevent-dev,
-                    libzstd-dev,
+                    libzstd-dev (>= 1.5.4),
                     libgmock-dev (>= 1.11.0) <!nocheck> | googletest (>= 1.11.0) <!nocheck>,
                     libssl-dev,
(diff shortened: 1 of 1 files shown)


kiplingw (changes requested):
Please see my [comments](https://github.com/pistacheio/pistache/pull/1261#issuecomment-2469469522).

EdgarModesto23:
@kiplingw Got it. I think in this case, option 4 is the easiest solution. Since ```ZSTD_defaultCLevel()``` just returns the default for ZSTD_CLEVEL_DEFAULT, which is 3 according to [documentation](https://raw.githack.com/facebook/zstd/release/doc/zstd_manual.html#Chapter3).

kiplingw:
In that case then I'll close this PR and you can amend #1247.

PR #1262, opened 2024-11-12
Fix debian build - libzstd

EdgarModesto23:
https://github.com/pistacheio/pistache/pull/1247 | #1261 | #1200

This patch fixes issues mentioned by @kiplingw on https://github.com/pistacheio/pistache/pull/1247#issuecomment-2469344358. aswell as addressing the concerns made in #1261.

It removes the use of ```ZSTD_CLEVEL_DEFAULT(void)``` (requires libzstd v1.5.0+) in order to keep support for Jammy (which ships with [libzstd v1.4.8](https://packages.ubuntu.com/search?keywords=+libzstd-dev&searchon=names&suite=all&section=all)). And opts for the use of 0 instead. Which is a [special value](https://raw.githack.com/facebook/zstd/release/doc/zstd_manual.html#Chapter5) that means the compression algorithm should use a default value.


EdgarModesto23:
@kiplingw Awesome!. Please update me if anything comes up.

PR #1264, opened 2024-11-22
Launchpad fix for Windows support

dgreatwood:
There were 4 issues (1 error, 3 warnings) that showed up in the Launchpad Arm-32 build following merge of the Windows-support PR, all to do with integer sign and width conversions. This PR attempts to fix those issues.


dgreatwood:
Accidentally included a different change... will have another try soon

PR #1271, opened 2024-12-13
C++23 aligned storage

sjoubert:
This removes the usage of `std::aligned_storage` that is now deprecated in C++23 and replace it with an aligned array of `std::byte` that is C++17 compliant.

I've also replaced the manual destructor calls with `std::destroy_at` (replacing placement new with `std::construct_at` would require C++20 so these are untouched). It is in a separate commit, I can just remove it or make a separate PR if you prefer.

Side note: looking at various examples elsewhere I've seen a lot of usage of `std::launder` that seem missing here (mostly after each `reinterpret_cast`). I have no idea how and why it is useful, but you might want to take a look at some point.


sjoubert:
* I've fixed the `conventional-commits` error on my commits, but I don't know why it insists on checking a commit that is not part of this PR
* I don't really know how to interpret the `abidiff` output and if it is actually an issue (given pistache is not at a 1.0 release yet)

kiplingw (approved):

kiplingw:
Thanks @sjoubert. Looks fine to me. Thank you for those syntactical improvements. If there's no other feedback from my peers by the end of the weekend, remind me to merge if I forget.

sjoubert:
ping @kiplingw

PR #1272, opened 2024-12-22
Fix: Extended the homebrew Httplib Nofallback Workaround to Linux

dgreatwood:
As per PR #1270, "Enable use of subproject httplib in macOS despite nofallback", the same workaround is now applied in Linux also, so Homebrew in Linux (aka Linuxbrew) can build Pistache.


kiplingw:
Looks fine to me, @dgreatwood. If nothing new breaks in CI after the runners are done, feel free to merge. Good work.

PR #1279, opened 2025-01-08
feat: compiling using msys2 and mingw

Kreijstal:
Yes, it's not as pretty as msvc, but it's open source tooling :/. I think this satisfies the immediate need, and if there are further changes I think they might be patchable.

Adding compatibility to use windmc.

@dgreatwood This is what I can muster up to. I know it's not pretty, but this is of course a short term solution. If there are big changes in the manifest, you can ping me and I'll try to adapt it, it should be less work. But here you can cross-compile from linux, that ain't bad right? LMK what you think.


Kreijstal:
I'll try to fix CI.

dgreatwood:
Hi @Kreijstal -

Firstly, thanks so much for this effort! If I may so, I think it is quite promising.

May I try and describe your strategy? Please correct me if I have misunderstood:
1. You have produced pist_winlog.mc, which is an ".mc" version of pist_winlog.man. I guess you did this manually? Or if you used a tool, what did you use - is it something we could automate?
2. In pist_winlog_impl.cc, you create explicit functions (e.g. EventWritePSTCH_DEBUG_NL_AssumeEnabled) which in the mc.exe case are macros in the generated header file pist_winlog.h. Again, I presume you did this more or less by hand, but again LMK if you used a tool.

Initial questions:

1. Don't we need a header file that corresponds to pist_winlog_impl.cc, let's say pist_winlog_impl.h? Or could we make it a pure header file, pist_winlog_impl.h, inlining all the functions since they're all simple I think?
2. And wouldn't pist_winlog_impl.h need to be included in pist_syslog.cc instead of pist_winlog.h (which mc.exe generates) in the non-mc.exe case? I believe pist_winlog.h should not be in the build if mc.exe is not used.

More generally:

1. The script winscripts/gccsetup.ps1 will always find mc.exe and place it on the path, so perhaps we can add a "--nomcexe" option for gccsetup?
2. We could add tests to meson.build so an error is produced if pist_winlog.man is newer than pist_winlog.mc or pist_winlog_impl.h/.cc. WDYT?

Thanks again!

Kreijstal:
> Hi @Kreijstal -
> 
> Firstly, thanks so much for this effort! If I may so, I think it is quite promising.
> 
> May I try and describe your strategy? Please correct me if I have misunderstood:
> 
> 1. You have produced pist_winlog.mc, which is an ".mc" version of pist_winlog.man. I guess you did this manually? Or if you used a tool, what did you use - is it something we could automate?
> 2. In pist_winlog_impl.cc, you create explicit functions (e.g. EventWritePSTCH_DEBUG_NL_AssumeEnabled) which in the mc.exe case are macros in the generated header file pist_winlog.h. Again, I presume you did this more or less by hand, but again LMK if you used a tool.
> 
> Initial questions:
> 
> 1. Don't we need a header file that corresponds to pist_winlog_impl.cc, let's say pist_winlog_impl.h? Or could we make it a pure header file, pist_winlog_impl.h, inlining all the functions since they're all simple I think?
> 2. And wouldn't pist_winlog_impl.h need to be included in pist_syslog.cc instead of pist_winlog.h (which mc.exe generates) in the non-mc.exe case? I believe pist_winlog.h should not be in the build if mc.exe is not used.
> 
> More generally:
> 
> 1. The script winscripts/gccsetup.ps1 will always find mc.exe and place it on the path, so perhaps we can add a "--nomcexe" option for gccsetup?
> 2. We could add tests to meson.build so an error is produced if pist_winlog.man is newer than pist_winlog.mc or pist_winlog_impl.h/.cc. WDYT?
> 
> Thanks again!

It was kind of manual, was it macros? I guess we can convert those to macros...too, what is the script gccsetup for? If you have mc.exe you should just use that no?

There is no tool that did this work automatically, but it is reasonable to think one could be written... This would aid mimgw operability! With other projects that use manifests as well! So I really like the idea, this was however something more short term..

Kreijstal:
@dgreatwood I have addresses the changes, I added a --nomcexe option, 
But I also have to thank you, because you gave me the idea of writing the header file way shorter, and way more mantainable I think.

I also wrote a kind of convoluted step, to check the file age, it requires bash, however, not sure if this is undesirable. WDYT?

Kreijstal:
I managed to write a man to mc generator
```
import xml.etree.ElementTree as ET

# Define the namespaces
namespaces = {
    'ns': 'http://schemas.microsoft.com/win/2004/08/events',
    'win': 'http://manifests.microsoft.com/win/2004/08/windows/events',
    'xs': 'http://www.w3.org/2001/XMLSchema'
}

# Parse the .man file
tree = ET.parse('pist_winlog.man')
root = tree.getroot()

# Extract provider information
provider = root.find('.//ns:provider', namespaces)
if provider is None:
    print("Error: Could not find the <provider> element in the XML.")
    sys.exit(1)

provider_name = provider.get('name')
provider_guid = provider.get('guid')
provider_symbol = provider.get('symbol')

# Extract events
events = provider.find('ns:events', namespaces).findall('ns:event', namespaces)

# Extract localization strings
strings = {}
string_table = root.find('.//ns:stringTable', namespaces).findall('ns:string', namespaces)
for string in string_table:
    strings[string.get('id')] = string.get('value')

# Start building the .mc content
mc_content = ';// SPDX-FileCopyrightText: 2024 Duncan Greatwood\n'
mc_content += ';// SPDX-License-Identifier: Apache-2.0\n\n'
mc_content += 'MessageIdTypedef=DWORD\n\n'

# Provider Information
mc_content += f';// Provider Information\n'
mc_content += f';// Name: {provider_name}\n'
mc_content += f';// GUID: {provider_guid}\n'
mc_content += f';// Symbol: {provider_symbol}\n\n'

# Events
mc_content += ';// Events\n'
for event in events:
    message_id = event.get('value')
    symbolic_name = event.get('symbol')
    message = event.get('message')
    if message.startswith('$(string.Event.'):
        message_id_str = message.lstrip('$(string.Event.').rstrip(')')
        message_text = strings.get(message_id_str, '')
    else:
        message_text = message
    mc_content += f'MessageId={message_id}\n'
    mc_content += f'SymbolicName={symbolic_name}\n'
    mc_content += 'Language=English\n'
    mc_content += f'{message_text}\n.\n\n'
    # Add comments for channel, level, task, and template
    channel = event.get('channel')
    level = event.get('level')
    task = event.get('task')
    template = event.get('template')
    mc_content += f';// Channel: {channel}\n'
    mc_content += f';// Level: {level}\n'
    mc_content += f';// Task: {task}\n'
    mc_content += f';// Template: {template}\n\n'

# Generate Event Descriptors
mc_content += ';// Event Descriptors\n'
for event in events:
    message_id = event.get('value')
    symbolic_name = event.get('symbol')
    mc_content += f';static const EVENT_DESCRIPTOR EventDesc_{symbolic_name} = {{ {message_id}, 1, 0, 0, 0, 0, 0 }};\n'
mc_content += '\n'

# Generate Provider Handle and GUID
mc_content += ';// Provider Handle and GUID\n'
mc_content += ';static REGHANDLE PistacheProviderHandle = 0;\n'

# Format the GUID correctly
guid_parts = provider_guid.strip('{}').split('-')
guid_hex = [
    f'0x{guid_parts[0]}',  # Data1
    f'0x{guid_parts[1]}',  # Data2
    f'0x{guid_parts[2]}',  # Data3
    f'0x{guid_parts[3][0:2]}',  # Data4[0]
    f'0x{guid_parts[3][2:4]}',  # Data4[1]
    f'0x{guid_parts[4][0:2]}',  # Data4[2]
    f'0x{guid_parts[4][2:4]}',  # Data4[3]
    f'0x{guid_parts[4][4:6]}',  # Data4[4]
    f'0x{guid_parts[4][6:8]}',  # Data4[5]
    f'0x{guid_parts[4][8:10]}',  # Data4[6]
    f'0x{guid_parts[4][10:12]}'  # Data4[7]
]
mc_content += f';static const GUID PISTACHE_GUID =\n'
mc_content += f';{{ {guid_hex[0]}, {guid_hex[1]}, {guid_hex[2]}, {{ {guid_hex[3]}, {guid_hex[4]}, {guid_hex[5]}, {guid_hex[6]}, {guid_hex[7]}, {guid_hex[8]}, {guid_hex[9]}, {guid_hex[10]} }} }};\n\n'

# Generate Provider Registration Code
mc_content += ';// Provider Registration and Unregistration\n'
mc_content += ';static inline ULONG EventRegisterPistache_Provider() {\n'
mc_content += ';    return EventRegister(&PISTACHE_GUID, nullptr, nullptr, &PistacheProviderHandle);\n'
mc_content += ';}\n\n'
mc_content += ';static inline ULONG EventUnregisterPistache_Provider() {\n'
mc_content += ';    return EventUnregister(PistacheProviderHandle);\n'
mc_content += ';}\n\n'

# Generate Event Writing Functions
mc_content += ';// Event Writing Functions\n'
mc_content += ';#define GENERATE_EVENT_WRITE_FUNCTION(event_name, event_descriptor) \\\n'
mc_content += ';    static inline ULONG event_name(PCWSTR message) { \\\n'
mc_content += ';        EVENT_DATA_DESCRIPTOR descriptor; \\\n'
mc_content += ';        EventDataDescCreate(&descriptor, message, (ULONG)((wcslen(message) + 1) * sizeof(WCHAR))); \\\n'
mc_content += ';        return EventWrite(PistacheProviderHandle, &event_descriptor, 1, &descriptor); \\\n'
mc_content += ';    }\n\n'

# Generate AssumeEnabled Macros
mc_content += ';// AssumeEnabled Macros\n'
mc_content += ';#define GENERATE_ASSUME_ENABLED_MACRO(event_name) \\\n'
mc_content += ';    static inline ULONG event_name##_AssumeEnabled(PCWSTR message) { \\\n'
mc_content += ';        return event_name(message); \\\n'
mc_content += ';    }\n\n'

# Generate Functions for Each Event
for event in events:
    symbolic_name = event.get('symbol')
    mc_content += f';GENERATE_EVENT_WRITE_FUNCTION(EventWrite{symbolic_name}, EventDesc_{symbolic_name})\n'
    mc_content += f';GENERATE_ASSUME_ENABLED_MACRO(EventWrite{symbolic_name})\n\n'

# Write to .mc file
with open('pist_winlog.mc', 'w', encoding='utf-8') as mc_file:
    mc_file.write(mc_content)
```

dgreatwood:
> ...what is the script gccsetup for? If you have mc.exe you should just use that no?
gccsetup.ps1 checks whether gcc is already setup, and puts it on the path / installs it as needed. It also installs other components that are needed such as meson and even Visual Studio, including mc.exe.

> I added a --nomcexe option
Great.

> I managed to write a proto man to mc generator
Very cool. This does seem like the best approach if practical.

> I also wrote a kind of convoluted step, to check the file age, it requires bash, however, not sure if this is undesirable. WDYT?
So far I've avoided using bash scripts in the Windows cases. I have used some PowerShell scripts e.g. src/winlog/installman.ps1, which is invoked on certain conditions from src/meson.build. Perhaps a similar approach can also work here? And/or this is something I could look at myself.

From what I see (forgive me if I'm not looking correctly), you have some changes you're working on which you haven't pushed yet. Which is fine of course. LMK when you're ready / have pushed, and I can look more closely. 

Thanks again.

Kreijstal:
> > ...what is the script gccsetup for? If you have mc.exe you should just use that no?
> 
> gccsetup.ps1 checks whether gcc is already setup, and puts it on the path / installs it as needed. It also installs other components that are needed such as meson and even Visual Studio, including mc.exe.
> 
> > I added a --nomcexe option
> 
> Great.
> 
> > I managed to write a proto man to mc generator
> 
> Very cool. This does seem like the best approach if practical.
> 
> > I also wrote a kind of convoluted step, to check the file age, it requires bash, however, not sure if this is undesirable. WDYT?
> 
> So far I've avoided using bash scripts in the Windows cases. I have used some PowerShell scripts e.g. src/winlog/installman.ps1, which is invoked on certain conditions from src/meson.build. Perhaps a similar approach can also work here? And/or this is something I could look at myself.
> 
> From what I see (forgive me if I'm not looking correctly), you have some changes you're working on which you haven't pushed yet. Which is fine of course. LMK when you're ready / have pushed, and I can look more closely.
> 
> Thanks again.

Powershell is not really portable (Yes I know .net core exist but it is not provided by all linux distros because the compiler is not fully open source (It's not bootstrapable with an open source toolchain)), I think since I have done the python script I'll just generate the .mc file without doing the history check, This should work on most cases. I hope so at least!

dgreatwood:
> since I have done the python script I'll just generate the .mc file without doing the history check

That seems a good approach.

> Powershell is not really portable

Sure, but we should probably only be doing the .mc file generation for gcc builds on Windows, and PowerShell is available on Windows at least. But of course if you can generate everything then this is a non-issue as you said.

Kreijstal:
> > since I have done the python script I'll just generate the .mc file without doing the history check
> 
> That seems a good approach.
> 
> > Powershell is not really portable
> 
> Sure, but we should probably only be doing the .mc file generation for gcc builds on Windows, and PowerShell is available on Windows at least. But of course if you can generate everything then this is a non-issue as you said.



> > since I have done the python script I'll just generate the .mc file without doing the history check
> 
> That seems a good approach.
> 
> > Powershell is not really portable
> 
> Sure, but we should probably only be doing the .mc file generation for gcc builds on Windows, and PowerShell is available on Windows at least. But of course if you can generate everything then this is a non-issue as you said.

I think I'm done.

dgreatwood:
Very good. I'll plan to try this out tomorrow.

Thanks again!

dgreatwood:
Hi @Kreijstal -

I've been running your branch here. It builds quite nicely 👍 

I do see an issue with logging when running the "no mc.exe" configuration. In that configuration, when I look at the events deposited into the Windows ETW log, the event viewer shows me:
```
The description for Event ID 102 from source Pistache-Provider cannot be
found. Either the component that raises this event is not installed on your
local computer or the installation is corrupted. You can install or repair the
component on the local computer.
```

Whereas, when I run the "WITH mc.exe" configuration (but still building with gcc/msys2 otherwise), I see in the log:
```
Event 102
INFO Pistache exiting
```

Similarly for all the "Event 1" events - debug log messages - in the "with mc.exe" case I see meaningful messages like:
`DEBUG (11 PSTCH) reactor.cc:343 in shutdown(): Locking shutdown_mutex_ (at 0000012cf92fd0a8)`

To get the "description cannot be found" results, I did:
```
.\winscripts\gccsetup.ps1 -nomcexe
.\winscripts\mesbuilddebug.ps1
.\winscripts\mestestdebug.ps1
```
(You don't actually have to run the all the tests; e.g. cookie_test_3 should be sufficient to generate log entries).

I note that even when the manifest from the "with mc.exe" is installed, Windows Event Viewer still cannot interpret the events from the "no mc.exe" configuration. This suggests to me that the events written to the log are corrupt in the "no mc.exe" configuration, it's not purely a matter of the manifest; but it's possible something else is going on.

IMPORTANT: The file _Building on Windows.txt_, the _Logging_ section, explains how I captured and viewed these logs.

BTW, please note the pistachelog.dll and manifest are installed automatically by the build process (the build process invokes src\winlog\installman.ps1, which copies pistachelog.dll and executes wevtutil).

Could you take a look at this issue?

Thanks much!

Kreijstal:
I don't think I know the reason.. shame.

Kreijstal:
Could you share the header file produced by mc.exe, I do not have msvc.

Kreijstal:
Also, maybe that's a great test case to add to CI, I think this would help me as well, but I am not so sure how to retrieve and parse events on CI, maybe you have an idea?

dgreatwood:
> Could you share the header file produced by mc.exe, I do not have msvc.

I'm enclosing here.

BTW, to try yourself with Visual Studio, all you need to is just got to a PowerShell prompt, then execute the Pistache script:
`.\winscripts\msvcsetup.ps1`
It should install Visual Studio Community Edition and configure the shell to use it. You don't need the Windows GUI, you can do it all the via the PowerShell command line. Once msvcsetup.ps1 has run, then you can:
`.\winscripts\mesbuild.ps1`
OR
`.\winscripts\mesbuilddebug.ps1`
If you want to uninstall VS later, you can follow [these](https://learn.microsoft.com/en-us/visualstudio/install/uninstall-visual-studio?view=vs-2022) instructions. 

> Also, maybe that's a great test case to add to CI, I think this would help me as well, but I am not so sure how to retrieve and parse events on CI, maybe you have an idea?

Good idea! I can take a look.

[pist_winlog.h.zip](https://github.com/user-attachments/files/18384011/pist_winlog.h.zip)

dgreatwood:
Further on...
> Also, maybe that's a great test case to add to CI, I think this would help me as well, but I am not so sure how to retrieve and parse events on CI, maybe you have an idea?

Although I don't have time to flesh this out right now, it certainly looks doable. Contrast these two cases, first using mc.exe, second with windmc.exe.

With mc.exe:
```
PS> Get-WinEvent -path pistache-withmc.etl -Oldest | Select-Object -First 5


   ProviderName:

TimeCreated                      Id LevelDisplayName Message
-----------                      -- ---------------- -------
1/10/2025 11:30:33 AM             0 Information


   ProviderName: Pistache-Provider

TimeCreated                      Id LevelDisplayName Message
-----------                      -- ---------------- -------
1/10/2025 11:31:05 AM           102 Information      INFO Pistache start. INFO and up log messages visible in Event Viewer. See pist_s...
1/10/2025 11:31:05 AM             1 Verbose          DEBUG (253c PSTCH) pist_sockfns.cc:243 in pist_sock_startup_check(): Locking lWsa...
1/10/2025 11:31:05 AM             1 Verbose          DEBUG (253c PSTCH) pist_check.cc:273 in ~GuardAndDbgLog(): lWsaStartupDoneMutex (...
1/10/2025 11:31:05 AM             1 Verbose          DEBUG (253c PSTCH) eventmeth.cc:776 in create(): <Ctr:1 Fri Jan 10 19:31:05 2025,...
```

Then with windmc.exe:
```
PS> Get-WinEvent -path pistache.etl -Oldest | Select-Object -First 5


   ProviderName:

TimeCreated                      Id LevelDisplayName Message
-----------                      -- ---------------- -------
1/10/2025 11:08:49 AM             0 Information


   ProviderName: Pistache-Provider

TimeCreated                      Id LevelDisplayName Message
-----------                      -- ---------------- -------
1/10/2025 11:09:47 AM           102 Information
1/10/2025 11:09:47 AM             1 Information
1/10/2025 11:09:47 AM             1 Information
1/10/2025 11:09:47 AM             1 Information
```

I think I can integrate this as a test into CI, unless you'd like to :-)

Kreijstal:
> I think I can integrate this as a test into CI, unless you'd like to :-)

Please do.
I discovered a small bug on the converter script, I hope the messages are viewable, unfortunately couldn't test with the powershell script.

```
kreijstal@windows UCRT64 ~/git/pistache
# ./winscripts/msvcsetup.ps1
Looking for vcpkg (may take some time)
```
simply hung around too long, should I've been more patient?

Please test to see if the messages are not there. I couldn't test myself. hehe

dgreatwood:
`Looking for vcpkg (may take some time)`
> simply hung around too long, should I've been more patient?

Yes, this will scan the whole system drive, and the whole home drive if the home drive is different to system drive, looking for vcpkg. 

Thank you for letting it run longer on the second attempt, it looks like it did eventually install vcpkg. 

It managed to download and run the Visual Studio installer, but (per the output) the VS installer returned an error, apparently a write error. **Could you help me understand what caused that?**

Is your disk full, or nearly full? Either your system drive (probably C:) or any other drive?

Was the shell you ran the script from a local shell, or remote? Did the shell have administrative privileges, or not? (It doesn't have to have admin privileges for the script, but the code path is different if it does not).

From your prompt, can you try:
`./vs_community.exe --add Microsoft.VisualStudio.Workload.NativeDesktop --includeRecommended --wait`
(_vs_community.exe_ will most likely have been downloaded to your user's home directory)

This should pop you into the GUI to install VS. Just "OK/Continue" each thing as prompted. You may see a more meaningful error this way.

Thanks!

Meanwhile, I will try with your update.

P.S. If you want to try it yourself, you should be able to do the logging as described in the file _Building on Windows.txt_, the _Logging_ section, using your build, ideally the debug version of your build but even the standard version should produce something - you don't need Visual Studio for that.

Kreijstal:
> `Looking for vcpkg (may take some time)`
> 
> > simply hung around too long, should I've been more patient?
> 
> Yes, this will scan the whole system drive, and the whole home drive if the home drive is different to system drive, looking for vcpkg.
> 
> Thank you for letting it run longer on the second attempt, it looks like it did eventually install vcpkg.
> 
> It managed to download and run the Visual Studio installer, but (per the output) the VS installer returned an error, apparently a write error. **Could you help me understand what caused that?**
> 
> Is your disk full, or nearly full? Either your system drive (probably C:) or any other drive?
> 
> Was the shell you ran the script from a local shell, or remote? Did the shell have administrative privileges, or not? (It doesn't have to have admin privileges for the script, but the code path is different if it does not).
> 
> From your prompt, can you try: `./vs_community.exe --add Microsoft.VisualStudio.Workload.NativeDesktop --includeRecommended --wait` (_vs_community.exe_ will most likely have been downloaded to your user's home directory)
> 
> This should pop you into the GUI to install VS. Just "OK/Continue" each thing as prompted. You may see a more meaningful error this way.
> 
> Thanks!
> 
> Meanwhile, I will try with your update.
> 
> P.S. If you want to try it yourself, you should be able to do the logging as described in the file _Building on Windows.txt_, the _Logging_ section, using your build, ideally the debug version of your build but even the standard version should produce something - you don't need Visual Studio for that.

Ahh that might be the case, I'm running windows sshd as a service, this means normal windowed applications do not work, because I'm in "service mode" and not as "user mode". 
https://softwareengineering.stackexchange.com/questions/180833/why-cant-windows-services-have-a-gui
I mean not much different than CI, thing is headless.

Essentially I connect through windows through ssh :D (using msys2, I compile all sort of things from there), but it doesn't allow me to have anything with gui on it. and it is a big computer so I don't need to wait a long time for compilations)

Also, about C being full, yeah it is almost full, I have more space on the E drive though.

dgreatwood:
> Essentially I connect through windows through ssh :D (using msys2, I compile all sort of things from there), but it doesn't allow me to have anything with gui on it. and it is a big computer so I don't need to wait a long time for compilations)

That should be fine, I also do most of the dev through ssh, and of course the scripts are designed to be run through the command line. Occasionally GUI is useful for diagnostics...

> Also, about C being full, yeah it is almost full, I have more space on the E drive though.

Ah, that's probably the issue then. What's the value of your `$env:ProgramFiles` and `${env:ProgramFiles(x86)}` ? If they're on C:, then that's likely the problem.

You could try (though I've never done this...):
`./vs_community.exe --installPath E:\MSVC --add Microsoft.VisualStudio.Workload.NativeDesktop --includeRecommended --quiet --wait`

Or could you free up space on C: ?

dgreatwood:
Regarding the updated branch, I still seem to have the same (failing) result:
```
logman start -ets Pistache -p "Pistache-Provider" 0 0 -o pistache11.etl
./run_cookie_test_3.exe
logman stop Pistache -ets
```

Then:
```
Get-WinEvent -path pistache11.etl -Oldest | Select-Object -First 5

   ProviderName:

TimeCreated                      Id LevelDisplayName Message
-----------                      -- ---------------- -------
1/11/2025 11:00:53 AM             0 Information


   ProviderName: Pistache-Provider

TimeCreated                      Id LevelDisplayName Message
-----------                      -- ---------------- -------
1/11/2025 11:00:59 AM           102 Information
1/11/2025 11:00:59 AM             1 Information
1/11/2025 11:00:59 AM             1 Information
1/11/2025 11:00:59 AM             1 Information
```

("Message" field is empty, and for Id=1 entries the "LevelDisplayName" is incorrect - the LevelDisplayName should be _Verbose_ for Id=1).

dgreatwood:
Regarding the failure to install Visual Studio with winscripts/msvcsetup.ps1, prompted by your report I tried out msvcsetup.ps1 on Windows 10, and the problem reproduced just as you described. 

This doesn't happen on Windows 11, or Windows Server 2019, but it does happen on Windows 10.

We have merged a small change into Pistache master (PR #1283) in msvcsetup.ps1. It's a one line change, plus a comment. 

Is it possible for you to test whether that change addresses the issue in your environment? I.e. try running msvcsetup.ps1 again and see if Visual Studio installs correctly?

Issue #1006, opened 2021-10-20
[BUG] Server ResponseStream in separate thread gives Assertion Error: could not find write data


tn-jed:
## Current Behavior
Server streaming back to client with a ResponseStream from a separate thread, the following exception is thrown:
```
terminate called after throwing an instance of 'std::runtime_error'
  what():  Assertion Error: could not find write data
```
Same ResponseStream works fine if it's in the original thread.

## Expected behavior
Expect that the new thread should be able to stream data back to the client indefinitely via the ResponseStream, even if multiple `Http::flush` are issued.

The `Http::flush` stream operation should force the server to send to the client all of data that's buffered so far but _keep the connection open_ so that more data can be streamed.

Two use cases come to mind:
* Time sensitive data over a long-running session
* Sending large responses without needing to buffer the entire response on the server (send chunks at a time)

## Steps to reproduce the bug
```
  void long_response(const Rest::Request& request, Http::ResponseWriter response){
    std::thread worker([](Http::ResponseWriter writer) {
      auto stream = writer.stream(Http::Code::Ok);
      for(int i=0; i<10; i++){
        stream << "Data...\n";
        stream << Http::flush;
        sleep(2);
      }
      stream << Http::ends;
    }, std::move(response));
    worker.detach();
  }
```
The client sees the first two `Data...` lines before the exception is thrown.

Seems to be triggered by the flush.

## Context (Environment)

- Pistache version: 0.0.002-20211002 `ff9db0d9439a4411b24541d97a937968f384a4d3`
- Operating System: Ubuntu 20.04.3 LTS 
- Architecture: amd64
- Installed from: Built from source
- Used `pkg-config`: no

Screenshots, logs, stacktraces or other relevant information here.

PR #1280, opened 2025-01-10
fix ResponseStream when used from other threads

bkueng:
I'm running into the same issue as https://github.com/pistacheio/pistache/issues/1006: I'm trying to flush a stream from another thread.

This PR adds 3 changes:
#### do not add NotifyOn::Write when flushing buffers
- if called from the worker thread (doing the epoll) this has no effect
  as it's cleared before going back to epoll
- if called from another thread, it causes an unnecessary wakeup of the
  worker thread. The worker thread then throws an exception
  (`"Assertion Error: could not find write data"`), as the `toWrite` data is
  already removed.

In case the flush call (`asyncWriteImpl`) cannot finish completely due to
`EWOULDBLOCK`, it still sets `NotifyOn::Write`.

#### remove "Assertion Error: could not find write data" assertion
Even with the previous change there is still a case where this can happen:
- the worker thread returns from epoll with a write notification
  (e.g. due to a previous `EWOULDBLOCK`)
- if at the same time another thread calls flush, which writes all data
- then the worker thread finds an empty `toWrite`

#### prevent busy-loop in case of EWOULDBLOCK in asyncWriteImpl
By making sure the loop is exited. As `NotifyOn::Write` is set, the worker
thread will then be notified when data can be written.

Tested with code instrumentation by manually setting errno = EAGAIN,
and checking that the worker thread then gets woken up and completes the
write.

Fixes: https://github.com/pistacheio/pistache/issues/1006


Tachi107:
Thanks for the patch! Neatly separated in small, useful commits. Looks good to me :)

kiplingw (changes requested):
Thanks @bkueng. Please remember to bump the patch version, per our [documentation](https://github.com/pistacheio/pistache/?tab=readme-ov-file#versioning). Otherwise it looks fine.

bkueng:
> Please remember to bump the patch version, per our [documentation](https://github.com/pistacheio/pistache/?tab=readme-ov-file#versioning)

Done, let me know if you rather want me to amend that to the existing commit instead of a separate one.

Tachi107:
Thanks!

Issue #1281, opened 2025-01-13
Error compiling pistache on windows


Kreijstal:
Hello, it's me again, I am trying in another windows machine, (Without admin rights). I can install msys2 tools, so I tried reproducing the same steps, but maybe there is something going around with powershell, this was however on win10 and not in win11
```
"C:/rtools44/ucrt64/bin/meson" "--internal" "exe" "--unpickle" "C:/rtools44/home/SYSTEM/pistache/bui
ld/meson-private/meson_exe_powershell.exe_603cd33c7e8fa4e3e9a54f8f255f85658890ce9b.dat"
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "C:/rtools44/ucrt64/bin/meson.exe/__main__.py", line 7, in <module>
  File "C:/rtools44/ucrt64/lib/python3.12/site-packages/mesonbuild/mesonmain.py", line 293, in main
    return run(sys.argv[1:], launcher)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:/rtools44/ucrt64/lib/python3.12/site-packages/mesonbuild/mesonmain.py", line 281, in run
    return run_script_command(args[1], args[2:])
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:/rtools44/ucrt64/lib/python3.12/site-packages/mesonbuild/mesonmain.py", line 222, in run_s
cript_command
    return module.run(script_args)
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "C:/rtools44/ucrt64/lib/python3.12/site-packages/mesonbuild/scripts/meson_exe.py", line 112,
in run
    return run_exe(exe)
           ^^^^^^^^^^^^
  File "C:/rtools44/ucrt64/lib/python3.12/site-packages/mesonbuild/scripts/meson_exe.py", line 78, i
n run_exe
    print(stderr.decode(encoding=encoding, errors='replace'))
  File "C:/rtools44/ucrt64/lib/python3.12/encodings/cp1252.py", line 19, in encode
    return codecs.charmap_encode(input,self.errors,encoding_table)[0]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1282, opened 2025-01-14
Fix: In rest_server_test Allow "view-localhost" hostname

dgreatwood:
The test already allowed "localhost" and "ip6-localhost". "view-localhost" was seen on Windows 10 by @Kreijstal . The test now allows any hostname ending in "localhost" to be accepted as a hostname.

Addresses part of issue #1281.


kiplingw (approved):

tests/rest_server_test.cc
```diff
@@ -117,14 +117,14 @@ TEST(rest_server_test, basic_test)
     ASSERT_EQ(res->status, 200);
 
     // TODO: Clean this up to use proper gtest macros.
-    // NOTE: res->body is "ip6-localhost" on some architectures.
-    if (res->body == "ip6-localhost")
+    if ((res->body.length() >= 9) && // 9 being len of "localhost"
```
Tachi107, line 120: instead of hardcoding 9, you could write `strlen("localhost")`. It's self-documenting

Tachi107: Or, even better, you can use `EXPECT_THAT(res->body, testing::EndsWith("localhost"))`

PR #1285, opened 2025-01-21
fix: don't break C++23's std::unreachable

sjoubert:
If "winornix.h" were to be included before <utility> then the standard definition of std::unreachable is broken because the 'unreachable' is replaced by the Pistache macro content.
This patch avoids using a macro, puts the Pistache implementation in a detail namespace and simply pulls in the standard function if available.

I'm not sure why the commented out non-macro version was not used. Maybe it was not well tested and the author preferred to keep the original macro version for GCC. I think the non-macro version is better but if you prefer to keep a macro version somehow I can modify my patch and introduce a prefixed macro instead of a "standard" name to avoid the conflict.


Tachi107 (approved):
Looks good to me!

kiplingw (approved):

dgreatwood:
This is a good change to maintain C++23 compatibility.

Regarding:
> I'm not sure why the commented out non-macro version was not used.

Prior to the definition of unreachable() in winornix.h, unreachable() was defined in include/pistache/common.h:
`#define unreachable() __builtin_unreachable()`
That line was committed in 2015, Commit [5b65633](https://github.com/pistacheio/pistache/commit/5b65633daf9762c392a2551b0d819bdc05e6a238).

I believe the reason the non-macro version was not used in winornix.h was simply to avoid making an unneeded change in the gcc case.

I did compile your branch successfully in openbsd as an extra check (unfortunately our CI does not include BSD at present).

I think we are good to go.

dgreatwood (approved):
Approved as per comment

PR #1286, opened 2025-01-22
meson: check for availability of dladdr

eli-schwartz:
Delete erroneous comment claiming that compiler.has_function is broken. This is based on a misunderstanding of the original logfile. See: https://github.com/pistacheio/pistache/pull/1252#discussion_r1924598159

Checking for dladdr works fine. Checking for `dependency('dl')` on versions of meson that are too old to support it does not work, and this is what happened on RedHat ubi-8.

/cc @dgreatwood

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -207,26 +207,22 @@ if host_machine.system() != 'windows'
         # which is called by PS_LogWoBreak, used in turn by the stack-trace
         # logging macro PS_LOG_WO_BREAK_LIMITED and its derivatives. Issue
         # #1230.
-        # Note: If 'dl' is not available, per Meson it suggests that the
-        # functionality is provided by libc
-
-        # It would be nice to use compiler.has_function('dladdr') to test if
-        # we need to add libdl, but unfortunately that approach seems to break
-        # certain Redhat builds, specifically the Redhat builds that use
-        # ubi-8. In such cases, it appears meson creates a test file that
-        # includes the comment:
-        #   With some toolchains ... the compiler provides various builtins
-        #   which are not really implemented... [If] the user provides a
-        #   header, including the header didn't lead to the function being
-        #   defined, and the function we are checking isn't a builtin itself,
-        #   we assume the builtin is not functional and error out
-        # To avoid generating such an error, we take the simpler approach of
-        # trying to add libdl but making it optional (i.e. not required).
         if meson.version().version_compare('>=0.62.0')
-          deps_libpistache += dependency('dl', required: false)
+          dl_dep = dependency('dl')
         else
-          deps_libpistache += compiler.find_library('dl', required: false)
+          dl_dep = compiler.find_library('dl', required: false)
         endif
+        # Note: the Single Unix Specification version 2, issue 5 (1997) defines dlopen
+        # but not dladdr. POSIX 2024 issue 8 requires dladdr. Older systems might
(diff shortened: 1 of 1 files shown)


dgreatwood:
@eli-schwartz - thanks for cogent analysis and update.

I had just one remaining question. Would it be OK to guard this new meson.build code with an initial test of whether dladdr is already available? In particular, in the meson.version() < 0.62 case, if dladdr is already available, we are sending meson to look for library 'dl', and then, if library dl is found, we test it for dl_addr and add it as a dependency, none of which (presumably?)  we'd actually need to do in that scenario. In other words it would look like:
```
if host_machine.system() != 'windows'
    if not compiler.has_function('dladdr')
        # libdl may be required for function dladdr, used in logStackTrace,
        # which is called by PS_LogWoBreak, used in turn by the stack-trace
        # logging macro PS_LOG_WO_BREAK_LIMITED and its derivatives. Issue
        # #1230.
        if meson.version().version_compare('>=0.62.0')
          dl_dep = dependency('dl')
        else
          dl_dep = compiler.find_library('dl', required: false)
        endif
        # Note: the Single Unix Specification version 2, issue 5 (1997) defines dlopen
        # but not dladdr. POSIX 2024 issue 8 requires dladdr. Older systems might
        # therefore lack what we need, even if libdl exists.
        #
        # Meson's "dl" check looks for dlopen. We assume that, if dladdr exists
        # at all, it is in the same DSO (whether it is libc or -ldl) as dlopen.
        # But it may not exist at all, so check to be sure.
        if not compiler.has_function('dladdr', prefix: '#include <dlfcn.h>', dependencies: dl_dep)
          error('dladdr could not be detected')
        endif
        deps_libpistache += dl_dep
    endif
endif
```

WDYT?

eli-schwartz:
> I had just one remaining question. Would it be OK to guard this new meson.build code with an initial test of whether dladdr is already available? In particular, in the meson.version() < 0.62 case, if dladdr is already available, we are sending meson to look for library 'dl', and then, if library dl is found, we test it for dl_addr and add it as a dependency, none of which (presumably?) we'd actually need to do in that scenario.

On the flip side, in the >0.62 case we are now running 3 compile checks instead of 2.

Keep in mind: if dladdr is already available, then there will not be a separate "dl" library at all. On some systems, where there used to be one but isn't any longer (such as very recent Linux glibc systems), `-ldl` will add 4 bytes to the linker command line but resolves to /usr/lib64/libdl.a which is an empty archive containing no objects.

```
$ cat /usr/lib64/libdl.a 
!<arch>
$ ls /usr/lib64/libdl.*
/usr/lib64/libdl.a  /usr/lib64/libdl.so.2
```

You cannot link directly to the shared library since `-ldl` doesn't search for .so.2, past versions of glibc provided both a libdl.a that contained statically linkable object files and a libdl.so that was a symlink to libdl.so.2 (which contained a variety of symbols, but now is just a compatibility placeholder to avoid breaking existing compiled binaries with a DT_NEEDED entry for libdl.so.2)

There are no consequences for the resulting installable pistache and it's just there as a compatibility shim for users of old meson versions, so I figured it's easier to change less. That way once you bump the minimum supported meson version you can just delete the version check without rewriting the surrounding code.

dgreatwood:
> That way once you bump the minimum supported meson version you can just delete the version check without rewriting the surrounding code.

OK

PR #1287, opened 2025-02-05
Update README.md

bnjmnjrk:
It is good to know such specifics about minimal versions when working with old distributions on restricted systems, because it can save a lot of time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ A benchmark comparison of Pistache to other C++ RESTful APIs was created by gute
 
 Pistache has the following third party dependencies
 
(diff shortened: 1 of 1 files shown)


kiplingw:
@Tachi107, correct me if wrong, but didn't we deprecate CMake usage a long time ago?

dgreatwood:
I did make cmake work for the original macOS + libevent version, but haven't maintained it since. It isn't in CI/workflow, I didn't test it for BSD, and I never implemented it for Windows either. It is deprecated as you guys said...

PR #1289, opened 2025-02-12
Fix: Make Unversioned-DLL Target Build In Windows with Meson 1.7.0

dgreatwood:
As seen in github workflow, when we move from meson v1.6.1 to meson v1.7.0, the custom target unversioned_dll, defined in tests/meson.build and needed for test execution, no longer gets built. As a workaround, we have added "build_by_default: true" for the unversioned_dll custom target, causing unversioned_dll to be built once again as part of the windows.yaml workflow.

Prior to this change -

Failing with meson 1.7.0:
https://github.com/pistacheio/pistache/actions/runs/13157241901/job/36735628578

Succeeding with meson v1.6.1:
https://github.com/pistacheio/pistache/actions/runs/12913131184/job/36009642858

ALSO - when unversioned_dll is built, we now copy to the unversioned DLL name whenever the pistache DLL is newer than the unversioned DLL; previously would only copy when the unversioned DLL did not exist, which of course risks leaving a stale version of the unversioned DLL in place if the pistache DLL has been rebuilt.


kiplingw (approved):

eli-schwartz:
> As seen in github workflow, when we move from meson v1.6.1 to meson v1.7.0, the custom target unversioned_dll, defined in tests/meson.build and needed for test execution, no longer gets built. As a workaround, we have added "build_by_default: true" for the unversioned_dll custom target, causing unversioned_dll to be built once again as part of the windows.yaml workflow.

This behavior change was specifically called out in the release notes: https://mesonbuild.com/Release-notes-for-1-7-0.html#test-targets-no-longer-built-by-default

The issue is e.g. here: https://github.com/pistacheio/pistache/blob/c1e0d761d797f727d5cafb604f7d749d65b66f53/.github/workflows/windows.yaml#L343

no-rebuild prevents building test-only artifacts. You may want to follow the recommendations from the release notes and have ninja build `ninja all meson-test-prereq` instead of just "all", assuming you have specific reasons for avoiding dependency calculation and re-execution during test runs.

dgreatwood:
Thanks @eli-schwartz, very helpful.

I'm still not sure I 100% understand however. In the meson release notes, from your reference, it states:
> The behavior is now ... to only build the actual set of targets that are used by the test.

In our _tests/meson.build_ though we have:
```
        test(
                test_name,
                executable(
                        'run_'+test_name,
			test_name+'.cc',
                        link_args: test_link_args,
			dependencies: [
                                pistache_dep,
                                tests_helpers_dep,
                                deps_libpistache,
                                gtest_main_dep,
				gmock_dep,
                                curl_dep,
                                cpp_httplib_dep,
                                brotli_dep,
                                zstd_dep
                        ]
                ),
                depends: unversioned_dll_cts,
                timeout: 600,
                workdir: meson.current_build_dir(),
                is_parallel: false,
                kwargs: suite
        )

# Where
    unversioned_dll_ct1 = custom_target('unversioned_dll', ...)
# And
    unversioned_dll_cts = []
    unversioned_dll_cts += unversioned_dll_ct1
```

Doesn't this (the `test` command's `depends`) mean that meson would know that it needs to build the custom target `unversioned_dll_ct1` before running the test? What am I missing?

eli-schwartz:
> Doesn't this (the `test` command's `depends`) mean that meson would know that it needs to build the custom target `unversioned_dll_ct1` before running the test? What am I missing?

Like I said:

The issue is e.g. here: https://github.com/pistacheio/pistache/blob/c1e0d761d797f727d5cafb604f7d749d65b66f53/.github/workflows/windows.yaml#L343

no-rebuild prevents building test-only artifacts. You may want to follow the recommendations from the release notes and have ninja build `ninja all meson-test-prereq` instead of just "all", assuming you have specific reasons for avoiding dependency calculation and re-execution during test runs.

dgreatwood:
@eli-schwartz - sure, that's clear. I guess I'm not following the rationale whereby, in a default build phase, tests are built but known test prerequisites are not.

In any case, taking the newly defined behavior as a given, I made the change as per the release notes and your comments - i.e. use `meson-test-prereq` in _windows.yaml_,  and remove the `build_by_default: true` from _tests/meson.build_. 

All seems to be building correctly 👍 

Thanks much for pointing this out!

P.S. JFYI, the reason for using `--no-rebuild` in the test phase is just to avoid any risk of the WIndows development environment not working predictably in the test phase. Configuring the development enviroment within the _windows.yaml_ file is rather painful, and it seems less fragile to avoid configuring it again if possible during the test phase. Which we've avoided needing to do with the `meson-test-prereq` approach.

eli-schwartz:
> @eli-schwartz - sure, that's clear. I guess I'm not following the rationale whereby, in a default build phase, tests are built but known test prerequisites are not.

Any target that is `install: true` is also by inference `build_by_default: true`.

For historical reasons, executable() is by default `build_by_default: true` but custom_target() is not. You can explicitly annotate executables which are used solely by testsuites, as `build_by_default: false`. If you do so, then people who download and run `meson setup builddir/ && ninja -C builddir/ all && meson install -C builddir/` will not compile test executables they don't need -- but if they run `meson test` then the test executables (and the test prerequisites in general) will be built first and then tested.

This is useful to avoid performing "unnecessary" work by people that aren't intending to run testsuites, or for distribution packagers where the package manager tooling supports an optional test stage that can be skipped. It's especially useful if projects don't have a `-D PISTACHE_BUILD_TESTS=true` option because the testsuite doesn't require external dependencies, as you don't need to pass a special setup option just to avoid building unnecessary executables.

eli-schwartz:
> Configuring the development enviroment within the _windows.yaml_ file is rather painful, and it seems less fragile to avoid configuring it again if possible during the test phase.

Makes sense. Meson has some logic to try to launch a VS Dev Shell itself, but the whole VS thing is a pile of hacks any way you slice it. :)

PR #1291, opened 2025-02-14
Experimental HTTPS/SSL/TLS Support for the Pistache Client

dgreatwood:
Basic and experimental but functional HTTPS client for Pistache derived from 3rd-party sample code. Passes tests against Pistache HTTPS server; for fetching search results from google.com, bing.com and aol.com; and (test not included in check in) can read and write blobs from GCP and Azure storage APIs.

Move some prior code to Pistache. Prior code derives from:
 - https://codereview.stackexchange.com/questions/108600/complete-async-openssl-example
 - https://wiki.openssl.org/index.php/SSL/TLS_Client.

client.h/.cc:
 - Detect when HTTPS is/is not being used
 - Added class FdOrSslConn, which can hold either an Fd (for a regular connection) or an ssl connection (SslConnection). Adjust rest of code to use FdOrSslConn as appropriate.
 - Add setHostChainPemFile to API for application to provide cert pem file
 - Make sure not to remove the subdomain/"www" from URIs (as per www.google.com) when constructing a REST request; see code comment for more info.
 - Adjust definitions of "connect" so can be used for SSL or non SSL.
 - In SSL case, invoke SSL-related functions (e.g. sslRawSend) where appropriate
 - Adjust splitUrl for HTTPS cases
 - Adjust SSL features for Windows
 - Enable "localhost" to be treated as a special case for SSL verification (localhost not expected to have valid cert by default)
 - Allow SSL verification to be turned on/off via Pistache interface

net.h/.cc:
 - Add "enum class Scheme" in the Address type. Address can now hold a scheme (HTTP or HTTPS) or leave the scheme unspecified.
 - Allow addresses to hold an HTTPS URI

pist_sockfns.h/.cc and winornix.h
 - Add pist_sock_select, used by SSL client code, and associated macros

ssl_async.h/.cc:
 - Handles low-level/socket-level SSL client actions
 - Includes license information regarding the sample code utilized

sslclient.h/.cc
 - Encapsulates SSL connection operations - connect, send, receive and close
 - Turn HTTPS verification on/off via interface

https_client_test.cc
 - Copy all the tests from http_client_test and make them use SSL
 - Add tests for fetching from google.com, bing.com and aol.com
 - Add test for SSL file-location verification
 - Add test for checking on verification failure when appropriate
 - Add test for getting a HTTP 404 error doing a badly formed request from https://www.google.com

tests/certs/meson.build
 - Add cacert.pem, include in update script
 - Add cacert.license, provide SPDX info for cacert.pem

http_defs.cc
 - Cope with HTTP using Google's date format

bldscripts/mesbuildnossl.sh etc.
 - Add convenience scripts to build and test non-SSL builds


bldscripts/mesbuildnossldebug.sh
```diff
@@ -0,0 +1,37 @@
...
+else
+    meson setup ${MESON_BUILD_DIR} \
+    --buildtype=debug \
+    -DPISTACHE_BUILD_EXAMPLES=true \
+    -DPISTACHE_BUILD_TESTS=true \
+    -DPISTACHE_BUILD_DOCS=false \
+    -DPISTACHE_USE_CONTENT_ENCODING_DEFLATE=true \
+    -DPISTACHE_USE_CONTENT_ENCODING_BROTLI=true \
+    -DPISTACHE_USE_CONTENT_ENCODING_ZSTD=true \
+    -DPISTACHE_DEBUG=true \
```
kiplingw, line 31: Be careful of unterminated '\' here.

bldscripts/mesbuildnossldebug.sh
```diff
@@ -0,0 +1,37 @@
...
+    --prefix="${MESON_PREFIX_DIR}" \
```
kiplingw, line 32: Be careful of unterminated '' here.

dgreatwood: Changed

src/client/client.cc
```diff
@@ -66,8 +69,26 @@ namespace Pistache::Http::Experimental
...
+            // whatever), work back to the domain, and then isolate the
+            // subdomain part of the URL. See https://publicsuffix.org and e.g.
+            // https://stackoverflow.com/questions/288810/
+            //                                     get-the-subdomain-from-a-url
+            // TODO !!!!
+            if (remove_subdomain)
+            {
+                match_string("www", cursor);
+                match_literal('.', cursor);
+            }
```
Tachi107, line 91: I'm not sure I understand the reason behind this change. Also, why not fixing the underlying issue directly? Maybe in a different PR, so that it can me reviewed and merged independently from this big one :)

Also, it's API-breaking. Maybe you could add the new boolean as the last parameter, with a default value of `true`, as not to change behaviour.

src/client/client.cc
```diff
@@ -151,8 +174,12 @@ namespace Pistache::Http::Experimental
             writeCookies(streamBuf, request.cookies());
             writeHeaders(streamBuf, request.headers());
 
+            std::string host_str(
+                std::string(host) + std::string((is_https && (std::string(host).find(':') == std::string::npos)) ? ":443" : ""));
```
Tachi107, line 178: This one liner is a bit cryptic

src/client/client.cc
```diff
@@ -218,7 +245,25 @@ namespace Pistache::Http::Experimental
...
                 , addr_len(_addr_len)
             {
-                std::memcpy(&addr, _addr, addr_len);
+                // Note: Sanitizer may object to invocation of memcpy with null
+                // _addr, even if addr_len is zero
+                if (_addr_len)
+                {
+                    // Note - the "cast" here avoids a warning of mismatched
+                    // signedness in Windows
+                    if (addr_len > (static_cast<socklen_t>(sizeof(addr))))
```
Tachi107, line 254: Here, with `sizeof(addr)`, you're taking the size of the pointer. Is this intentional?

src/client/client.cc
```diff
@@ -399,22 +444,48 @@ namespace Pistache::Http::Experimental
...
+            const PST_SSIZE_T len    = buffer.size() - totalWritten;
+            PST_SSIZE_T bytesWritten = -1;
+
+#ifdef PISTACHE_USE_SSL
+            if (conn->isSsl())
+            {
+// Comment out PST_SSL_REQ_DBG to suppress log of send content
+#ifdef DEBUG
+// #define PST_SSL_REQ_DBG DEBUG // Comment in/out as desired
+#endif
```
Tachi107, line 471: Editing source code to change functionality is a bit odd. I'd just remove this code; users can always pass `-DPST_SSL_REQ_DBG` if desired

src/client/client.cc
```diff
@@ -709,37 +872,64 @@ namespace Pistache::Http::Experimental
...
+    void Connection::connect(Address::Scheme scheme,
+#ifdef PISTACHE_USE_SSL
+                             SslVerification sslVerification,
+#endif // PISTACHE_USE_SSL
+                             const std::string& domain,
+                             const std::string* page)
+    {
+        const Address addr(helpers::httpAddr(
+            domain,
+            (Address::Scheme::Https == scheme) ? 443 : 0, // default port
```
Tachi107, line 912: nit

```suggestion
            (scheme == Address::Scheme::Https) ? 443 : 0, // default port
```

src/client/ssl_async.cc
```diff
@@ -0,0 +1,1266 @@
...
+  }
+
+  return NEITHER;
+}
+
+// ---------------------------------------------------------------------------
+
+
+// Helper function to be called only by initOpenSslIfNotAlready
+static void init_openssl_library(void)
```
Tachi107, line 370: Wait, is this needed?

Library initialization is a hairy problem, especially for libraries (like pistache).

What if Pistache is embedded into another program, which performs its own OpenSSL initialization?

src/client/ssl_async.cc
```diff
@@ -0,0 +1,1266 @@
...
+{
+    /* https://www.openssl.org/docs/ssl/SSL_library_init.html */
+    (void)SSL_library_init();
+    /* Cannot fail (always returns success) ??? */
+
+    OpenSSL_add_all_algorithms();
```
Tachi107, line 376: These two functions have been deprecated for a while (didn't check the others).

Now there's <https://docs.openssl.org/3.3/man3/OPENSSL_init_ssl/>, but maybe it's not needed for our use case?

src/client/ssl_async.cc
Tachi107, line 1: This file is really big and unfortunately I don't have enough free time to look at it in its entirety now, but for what I've seen it seems to be using lots of deprecated OpenSSL interfaces... Have you tried compiling it with `OPENSSL_NO_DEPRECATED` defined?

src/client/ssl_async.cc
kiplingw: I'm not sure where `init_openssl_library` is defined, but if it's calling `SSL_library_init` then it was [deprecated](https://docs.openssl.org/master/man3/SSL_library_init/#warnings) in OpenSSL 1.1.0 by `OPENSSL_init_ssl`.

src/client/client.cc
dgreatwood: The reason for the parameter is that we DO want to remove the "www" when making a transport-level (TCP or SSL) connection address, but DO NOT want to remove the "www" when constructing a a REST request, to avoid getting a 302 back (as per the comment). I'm not sure there is an "underlying issue" beyond that. As you saw, previously it would always remove the "www" - but that was then resulting in getting 302s back from the Google endpoint at the HTTP level.

I admit this is one change that could have been in a separate PR :-)

Regarding API breaking - splitURL is internal to client.cc - so it shouldn't be API breaking, should it? Or am I missing the point @Tachi107 ?

src/client/client.cc
dgreatwood: I added a comment

src/client/client.cc
dgreatwood: Fair. I do like having it there as a prompt. I added a comment pointing out that you can do `-DPST_SSL_REQ_DBG`

src/client/client.cc
dgreatwood: Oh, why do prefer that way round @Tachi107? Since scheme is a non-const variable (parameter actually), and Address::Scheme::Https is a constant, I would think it best to do `(Address::Scheme::Https == scheme)` to avoid any risk of using a single `=` and so doing an assignment. Though I admit I'm not super consistent about that.

What's the argument for the other way around?

src/client/ssl_async.cc
dgreatwood: Excellent point, both. 

I've commented out the OpenSSL functions in init_openssl_library - they are all deprecated. I haven't deleted it completely, just in case someone wants to see how to do it prior to OpenSSL 1.1.0.

Per the documentation, OpenSSL now (OpenSSL 1.1.0 and later) handles this initialization automatically for itself, unless you need non-default parameters, which we don't, I believe. 

> What if Pistache is embedded into another program, which performs its own OpenSSL initialization?

As you suggested, with the newly modified code, since we don't actually need to call OPENSSL_init_ssl in Pistache, this problem should not arise. Provided a user doesn't themselves call OPENSSL_init_ssl with parameters that are somehow incompatible with Pistache; but hopefully that's a remote risk.

I have included:
```
#ifndef OPENSSL_NO_DEPRECATED
#define OPENSSL_NO_DEPRECATED
#endif
```
in _ssl_async.cc_, since using deprecated OpenSSL functions is often regarded as insecure. I would put `-DOPENSSL_NO_DEPRECATED` in _meson.build_, but right now _listener.cc_ won't compile that way - we should go back and fix that in a separate PR. BTW, I believe OpenSSL currently uses v3.02.00 of the API by default, to determine what is deprecated.

src/client/ssl_async.cc
dgreatwood: `Have you tried compiling it with OPENSSL_NO_DEPRECATED defined
`
I have now (see prior comment)

src/client/ssl_async.cc
dgreatwood: See below

PR #1292, opened 2025-02-18
chore: http_header: add support for ETag header

n2h9:
Add support for the ETag header.

According to [RFC 9110](https://httpwg.org/specs/rfc9110.html#field.etag).
Description in mozilla [docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag).


kiplingw (changes requested):
Thank you for your contribution, @n2h9. The only change I can see needed is the minor version should be bumped too since you are adding to the API without breaking anything retroactively.

dgreatwood (approved):
I'm no _etags_ expert, but looks fine to me

n2h9:
> Thank you for your contribution, @n2h9. The only change I can see needed is the minor version should be bumped too since you are adding to the API without breaking anything retroactively.

Hey @kiplingw hello, thank you for your review :innocent: 
I have updated the minor version in versions.txt

Please let me know if any further changes are needed.

kiplingw (approved):

n2h9:
Quick question: can I ignore abidiff check failure, or should I address it somehow @kiplingw ?

kiplingw:
> Quick question: can I ignore abidiff check failure, or should I address it somehow @kiplingw ?

I think you should be ok. That [warning](https://github.com/pistacheio/pistache/actions/runs/13412322165/job/37483606798?pr=1292) is just stating you added symbols, hence the bump in the minor version. We should only be worried if symbols were removed or changed.

tests/headers_test.cc
Tachi107, line 1: Some tests checking for edge cases are missing, like:

- what happens if I pass an otherwise valid `etagc*` value to the constructor, but quoted?
- what happens if there's a `"` in the `value` of the constructor?
- what happens if there's a `"` in the string parsed by `parse`?
- what happens if there's an ASCII space or a `DEL` in the etag value? (the `etagc` definition does not allow those in an etag value)

tests/headers_test.cc
```diff
@@ -1016,3 +1016,101 @@ TEST(headers_test, cookie_headers_are_case_insensitive)
...
+        try
+        {
+            etag.parse(data);
+        }
+        catch (std::exception& e)
+        {
+            what = e.what();
+        }
+
+        ASSERT_EQ("Invalid ETag format", what);
```
Tachi107, line 1114:
```suggestion
        EXPECT_THAT([&] { etag.parse(data); }, ThrowsMessage<std::runtime_error>("Invalid ETag format"));
```

Tachi107 (changes requested):
Hi, thanks for the patch!

Some feedback below :)

include/pistache/http_header.h
```diff
@@ -551,6 +551,38 @@ namespace Pistache::Http::Header
...
+
+        explicit ETag(std::string&& value, bool useWeakValidator = false)
+            : value_(std::move(value))
+            , useWeakValidator_(useWeakValidator)
+        { }
+
+        explicit ETag(const std::string& value, bool useWeakValidator = false)
+            : value_(value)
+            , useWeakValidator_(useWeakValidator)
+        { }
```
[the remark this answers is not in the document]
n2h9: Hey @Tachi107 hello :innocent: thank you for your review :bowing_man:  :slightly_smiling_face: 

- two constructors to one with string_view -> done :heavy_check_mark: 
- add couple comments, in Doxygen format (or close to it) :slightly_smiling_face:  -> done :heavy_check_mark:

tests/headers_test.cc
n2h9: Hey @Tachi107 hello :innocent: thank you for your review :bowing_man: 

That is actually a very useful points you mentioned, thank you!  

I have added a validation that etagc is from the range as in RFC: 
```sh
etagc      = %x21 / %x23-7E / obs-text
```
where 
```sh
obs-text       = %x80-FF
```

and with this addressing corner cases:
>what happens if I pass an otherwise valid etagc* value to the constructor, but quoted?

so, this is not allowed as etagc* can not contains `"` asci char -> throw exception -> added test case for this;

>what happens if there's a " in the value of the constructor?

I would say it is the same case as above :slightly_smiling_face: 

>what happens if there's a " in the string parsed by parse

throw exception, not allowed, added test case

> what happens if there's an ASCII space or a DEL in the etag value? (the etagc definition does not allow those in an etag value)

both not allowed -> throw exception -> added test cases; 

---

Also I updated the name `value`_ to `etagc_` and `useWeakValidator_` to `isWeak_` to be more close to RFC terminology. Though I left  useWeakValidator in constructir intentionally, I think it makes more clear the meaning of constructor param.  

---

And I made `validateEtagc` function public, so that clients could use it to validate value if they need, but validateEtagcWithException made private as it only makes sense within the class. 

---

please let me know if you have any further suggestions :innocent:

Tachi107:
Thanks for your fixes! Merging.

PR #1295, opened 2025-02-26
Fix: Give write access to Http::Message::body

sjoubert:
I don't know if allowing write access to the body of the message makes sense in the overall design but:
* My initial need is to be able to modify the body of an input `Http::Request` in a middleware hook (specifically converting/replacing YAML to JSON in a single location and only have to deal with JSON afterwards)
* This makes things consistent with `cookies()` and `headers()` that already provide both read and write access
* In any case, with the presence of a const method returning a const reference, I don't think such a method (as is) is useful, so I guess it should at least be removed if you decide not to allow write access.


kiplingw:
Hey @sjoubert. Thank you for your PR.

I'm personally wary of this design change. I think if you want YAML or JSON, then that's something the client should know to send in advance. You can check the incoming headers and verify it's one or the other and issue a response accordingly if you want to be able to handle both. But I don't think allowing the server to alter the message body makes logical sense.

Happy to hear from my colleagues.

Tachi107:
Yeah, this looks like a small change with big potential implications. I'd prefer to keep the body const too.

sjoubert:
> I'm personally wary of this design change. I think if you want YAML or JSON, then that's something the client should know to send in advance. You can check the incoming headers and verify it's one or the other and issue a response accordingly if you want to be able to handle both. But I don't think allowing the server to alter the message body makes logical sense.

I was indeed planning on documenting this in my api, and doing this based on client request and header value yes.

I'm using pistache through the openapi generator. What I get when implementing the generated stubs is the C++ data/model classes already parsed from JSON not the raw body. But the generator only supports JSON. Before digging into adding full support for YAML in the generator I tried the approach of injecting a middleware (which I can do even with the openapi generator output) to convert the input to JSON (based on client header request) before it hits the JSON parser. I tested locally with with the patch, it works pretty well.

But on first glance I agree that being able to modify the message body seems weird. I  guess I'll have to dig into the openapi generator then.

Do you want me to update the PR to remove the non-const (useless?) method instead?

kiplingw:
> Do you want me to update the PR to remove the non-const (useless?) method instead?

Well if you remove the non-const return by reference for the string body method, that's the only substantive change you made. I think that we can't merge this as is for the reasons already discussed. 

But @dennisjenkins75 did raise a good point that the `Request` object probably should be passed to the application as a const. I'd suggest opening a new PR, if you like, for that. It will break a lot of applications, but the fix is pretty trivial and might make the library safer. Although I hesitate at first because @oktal probably had a good reason for why it's not const.

sjoubert:
Since you closed the PR, that's fine with me as it is. Thanks for looking at it anyway 🙂

Issue #1290, opened 2025-02-14
Race condition when using ResponseStream::flush() from another thread


bkueng:
I have a use-case where I call `ResponseStream::flush()` from another thread for an ongoing connection.

In rare cases ThreadSanitizer raises an issue in the form of:
```
  Read of size 8 at 0x722000004188 by main thread (mutexes: write M0):
    #0 std::__shared_count<(__gnu_cxx::_Lock_policy)2>::_M_swap(std::__shared_count<(__gnu_cxx::_Lock_policy)2>&) /usr/include/c++/14/bits/shared_ptr_base.h:1097 (param-server+0x46a270) (BuildId: 1b65773e8330da60c73033ad3abacf9d10ca9aa8)
    #1 std::__shared_ptr<Pistache::Async::Private::Core, (__gnu_cxx::_Lock_policy)2>::__shared_ptr(std::__shared_ptr<Pistache::Async::Private::Core, (__gnu_cxx::_Lock_policy)2>&&) <null> (param-server+0x777a90) (BuildId: 1b65773e8330da60c73033ad3abacf9d10ca9aa8)
   ...

  Previous write of size 8 at 0x722000004188 by thread T3 (mutexes: write M1, write M2, write M3):
    #0 operator new(unsigned long) <null> (libtsan.so.2+0x66286) (BuildId: c08afb1c60772d9b4e4d4be38d0c0434c5b41990)
    #1 void Pistache::Queue<Pistache::Tcp::Transport::WriteEntry>::push<Pistache::Tcp::Transport::WriteEntry>(Pistache::Tcp::Transport::WriteEntry&&) <null> (param-server+0x77d12e) (BuildId: 1b65773e8330da60c73033ad3abacf9d10ca9aa8)
    ...

  Location is heap block of size 120 at 0x722000004180 allocated by thread T3:
    #0 operator new(unsigned long) <null> (libtsan.so.2+0x66286) (BuildId: c08afb1c60772d9b4e4d4be38d0c0434c5b41990)
    #1 void Pistache::Queue<Pistache::Tcp::Transport::WriteEntry>::push<Pistache::Tcp::Transport::WriteEntry>(Pistache::Tcp::Transport::WriteEntry&&) <null> (param-server+0x77d12e) (BuildId: 1b65773e8330da60c73033ad3abacf9d10ca9aa8)
   ...

  Mutex M0 (0x7220000001d8) created at:
 [ mutex in my application ]

[...]

PR #1299, opened 2025-03-09
Fix: writesQueue Problem in ResponseStream::flush - Issue #1290

dgreatwood:
http.cc: Remove the call to transport_->flush in ResponseStream::flush. The call was creating corruption in writesQueue as per Issue #1290.

Calling transport_->flush from ResponseStream::flush is unnecessary - ResponseStream::flush already places the write on the transport's writesQueue with a call to asyncWrite; the transport will send the write just as soon as the fd becomes writable through Pistache's normal ePoll system. In other words, it is sufficient for ResponseStream::flush to move the send from out of the ResponseStream's buffer and send it to the Transport.

Calling transport_->flush is also dangerous - writesQueue is supposed to be single consumer queue, but calling flush from ResponseStream::flush initiates a writesQueue pop from a different thread (i.e. from the thread executing ResponseStream::flush). This can cause queue corruption if both the ResponseStream::flush thread and the Pistache consumer thread are popping at the same moment - exactly what happens in Issue #1290.


kiplingw:
@bkueng, can you let us know if this solved your problem?

kiplingw (approved):

bkueng:
Thanks for the fix. I tested this and the flush still works, i.e. the client immediately receives the data.

Oddly I still see a warning from the thread sanitizer. It seems to be a false positive though.
The reason that makes me think so is that before I saw memory corruptions when running on an ARM target, as a result of the race condition: it usually took about 5-10min to trigger, whereas with this change here there is no issue after some hours of runtime with continuous requests.

dgreatwood:
@bkueng 

Thanks for the confirmation.

Regarding the remaining tsan warning, does that happen in the Pistache code? If it does, could you post the details?

Thanks again!

bkueng:
Yes it's in pistache.
```
WARNING: ThreadSanitizer: data race (pid=28940)
  Read of size 8 at 0x722000001a08 by thread T3 (mutexes: write M0, write M1):
    #0 std::__shared_count<(__gnu_cxx::_Lock_policy)2>::_M_swap(std::__shared_count<(__gnu_cxx::_Lock_policy)2>&) /usr/include/c++/14/bits/shared_ptr_base.h:1097 (param-server+0x46a2de) (BuildId: 3ee5af31e6ff4d7216ba60b3b9bb47a45a9ab738)
    #1 std::__shared_ptr<Pistache::Async::Private::Core, (__gnu_cxx::_Lock_policy)2>::__shared_ptr(std::__shared_ptr<Pistache::Async::Private::Core, (__gnu_cxx::_Lock_policy)2>&&) <null> (param-server+0x777bde) (BuildId: 3ee5af31e6ff4d7216ba60b3b9bb47a45a9ab738)

  Previous write of size 8 at 0x722000001a08 by main thread (mutexes: write M2):
    #0 operator new(unsigned long) <null> (libtsan.so.2+0x66286) (BuildId: a4fcf6899c58ea19315282d1878579e13b50c9f5)
    #1 void Pistache::Queue<Pistache::Tcp::Transport::WriteEntry>::push<Pistache::Tcp::Transport::WriteEntry>(Pistache::Tcp::Transport::WriteEntry&&) <null> (param-server+0x77d282) (BuildId: 3ee5af31e6ff4d7216ba60b3b9bb47a45a9ab738)
    [...]

  Location is heap block of size 120 at 0x722000001a00 allocated by main thread:
    #0 operator new(unsigned long) <null> (libtsan.so.2+0x66286) (BuildId: a4fcf6899c58ea19315282d1878579e13b50c9f5)
    #1 void Pistache::Queue<Pistache::Tcp::Transport::WriteEntry>::push<Pistache::Tcp::Transport::WriteEntry>(Pistache::Tcp::Transport::WriteEntry&&) <null> (param-server+0x77d282) (BuildId: 3ee5af31e6ff4d7216ba60b3b9bb47a45a9ab738)
    [...]

  Mutex M0 (0x728400004c38) created at:
    #0 pthread_mutex_lock <null> (libtsan.so.2+0x1f60e) (BuildId: a4fcf6899c58ea19315282d1878579e13b50c9f5)
    #1 __gthread_mutex_lock /usr/include/c++/14/x86_64-redhat-linux/bits/gthr-default.h:762 (param-server+0x50e64b) (BuildId: 3ee5af31e6ff4d7216ba60b3b9bb47a45a9ab738)
    #2 std::mutex::lock() /usr/include/c++/14/bits/std_mutex.h:113 (param-server+0x50e990) (BuildId: 3ee5af31e6ff4d7216ba60b3b9bb47a45a9ab738)
    #3 std::lock_guard<std::mutex>::lock_guard(std::mutex&) /usr/include/c++/14/bits/std_mutex.h:250 (param-server+0x5134bc) (BuildId: 3ee5af31e6ff4d7216ba60b3b9bb47a45a9ab738)
    #4 Pistache::Aio::SyncImpl::addHandler(std::shared_ptr<Pistache::Aio::Handler> const&, bool) <null> (param-server+0x80dc71) (BuildId: 3ee5af31e6ff4d7216ba60b3b9bb47a45a9ab738)
    [...]

  Mutex M1 (0x728400004c08) created at:
    #0 pthread_mutex_lock <null> (libtsan.so.2+0x1f60e) (BuildId: a4fcf6899c58ea19315282d1878579e13b50c9f5)
    #1 __gthread_mutex_lock /usr/include/c++/14/x86_64-redhat-linux/bits/gthr-default.h:762 (param-server+0x50e64b) (BuildId: 3ee5af31e6ff4d7216ba60b3b9bb47a45a9ab738)
    #2 std::mutex::lock() /usr/include/c++/14/bits/std_mutex.h:113 (param-server+0x50e990) (BuildId: 3ee5af31e6ff4d7216ba60b3b9bb47a45a9ab738)
    #3 std::lock_guard<std::mutex>::lock_guard(std::mutex&) /usr/include/c++/14/bits/std_mutex.h:250 (param-server+0x5134bc) (BuildId: 3ee5af31e6ff4d7216ba60b3b9bb47a45a9ab738)
    #4 Pistache::Aio::SyncImpl::runOnce() <null> (param-server+0x80e1ea) (BuildId: 3ee5af31e6ff4d7216ba60b3b9bb47a45a9ab738)

  Mutex M2 (0x7220000001d8) created at:
    [application mutex]

  Thread T3 (tid=28944, running) created by thread T2 at:
    #0 pthread_create <null> (libtsan.so.2+0x205e6) (BuildId: a4fcf6899c58ea19315282d1878579e13b50c9f5)
    #1 std::thread::_M_start_thread(std::unique_ptr<std::thread::_State, std::default_delete<std::thread::_State> >, void (*)()) <null> (libstdc++.so.6+0x4b5f8) (BuildId: 2433a579334ce4caa8fb2d2fee7b787288792ae0)

SUMMARY: ThreadSanitizer: data race /usr/include/c++/14/bits/shared_ptr_base.h:1097 in std::__shared_count<(__gnu_cxx::_Lock_policy)2>::_M_swap(std::__shared_count<(__gnu_cxx::_Lock_policy)2>&)
```
With the full stack trace:
```
__tsan::OnReport(const __tsan::ReportDesc *, bool) 0x00007ffff6e7a310
<unknown> 0x00007ffff6e7d6b6
<unknown> 0x00007ffff6e7e9a7
std::__shared_count::_M_swap shared_ptr_base.h:1097
std::__shared_ptr::__shared_ptr(std::__shared_ptr<…> &&) 0x0000000000777bdf
std::shared_ptr::shared_ptr(std::shared_ptr<…> &&) 0x000000000077468c
Pistache::Async::Resolver::Resolver(Pistache::Async::Resolver &&) 0x00000000007719f5
Pistache::Async::Deferred::Deferred(Pistache::Async::Deferred<…> &&) 0x0000000000772967
Pistache::Tcp::Transport::WriteEntry::WriteEntry(Pistache::Tcp::Transport::WriteEntry &&) 0x000000000077f2dd
Pistache::Queue::pop() 0x00000000007c75c2
Pistache::PollableQueue::pop() 0x00000000007d1584
Pistache::Queue::popSafe() 0x00000000007c523d
Pistache::Tcp::Transport::handleWriteQueue(bool) 0x00000000007c174f
Pistache::Tcp::Transport::onReady(const Pistache::Aio::FdSet &) 0x00000000007c0178
Pistache::Http::TransportImpl::onReady(const Pistache::Aio::FdSet &) 0x00000000007d1eba
Pistache::Aio::SyncImpl::handleFds(std::vector<…>) const 0x000000000080e692
Pistache::Aio::SyncImpl::runOnce() 0x000000000080e23f
Pistache::Aio::SyncImpl::run() 0x000000000080e36f
<lambda#1>::operator()() const 0x000000000080fa0b
std::__invoke_impl<…>(std::__invoke_other, <lambda#1> &&) 0x0000000000815880
std::__invoke<…>(<lambda#1> &&) 0x0000000000815843
std::thread::_Invoker::_M_invoke<…>(std::_Index_tuple<…>) 0x00000000008157f0
std::thread::_Invoker::operator()() 0x00000000008157c4
std::thread::_State_impl::_M_run() 0x00000000008157a8
<unknown> 0x00007ffff6a4b524
<unknown> 0x00007ffff6e15c7a
start_thread 0x00007ffff6798168
__clone3 0x00007ffff681c14c
```

It again points to the `Queue`, which uses atomics for synchronization, and from what I can tell I did not see any problem there.

dgreatwood:
Second question - are you building Pistache with sanitize defined (via `meson -Db_sanitize=<xxx>`) - or is sanitize defined only on your own code?

If that question makes sense given what you're doing?

dgreatwood:
> ...indeed the flag was not passed along to the pistache lib. I added it and with that there is no more warning on this PR.

Great. 

I believe that tsan can be prone to false positives within libraries that are not compiled with sanitize enabled, which would be consistent with your suspicion that what you were seeing before was a false positive. For future reference if needed, I've added a few more notes below, which also seem to point to it being a false positive previously.

The stack traces on master (without the PR) look consistent with the prior analysis too.

I think **we can declare this one done**. Thanks much for your inputs and follow up.

**Notes on Prior tsan [Warning](https://github.com/pistacheio/pistache/pull/1299#issuecomment-2713120748)** 
NB: Per @bkueng, this warning showed up only when Pistache was compiled _without_ sanitize enabled, which may suggest a false positive.

Tsan complains about a data race between a read in thread T3 and a write in main thread. The read (T3) is in a `shared_ptr` constructor, while the write is in a call to `new`.

The read (T3) is in the `Async::Resolver` constructor, while setting the `Resolver` instance's `std::shared_ptr<Private::Core> core_`. This could be a call to the `Resolver` default empty constructor, but more likely I think is that it's the `Resolver` assignment operator.

That assignment is triggered from `Transport::handleWriteQueue` which calls `writesQueue.popSafe()` to get a `unique_ptr` to a `WriteEntry`. `popSafe` pops an entry off the `writesQueue`, and manufacturers a new `WriteEntry`, the variable `object`, which it returns, using `std::move`. 

The `WriteEntry` constructor is thus triggered by `popSafe`, and it in turn calls the `Async::Deferred` constructor, which calls the `Async::Resolver` constructor, which calls the `shared_ptr<Private::Core>` constructor as described above.

Meanwhile, per tsan, what appears to be the shared counter for `Resolver core_` ptr was written in thread main. `Transport::asyncWrite` indeed calls `writesQueue.push(std::move(write))` in exactly this way.

However, this doesn't appear to be a real data race. The writesQueue is protected using atomic pointers (see _mailbox.h_), so thread T3 cannot pop the `writeEntry` (including `shared_ptr core_`) from the queue until the `writeEntry` has been fully constructed and pushed on the `writesQueue` by thread main.

PR #1305, opened 2025-04-12
Fix: Date header parsing when date is an epoch

vaxerski:
The Date header sometimes is an epoch (e.g. git)

Handle such cases instead of dropping the request.

Tested locally, works fine.

### DIFF
--- a/src/common/http_defs.cc
+++ b/src/common/http_defs.cc
@@ -62,6 +62,26 @@ namespace Pistache::Http
             return !in.fail();
         }
 
+        bool parse_epoch(const std::string& s, time_point& tp)
+        {
+            for (unsigned int i = 0; i < s.size(); ++i)
+            {
+                if (!std::isdigit(s[i]))
+                    return false;
+            }
+
+            try
+            {
+                tp = time_point(std::chrono::seconds(std::stoull(s)));
+            }
+            catch (std::out_of_range& e)
+            {
+                return false;
+            }
+
+            return true;
+        }
+
     } // anonymous namespace
 
     CacheDirective::CacheDirective(Directive directive)
@@ -126,6 +146,8 @@ namespace Pistache::Http
         else if (parse_RFC_850(str, tp))
             return FullDate(tp);
         else if (parse_asctime(str, tp))
+            return FullDate(tp);
+        else if (parse_epoch(str, tp))
             return FullDate(tp);
 
         PS_LOG_DEBUG_ARGS("Failed parsing date: %s", str.c_str());


src/common/http_defs.cc
```diff
@@ -62,6 +62,17 @@ namespace Pistache::Http
...
         }
 
+        bool parse_epoch(const std::string& s, time_point& tp)
+        {
+            for (unsigned int i = 0; i < s.size(); ++i)
+            {
+                if (!std::isdigit(s[i]))
+                    return false;
+            }
+            tp = time_point(std::chrono::seconds(std::stoull(s)));
```
kiplingw, line 72: Is this sufficient to validate the input string? Can `time_point` signal to the caller in any way an error?

dgreatwood:
Thanks @vaxerski 

If you're aware of a well known public website that has an epoch date header, could you add a test for it? If you look in _tests/https_client_test_net.cc_, and treat the test `one_client_with_google_request` as your model, it will hopefully be straightforward.

Regarding the change, this is not my area of expertise at all, and it's a small point, but I wondered if 
`tp = time_point(std::chrono::seconds(std::stoull(s)));`
might have an issue with leap seconds? Might
`tp = time_point(std::chrono::system_clock::from_time_t(static_cast<std::time_t>(std::stoull(s)))));`
be better?
@Tachi107, what do you think?

src/common/http_defs.cc
dgreatwood: Oh yeah.

Well, `std::stoull` can have exceptions `std::invalid_argument` or `std::out_of_range.` I guess we should trap an exception, set `errno`, and return `false`. 

I don't see exceptions listed for `duration` or `time_point`, but maybe they could (e.g. overflow?). Probably safest would be to _try/catch_ round that line, set `errno` on `out_of_range (ERANGE)` or `EINVALID` on any other exception, and return false in the exception case.

vaxerski:
> If you're aware of a well known public website that has an epoch date header, could you add a test for it?

I am not. Never seen it myself, but I am building a reverse proxy atm with pistache and spotted it in `git`'s requests when cloning via http(s)

About leap seconds, no clue, tbh.

src/common/http_defs.cc
vaxerski: true, we could try catch there. `invalid_argument` won't be thrown as it is all digits, but out of range is a fair possibility.

vaxerski: I've added the try catch.

dgreatwood:
Merged. Thanks for the contribution @vaxerski

kiplingw:
Good PR, but we forgot to bump the patch version.

PR #1307, opened 2025-04-13
Fix: Don't automatically manage headers in doRequest

vaxerski:
This cumbers the availability of the API and leads to bad requests that should not be bad.

I am building a reverse-proxy with pistache, and the client API cumbers my ability to do so properly:
- The client API _always_ drops the UA and replaces it with pistache's.
- The client API _always_ adds the Host header and this leads to bad requests when we want to set the host header manually.

In this MR, I drop both. Instead, pistache's UA and Host are added iff they are not added by the client.


kiplingw:
@vaxerski, please don't forget to bump the patch version per our documentation since you've altered the underlying implementation.

vaxerski:
uhh... where is that documentation? I've checked the website and the repo and can't find anything related to contributing guidelines.

kiplingw:
See [here](https://github.com/pistacheio/pistache?tab=readme-ov-file#versioning).

vaxerski:
ah, my bad, I got a zoomer brain. Thanks!

vaxerski:
done, I've also reworded the commit, I think the linter should be happy now.

PR #1311, opened 2025-04-29
Make rapidjson dependency optional for CMake

JnCrMx:
This has the same reasoning as #1145 and will help people who use this project as a dependency in a CMake-based project (for example, with FetchContent).

Furthermore, "compatibility with CMake < 3.5 has been removed from CMake" (at least in version 3.28.3), which prevents this library from being used at all with CMake.


JnCrMx:
oopsie, this is not supposed to be on this branch :sweat_smile:

Finkman:
Is this merged or WIP somewhere, as RPID-Json with the out-dated cmake req is making me getting problems.

PR #1313, opened 2025-05-06
Attempt to expand listener to multi-threaded

### LINKED ISSUE #1312: Limited Performance on HTTPS server
Hello, team,

We use Pistache to build an API server in production, one of the core function of the component is to provide a simple kv store to the consumers, over this API server.

In the performance evaluation, we found the TPS (transaction per second) were capped to ~400, increasing the api server threads doesn't help, after profiling, we noticed one thread is significantly busier, while the rest threads are almost idle. Further investigation shows that thread spends most of the time doing crypto related calculation.

After looking at the code, I think the problem is in the Listener::run() method. This thread handles all tcp connection request and accepts them. If SSL is on, the SSL_accept() is a CPU heavy operation, and there is only one thread to do this job. In the performance test, when the number of request increases, this thread is saturated.

Below is the performance comparison with SSL on and off. The server is from the examples/http_server.cc, It just provides a no-op: for any kind request, directly send a 200 response. It has the option to run as plain http server or https server, when running as https server, the cert is a simple self-signed cert.

20 api threads, no ssl:
```
Progress: [================================================= ] 99.2% (297538/300000) | Success: 297538, Failed: 0 | Elapsed: 12.001636138s | Ops/sec: 24791.5
```
20 api threads, with ssl:
```
Progress: [==                                                ] 4.0% (12063/300000) | Success: 0, Failed: 12063 | Elapsed: 12.010370974s | Ops/sec: 1004.4
```

The api_server thread CPU stat, when ssl is on:
```
  PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND
  57959 jiankyu   20   0 1489760  10884   9056 S  88.7   0.0   0:23.94 runhttp_server
  57960 jiankyu   20   0 1489760  10884   9056 S   0.0   0.0   0:00.00 runhttp_server
  57961 jiankyu   20   0 1489760  10884   9056 S   0.0   0.0   0:00.00 runhttp_server
  57962 jiankyu   20   0 1489760  10884   9056 S   0.0   0.0   0:00.00 runhttp_server
  57963 jiankyu   20   0 1489760  10884   9056 S   0.0   0.0   0:00.00 runhttp_server
  57964 jiankyu   20   0 1489760  10884   9056 S   0.0   0.0   0:00.00 runhttp_server
  57965 jiankyu   20   0 1489760  10884   9056 S   0.0   0.0   0:00.00 runhttp_server
  57966 jiankyu   20   0 1489760  10884   9056 S   0.0   0.0   0:00.00 runhttp_server
```
The first thread handles all ssl handshakes.
If the cert involves more complicated cryptographic algorithm, or if it runs on a less powerful CPU, this 1004.4 ops would further drop down.

Here I would like the core developer team to make an enhancement on the Listener class, to make it multi-threaded. The most favorable solution is to throw the accept() into each api_thread, so that we simple increase the performance by increasing the api thread number. If this is too tough, probably we can add a new control option, the "listener_thread_num", to explicitly specify how many threads should work on the accept() part.

jiankyu:
Attempt to resolve #1312. Changes summary can be found in commit message.
This is still working in progress, stability issue present. Bring in early to check if any failed CI job can shed a light.


include/pistache/config.h
```diff
@@ -18,6 +18,7 @@ namespace Pistache::Const
     static constexpr size_t MaxBacklog     = 128;
     static constexpr size_t MaxEvents      = 1024;
     static constexpr size_t MaxBuffer      = 4096;
+    static constexpr size_t DefaultAcceptors = 1;
```
kiplingw, line 21: Please keep alignment of value assignment consistent.

version.txt
```diff
@@ -1 +1 @@
-0.5.6.20250328
+0.5.6.20250504
```
kiplingw, line 1: Don't forget to bump the major version if it's not backwards compatible, the minor version if it only added new interfaces, and the patch if only modifying the implementation. See our documentation [here](https://github.com/pistacheio/pistache?tab=readme-ov-file#versioning).

include/pistache/config.h
jiankyu: Ack

version.txt
jiankyu: This patch doesn't change any public interface, so existing users won't break. Also, the default value (DefaultAcceptors = 1) offers the same performance on existing user code.

jiankyu: Ack. I reviewed the doc, this PR adds new parameters to the endpoint options, yes, I should bump the patch version.

kiplingw: If you're changing the public interface then it's no longer backwards ABI compatible. Please read our documentation on [versioning](https://github.com/pistacheio/pistache?tab=readme-ov-file#versioning), otherwise you will break production binaries.

kiplingw: That is true, unless `Listener::init` is public because you've changed its signature. I don't remember how this class / method is used, but if it's not a public interface then you're right you should only bump the minor number from 5 to 6.

kiplingw: Thanks @jiankyu, but you bumped the patch rather than the minor. Please read our versioning documentation.

jiankyu: Really sorry about that.
Just bumped the version to 0.6.0

dgreatwood:
Hi @jiankyu -

Couple of things.

1/ Windows compile error

The build sees error:
`../src/server/listener.cc(589): error C2065: 'lLoggedSetThreadDescriptionFail': undeclared identifier`

This is in the code to set the "thread description" in the Windows case. I think you copied this over from _reactor.cc_? Which is fine. But you would also need to copy over (i.e. add to _listener.cc_) the declaration:
`static std::atomic_bool lLoggedSetThreadDescriptionFail = false;
`

2/ Test case

I take it from the prior conversation that you are seeing the SIGPIPE failure from your own application. It seems the SIGPIPE is not demonstrated by the existing Pistache unit tests? Presuming not, could you add a test that activates the problem?
Most likely you would add the test in _https_server_test.cc_, though you could also inspect _https_client_test_net.cc_ for possibly relevant example tests.

Thanks in advance.

kiplingw:
@jiankyu, I don't know if it's possible through a unit test, but it would be helpful if you are able to benchmark before and after your (finalized) PR when ready. That will help us evaluate whether the underlying resource bottleneck problem has been resolved.

jiankyu:
> Hi @jiankyu -
> 
> Couple of things.
> 
> 1/ Windows compile error
> 
> The build sees error: `../src/server/listener.cc(589): error C2065: 'lLoggedSetThreadDescriptionFail': undeclared identifier`
> 
> This is in the code to set the "thread description" in the Windows case. I think you copied this over from _reactor.cc_? Which is fine. But you would also need to copy over (i.e. add to _listener.cc_) the declaration: `static std::atomic_bool lLoggedSetThreadDescriptionFail = false; `
> 
> 2/ Test case
> 
> I take it from the prior conversation that you are seeing the SIGPIPE failure from your own application. It seems the SIGPIPE is not demonstrated by the existing Pistache unit tests? Presuming not, could you add a test that activates the problem? Most likely you would add the test in _https_server_test.cc_, though you could also inspect _https_client_test_net.cc_ for possibly relevant example tests.
> 
> Thanks in advance.

@dgreatwood 
1. Yes, I copied the code from reactor.cc, I will also copy the declaration to resolve the compile error.
2. Sure, I will add the test case to reveal the problem and show the performance data I got.

jiankyu:
> @jiankyu, I don't know if it's possible through a unit test, but it would be helpful if you are able to benchmark before and after your (finalized) PR when ready. That will help us evaluate whether the underlying resource bottleneck problem has been resolved.

I actually did the benchmarking. Below is the test data. The numbers are generated by a multi-threaded client request generator.
Column 1 is the number of acceptWorker threads, column 2 is the average requests per second during the test, column 3-5 are the median, 90th, and 99th percentile request latency. Last column, efficiency, is defined as `ops / num_threads / ops-with-1-thread`, for example, with 2 threads, I got 1979.47 ops, its efficiency is 1979.47 / 2 / 1062.23 = 93.18%.
```
#threads        op/s       P50     P90     P99   efficiency
1            1062.23    37.282  38.688  42.565   100%
2            1979.47    20.087  21.730  23.066   93.18%
3            2850.90    13.996  15.079  16.234   89.46%
4            3820.15    10.361  11.429  13.018   89.90%
5            4397.98     9.035   9.824  10.720   82.81%
6            5085.45     7.764   8.548   9.441   79.79%
7            5874.28     6.742   7.517   8.337   79.00%
8            6431.21     6.110   6.860   7.689   75.68%
9            6939.28     5.640   6.495   7.375   72.59%
10           7561.68     5.156   5.962   6.860   71.19%
15           9806.96     3.935   4.874   5.794   61.55%
20          12185.99     3.119   3.995   5.292   57.36%
25          12980.32     2.984   3.786   5.209   48.87%
30          14252.14     2.698   3.617   4.996   44.72%
35          14826.37     2.481   3.588   5.674   39.88%
40          14826.68     2.477   3.635   5.714   34.90%
45*         15187.71     2.710   4.101   6.111   31.77%
50*         15304.74     3.016   4.558   6.615   28.81%
60*         15156.50     3.672   5.490   7.697   23.78%
```
As we can see, the maximum ops we can get is 15304.74 at 50 threads, further increasing the threads doesn't increase the ops.
Also, as we increase the threads, the efficiency drops, the performance gain doesn't go linearly with the threads. This makes me think of a different approach to resolve the problem - maybe we can offload the ssl handshake to the `AsyncImpl::Worker` thread, listener is still single threaded, but it only does `accept()`. This requires change in the `Peer` class, and listener has to provide a read-only inteface to expose its ssl context to the workers so they can create ssl object later. I didn't take this approach because it involves more change across different classes, I'm not quite confident about it, but theoretically this approach should have better performance.

dgreatwood:
Cool. Out of interest, what kind of CPU was it, with how many parallel execution units?

Also, as per @kiplingw's prior comment, it would be awesome to put some form of this - e.g. 1, 2, 25, 50, 75 threads for a few seconds each - into a unit test if practical.

jiankyu:
```
$ lscpu
Architecture:            x86_64
  CPU op-mode(s):        32-bit, 64-bit
  Address sizes:         46 bits physical, 48 bits virtual
  Byte Order:            Little Endian
CPU(s):                  104
  On-line CPU(s) list:   0-103
Vendor ID:               GenuineIntel
  Model name:            Intel(R) Xeon(R) Gold 6230R CPU @ 2.10GHz
    CPU family:          6
    Model:               85
    Thread(s) per core:  2
    Core(s) per socket:  26
    Socket(s):           2
    Stepping:            7
    CPU(s) scaling MHz:  49%
    CPU max MHz:         4000.0000
    CPU min MHz:         1000.0000
    BogoMIPS:            4200.00
```

jiankyu:
Unfortunately the current Http::Experimental::Client() can't be used to benchmark this. There are couple of problems with it.
The blocking issue is, it reuses connection and there is no way to opt out, I use this client to send out 10k requests, there are only 1 connection establishment captured.
Secondly, the implementation seems also buggy, it allows setting the "threads" used by underlying reactor, but only the first thread is doing things, when it gets 100% CPU, it can do around 2400 requests per second, while all the other threads are idle.

The Client performance issue can be resolved by multiple clients making requests simultaneously, but I got stuck at the connection reuse problem, I tried different fixes but none of them works so far.

Issue #1316, opened 2025-05-18
Use uninitialized memory while parsing Date header


tyler92:
This issue was found by memory sanitizer (MSAN) enabled for the [existing fuzzer](https://github.com/pistacheio/pistache/blob/master/tests/fuzzers/fuzz_parser.cpp). The issue is in hinnant-date subproject and it was [fixed](https://github.com/HowardHinnant/date/commit/28b5106d4ca911d1e6682c83b03058ae68934eae) there some time ago.

MSAN report:

```
==14==WARNING: MemorySanitizer: use-of-uninitialized-value
    #0 0x563853cf29c5 in floor<std::__1::chrono::duration<long long, std::__1::ratio<1L, 1000000L> >, long double, std::__1::ratio<1L, 1L>, 0> /usr/local/bin/../include/c++/v1/__chrono/duration.h:130:7
    #1 0x563853cf29c5 in round<std::__1::chrono::duration<long long, std::__1::ratio<1L, 1000000L> >, long double, std::__1::ratio<1L, 1L>, 0> /usr/local/bin/../include/c++/v1/__chrono/duration.h:145:25
    #2 0x563853cf29c5 in round_i<std::__1::chrono::duration<long long, std::__1::ratio<1L, 1000000L> >, long double, std::__1::ratio<1L, 1L> > /src/pistache/src/../subprojects/hinnant-date/include/date/date.h:1355:12
    #3 0x563853cf29c5 in std::__1::basic_istream<char, std::__1::char_traits<char>>& date::from_stream<char, std::__1::char_traits<char>, std::__1::chrono::duration<long long, std::__1::ratio<1l, 1000000l>>, std::__1::allocator<char>>(std::__1::basic_istream<char, std::__1::char_traits<char>>&, char const*, date::fields<std::__1::chrono::duration<long long, std::__1::ratio<1l, 1000000l>>>&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>*, std::__1::chrono::duration<long, std::__1::ratio<60l, 1l>>*) /src/pistache/src/../subprojects/hinnant-date/include/date/date.h:7278:40
[...]

PR #1317, opened 2025-05-18
Fix: Use uninitialized memory while parsing Date header

tyler92:
Fixes https://github.com/pistacheio/pistache/issues/1316

I update the hinnant-date subproject up to version [v3.0.3](https://github.com/HowardHinnant/date/releases/tag/v3.0.3) without changes in the copyrights section. The main fix for https://github.com/pistacheio/pistache/issues/1316 is `long double S{};` instead of `long double S;`. Probably it's worth to make this project as a submodule, but I think it's out of the scope for this PR.


kiplingw:
@tyler92, I think it's a useful PR. But like @Tachi107 said, we think it might be in the wrong place. I'd encourage you to submit it [upstream](https://github.com/HowardHinnant/date).

tyler92:
@kiplingw The fix is already in that repo: https://github.com/HowardHinnant/date/pull/698

I didn't make any manual changes, only updated code from that repository

kiplingw:
Yes, looks like it's merged already upstream. We could either create an updated package for the library in our [Pistache Build Dependencies](https://launchpad.net/~pistache+team/+archive/ubuntu/pistache-deps/+packages) PPA until we have an updated system package available, or we could just wait for upstream to sync with Debian and its derivatives (looks like @Tachi107 may have already [taken care of this](https://tracker.debian.org/news/1637450/accepted-howardhinnant-date-303ds-2-source-into-unstable/)). The former is an immediate solution, but needs some work to create the package. The latter is not immediate, but a long term solution with less work on our end.

tyler92:
OK, thanks for your answers! I'm not very experienced with creating packages, so I’ll leave that part - and the decision about Debian - to whoever feels comfortable taking it on.

As for this PR, feel free to close or merge it, depending on what you decide

Tachi107:
On Sun May 18, 2025 at 7:11 PM CEST, Kip wrote:
> But like @Tachi107 said, we think it might be in the wrong place. I'd 
> encourage you to submit it 
> [upstream](https://github.com/HowardHinnant/date).

To be clear, this is not exactly what I said. I'm fine with this PR, as 
it just updates our bundled date library to the latest upstream version. 
What I was suggesting is to unbundle the header and use a meson wrap 
file instead. But this is a suggestion for the future.

Merging :)

PR #1344, opened 2026-04-12
CI: Add FreeBSD Workflow

dgreatwood:
Also other fixes needed for FreebSD workflow:
 - Use Pisatche's custom strerror_r for all BSD, not just NetBSD
 - Cope with socket connect returning 0 on non-blocking SSL socket
 - Use EAGAIN as alternate to ENODATA for SSL read on BSD


kiplingw (approved):

kiplingw (changes requested):
Looks like one CI failure on macOS and many on Windows. On the latter, it looks like minor syntax issues.

dgreatwood:
Fixed the syntax issue (missing semicolon) that was causing the Windows build issues.

For the MacOS build issues, it appears to be a change in their MacOS-26-on-Intel image, unrelated to this PR. I'll leave it to see if GitHub fixes it, but if it persists (which would suggest that it really is an issue in Pistache) I can investigate it separately.

PR #1346, opened 2026-04-15
bug: Added workflow to run examples on Windows from installed location

maximiliank:
This workflow runs an example against the installed pistache libraries on Windows.
Due to a mismatch in the dll name including soversion and the generated def file (expects pistache.dll) running the example with shared libraries fails.
Static builds fails on Windows due to a duplicate target caused by static_library('pistache') which is used for def file creation.

This workflow demonstrates the issues.


dgreatwood:
Hi @maximiliank.

To confirm, you believe this smoke test library should compile and link?

Could you say more about how you believe Pistache _should_ work in this case? I.e. what change would you suggest.

Thanks.

maximiliank:
> Hi @maximiliank.
> 
> To confirm, you believe this smoke test library should compile and link?
> 
> Could you say more about how you believe Pistache _should_ work in this case? I.e. what change would you suggest.
> 
> Thanks.

Hi @dgreatwood,

I am trying to build a workflow to show the following two errors on Windows:

* when building shared libraries the dll is named pistache-0.5.dll but the created lib/def which is created in a kind of workaround is expecting pistache.dll.
* static building is not working since the static_library("pistache") defined to do the lib/def workaround is then conflicting with the library("pistache") in the meson build file.

Both issues are straight forward to fix in the meson build file, but I would like to showcase this in the CI such that we can check it in the future as well. I will try to get some time to build the failing tests and then a fix.

maximiliank:
The added workflow shows:

* dll loading error at runtime, see https://github.com/pistacheio/pistache/actions/runs/24692233558/job/72216801420
  `Server exited during startup (exit code -1073741515 / 0xC0000135).` 

  The Microsoft error code STATUS_DLL_NOT_FOUND (hexadecimal 0xc0000135) indicates that an application failed to launch because a required Dynamic Link Library (.dll) file could not be located

* static library failed to configure, see https://github.com/pistacheio/pistache/actions/runs/24692233558/job/72216801421
  `src\meson.build:348:18: ERROR: Tried to create target "pistache", but a target of that name already exists.`

maximiliank:
@dgreatwood Do you think those changes have anything todo with the freebsd failures?

dgreatwood:
> Do you think those changes have anything todo with the freebsd failures?

[DG] No, I don't think so.

The FreeBSD failures seem to have things like:
```
Updating FreeBSD-ports repository catalogue...
  pkg: Failed to fetch https://pkg.freebsd.org/FreeBSD:15:aarch64/latest/meta.conf: Timeout was reached
...
pkg: Repository FreeBSD-ports cannot be opened. 'pkg update' required
  pkg: No packages available to install matching 'curl' have been found in the repositories
```
i.e. it looks like the freebsd.org repo was failing to respond.

I did see previously that the NetBSD repo was down for a while, so I added some code to the workflow to try a different repo (mirror) if the main one fails. I guess we could try something similar for FreeBSD if it proves to be a frequent issue.

maximiliank:
> > Do you think those changes have anything todo with the freebsd failures?
> 
> [DG] No, I don't think so.
> 
> The FreeBSD failures seem to have things like:
> 
> ```
> Updating FreeBSD-ports repository catalogue...
>   pkg: Failed to fetch https://pkg.freebsd.org/FreeBSD:15:aarch64/latest/meta.conf: Timeout was reached
> ...
> pkg: Repository FreeBSD-ports cannot be opened. 'pkg update' required
>   pkg: No packages available to install matching 'curl' have been found in the repositories
> ```
> 
> i.e. it looks like the freebsd.org repo was failing to respond.
> 
> I did see previously that the NetBSD repo was down for a while, so I added some code to the workflow to try a different repo (mirror) if the main one fails. I guess we could try something similar for FreeBSD if it proves to be a frequent issue.

@dgreatwood thank you. Can you review this PR?

dgreatwood:
> Can you review this PR?

[DG] It's great.

Could you bump the minor version number in version.txt? There's enough going on here for the version to change.

If you have the bandwidth to look at it, it would be cool to also build and run the examples under gcc / GNU tools on Windows; we try to avoid dependencies on proprietary build tools like MSVC, so far as practical. For the gcc case, windows.yaml has:
```
if ($env:mtrx_compiler -eq "gcc") {
          $env:force_msys_gcc = "Y";
          . winscripts\gccsetup.ps1
        }
```
windows.yaml also preserves certain environment variables between the workflow phases. No doubt there are alternate ways to setup and use gcc, if you prefer.

One final thing, if you don't mind. I saw it uses ilammy/msvc-dev-cmd@v1. Are you comfortable the ilammy is a trustworthy source of scripts?

With much thanks.

dgreatwood:
I'm merging this, it's helpful to include.

@maximiliank, in future, if you do find you have the time and appetite to add the equivalent gcc test, that would be great.

Thanks.

PR #1350, opened 2026-05-22
Download zlib from debian.org if zlib.net fails

dgreatwood:
Download zlib from debian.org if zlib.net download fails

Also, use GitHub doxygen "latest" redirected URL to find doxygen version number when downloading doxygen


dgreatwood:
Also, fixed the unrelated CI issue that the macOS-26-intel image had stopped working.

It had stopped working because we were assuming that libevent would already be present in the github image, and it was not there any longer in the macOS-26-intel image. On all macOS, we now check if libevent is installed, and install if not, using brew.

dgreatwood:
Also, note - the two openbsd workflows that fail are arm-64 (aarch64) images. This seems to be an upstream issue with the openbsd images; they are from `anyvm.org`, `vmactions/openbsd-vm@v1`. They should start working again once upstream is fixed.

PR #1353, opened 2026-09-10
Fix several cross-thread races and reentrant-lock deadlocks in endpoint/reactor/transport

raphaelbergamini:
## Summary

A set of independent, atomic fixes for cross-thread data races and reentrant-lock deadlocks found while auditing the transport/reactor/endpoint/listener/client code paths for thread-safety, plus a merge of `fix/router-tls-http-fixes` (listener fd-leak fixes, a router path-sanitization bug, and SSL read/write error-handling fixes) into this branch, since both branches touched overlapping code.

- **peer**: make `Peer::fd_` a `std::atomic<Fd>` and use `exchange()` in `closeFd()`, since `fd()`/`actualFd()` (readers) and `closeFd()` (writer) can run on different threads concurrently.
- **endpoint**: protect `Endpoint::handler_` with a new `handlerMutex_`, since it's read from the transport factory lambda (possibly on a different thread than the one calling `setHandler()`).
- **endpoint**: reset `timerFd` to `PS_FD_EMPTY` after closing it in `unregisterPoller()` to avoid a stale value being reused/double-closed.
- **eventmeth**: fix a lost-wakeup race in `EmEventCtr::read()`/`writeProt()` - the blocking `wait()` calls used dedicated `cv_read_mutex_`/`cv_write_mutex_` mutexes, separate from `counter_val_mutex_` (the mutex actually held while checking/mutating `counter_val_`), leaving a window where a notification could be missed and the waiter blocked forever. Now `wait()` uses `counter_val_mutex_` itself.
- **reactor**: `Reactor::Impl::detachFromReactor()` (public) performed unregistration work without taking `poller.reg_unreg_mutex_` at all, and `Reactor::Impl::handlers()` iterated `handlers_` without taking it either, racing with concurrent `attachHandler`/`detachFromReactor` calls. Both now lock the mutex; `detachAndRemoveAllHandlers()` (which already holds the lock) now calls a new `detachFromReactorMutexAlreadyLocked()` helper instead of the public, self-locking entry point, avoiding a reentrant-lock deadlock on `std::mutex`. `WorkerReactor::handlers()` now delegates to the underlying sync reactor's own locked `handlers()` instead of bypassing it via `handler()`.
- **transport**: protect the `timers` map with a new dedicated `timersLock` - unlike `toWrite`/`peers_`, `timers` can be touched directly from a thread other than the reactor thread (e.g. via `Http::Timeout::disarm()`). Also drain `peersQueue` unconditionally at the top of `onReady()`, before processing any other fd, closing a window where a peer-registration and a write notification arriving together could be processed in the wrong order and silently drop the write.
- **endpoint**: remove an unnecessary `weak_ptr`/`weak_from_this()` indirection that had been added to `TransportImpl::closePeer()`'s timeout continuation - `Reactor::HandlerList` already keeps a `shared_ptr` to every registered handler for its entire lifetime (released only in `~Reactor()`), so the "TransportImpl torn down while the response is in flight" scenario the indirection guarded against cannot actually happen. Reverting to capturing `this` removes dead defensive code and a comment that could otherwise mislead a future reader.
- **endpoint**: `Endpoint::setHandler()` dereferenced the incoming handler unconditionally, crashing on a null `shared_ptr` instead of failing predictably (now throws); renamed `handlerMutex_` to `stateMutex_` and extended it to also cover `options_` (previously written unprotected by `init()` while read only under the mutex elsewhere - a data race); and factored the repeated "lock mutex, copy `handler_`" pattern (duplicated three times) into a single `getHandler()` helper.

Each commit is self-contained and can be reviewed/reverted independently.

## ABI impact

`stateMutex_` (Endpoint) and `timersLock` (Transport) are new `std::mutex` data members added to public classes. This is an unavoidable ABI break for the underlying fix - protecting `handler_`/`options_` and `timers` requires *some* form of synchronization state on the object, and there's no way to add that without changing the object's layout.

The one thing that *is* avoidable is making the break worse than it needs to be. Both members were initially declared **before** pre-existing data members (`handler_`, `listener`, `options_`, `logger_` on `Endpoint`; `timers`, `peersQueue`, `loadRequest_`, `notifier`, `handler_`, `peers_mutex_`, `peers_` on `Transport`), which shifts the *offset* of every member that follows them - the dangerous kind of ABI break, since it means a binary built against the old layout would misinterpret those members' memory location entirely, not just fail to fit a resized object.

Both are now declared **last** in their class instead, so adding them only appends to the layout. `sizeof(Endpoint)`/`sizeof(Transport)` still change (that part can't be avoided short of a `pimpl`, which felt like overkill for two `std::mutex` members on a bugfix PR), but no pre-existing member's offset does. `abidiff` will still flag a difference, but it should no longer be the kind that corrupts memory if an old binary is linked against the new library without recompiling - only the kind that requires a recompile.

## Test plan

- [x] `-fsyntax-only` compiles cleanly for every file touched against current `master`.
- [x] `peer_test.cc` (new), and additions to `endpoint_initialization_test.cc` and `reactor_test.cc`, compiled, linked, and run standalone (outside the full build, which this environment's toolchain can't produce - see below) against real gtest: all pass. These are best-effort concurrency stress tests, not deterministic proofs of the underlying races (that would need a thread sanitizer run, which this sandboxed environment can't do either - confirmed by trying).
- [ ] Full build/test suite: this environment's toolchain is too old to build current `master` as a whole (`<charconv>` requires GCC 8+/libstdc++8), so the full suite could not be run locally. `endpoint_initialization_test` is now also registered in `meson.build` (it was previously only in `CMakeLists.txt`), so it will run in CI for the first time as part of this PR - worth a second look if something unrelated to this diff turns up there.


raphaelbergamini:
Closing this for now - a follow-up review of the endpoint.cc handler/options synchronization found a few more issues in the same area (missing null-check in setHandler(), options_ not covered by the new mutex, an unnecessary weak_ptr indirection in closePeer()). I'll fold the fixes in and reopen with an updated PR.

raphaelbergamini:
Reopening with two more commits addressing the follow-up review of this same area:

- Removed the unnecessary weak_ptr/weak_from_this() indirection in TransportImpl::closePeer() (Reactor::HandlerList already keeps handlers alive for their full lifetime, so the teardown race it was guarding against can't happen).
- Added a null-check to Endpoint::setHandler(), extended the mutex to also cover options_ (previously written unprotected by init() while read under lock elsewhere - a data race), and factored the repeated lock+copy-handler_ pattern into a single getHandler() helper.

kiplingw:
Thank you for your contribution, @raphaelbergamini. Are you able to add a unit test?

